PR #2, opened 2015-02-24
Add non-PIP dependency management

quis:
This commit adds dependency management for things which are not available as PIP packages. It uses a combination of:
- npm (GOV.UK front end toolkit)
- Bower (GOV.UK Jinja template, Digital Marketplace front end toolkit, jQuery)

There is a `postinstall` hook in the `package.json` which calls Bower, so **one command (`npm install`) will bring in all the dependencies at their specified versions**.

It does not:
- Wire the template into the app
- Any sort of front end build process or copying of assets (this will probably be Gulp but is something for a separate pull request)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -52,3 +52,8 @@ docs/_build/
 
 # PyBuilder
 target/
+
+# Front end dependencies
+node_modules
+bower_components
+npm-debug.log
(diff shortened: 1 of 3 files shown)


TheDoubleK:
Looks good... at least in that I now have some new folders of stuff that has been downloaded.

PR #5, opened 2015-02-25
Add template to other pages, use toolkit

quis:
Sorry for the size of this pull request, which is mostly in one single commit :disappointed: 

It does three main things:
- Adds the template to the remaining pages
- Makes (almost) all bits of the page use components from the Digital Marketplace frontend toolkit
- Prototype how the flow from page to page might work (without adding any new functionality, ie the app still doesn't _do_ anything)

image

image

image

image


TheDoubleK:
Looking very good once I ran the Gulp thing.  We should add that to the README until it is no longer required.  I have switched Travis on but haven't added a travis.yml so it's trying to run some Ruby tests at the moment.  I'll sort that out in new PR in a sec.  I'll do the README while I'm at it.

PR #7, opened 2015-02-27
Configuration driven admin app

quis:
**TL;DR This handles uploading of documents, but it will let us a build an admin tool that can quickly extend to edit more than documents**

It turns this… | …into this | …and this
---|---|---
image | image | image

In the final screenshot the only hard coded piece of content in the body of the page is the link that says "Admin home". Everything else either comes from the [content](https://github.com/alphagov/digital-marketplace-ssp-content) or the [API](https://github.com/alphagov/digitalmarketplace-api).

## File handling
At the files are stored to `./temp` when `post`ed to the same endpoint as the editing page.

It treats the process of uploading a file as a _validation_, much like checking if a user has inputted service title under 100 characters, or a percentage as just a number.

If the validation fails (eg no file is provided or file can't be stored for some reason) then it passes a message to the template on the `errors` object.

This is built in such a way that a dictionary of validations is provided in the content, eg:
```yaml
validations
  -
    name: is_acceptable_file_size
    message: Your file is too big
  -
    name: is_open_document
    message: That's not an open document format
```
…and then it's up to the app to implement a function called `is_open_document` which returns `True` or `False` when provided `post`ed data.

## Work in progress
- [x] Render service summary page from YAML file
- [x] Render edit section page from YAML file
- [x] Accept file upload(s)
- [x] Validate file format/size
- [x] Render validation errors back onto the form
- [x] Don't error if a file exists but no new one is uploaded
- [x] Save previous file with timestamp
- [x] Deal with optional questions (e.g. SFIA rate card)
- [x] Don't show a section in the summary if none of the questions are relevant
- [x] Make the tests pass on Python 2 and 3

## Out of scope
- Rename files with correct supplier/service ID
- Save files to S3, rather than locally
- Add authentication


app/assets/scss/_edit_section.scss
```diff
@@ -0,0 +1,5 @@
+.question-hint {
```
tombye: This is available in the digital marketplace frontend toolkit: https://github.com/alphagov/digitalmarketplace-frontend-toolkit/blob/master/toolkit/scss/forms/_hint.scss

quis: Addressed in: https://github.com/alphagov/digitalmarketplace-admin-frontend/commit/b6ab578e4c604420b8ef8bf43314add312061441

app/__init__.py
```diff
@@ -1,7 +1,7 @@
 from flask import Flask
 from flask.ext.bootstrap import Bootstrap
 from config import config
-
+from main import main as main_blueprint
```
robyoung: This should be a relative import (ie. prefixed with a `.`).

robyoung: I took a stab at fixing this and there is some import strangeness going on in the tests. Gonna take another look tomorrow.

app/main/views.py
```diff
@@ -1,80 +1,87 @@
...
 from . import main
-from flask import json, render_template, Response, request, redirect
+from .helpers.validation_tools import Validate
+from .helpers.content import Content_loader
+from .helpers.service import Service_loader
+
+
+service = Service_loader(
+    os.getenv('DM_API_URL'),
+    os.getenv('DM_API_BEARER')
```
allait: Based on a conversation with @quis and @TheDoubleK we've decided to use a common convention for ENV variable names: `DM_<app name>_<variable name>`.

So we'll rename `DM_API_BEARER` to `DM_ADMIN_FRONTEND_API_AUTH_TOKEN` so that we can use a different API token for different apps. We can omit the `<app name>` part for variables that will be used by more than one app, like `DM_API_URL`.

quis: @allait Changed in 75933e5

quis:
Would be good to have this PR on the content repo merged first: https://github.com/alphagov/digital-marketplace-ssp-content/pull/33

TheDoubleK:
The "other example fields" are rendering as Python-style dumps... is this a feature for the time being?

screen shot 2015-03-05 at 14 27 26

TheDoubleK:
This might be nicer with some line breaks in the text about each document?
screen shot 2015-03-05 at 14 29 26

TheDoubleK:
Is this 404 something missing from the Toolkit?  Or should it be there?
screen shot 2015-03-05 at 14 32 49

TheDoubleK:
I get a generic "There were problems with your response" message, but nothing specific about which validation rule failed - is that right at this stage?  (This screen shot from a wrong file extension upload):
screen shot 2015-03-05 at 14 34 03

app/main/helpers/validation_tools.py
```diff
@@ -0,0 +1,58 @@
...
+        return len(first_file.read()) > 0
+
+    def file_can_be_saved(self):
+
+        if not self.answer_required():
+            return False
+
+        first_file = self.posted_data[0]
+
+        tmp = "/Users/chs/gdsworkspace/digitalmarketplace-admin-frontend/temp/"
```
TheDoubleK: Is there a way to avoid this hard-coded path?  Would ./temp/ work?

quis: @TheDoubleK Addressed here: 437b4f1

quis:
@TheDoubleK: Thanks for the feedback.
- Python dumps—OK for now, eventually we should have template macros for rendering different kinds of data
- Not going to change the hint text under each question, for now should stay as in the SSP I think
- 404 is because there is no Javascript in this app yet—added an empty placeholder file for now: 20f5e98
- The specific error message is there, it's just not getting any CSS applied. You probably need to do an `npm install` followed by an `npm run frontend-build:development` to get it showing up.

app/main/helpers/content.py
```diff
@@ -0,0 +1,66 @@
+import yaml
+import inflection
+import re
+
+
+class Content_loader():
```
allait: `ContentLoader(object)`

app/main/helpers/content.py
```diff
@@ -0,0 +1,66 @@
...
+            )
+
+            question_content["id"] = question
+
+            # wrong way to do it? question should be shown by default.
+            question_content["depends_on_lots"] = (
+                self.__get_dependent_lots__(question_content["dependsOnLots"])
+            ) if "dependsOnLots" in question_content else (
+                ["saas", "paas", "iaas", "scs"]
+            )
```
allait: `self._question_cache`

app/main/helpers/content.py
```diff
@@ -0,0 +1,66 @@
...
+        for section in self.sections:
+            if section["id"] == requested_section:
+                return section
+
+    def get_question(self, question):
+
+        if question not in self.__question_cache__:
+
+            question_content = yaml.load(
+                open(str(self.directory) + str(question) + ".yml", "r")
```
allait:
```
with open(filename, "r") as f:
    yaml.load(f)
```

app/main/helpers/service.py
```diff
@@ -0,0 +1,32 @@
...
+            raise Exception("DM_API_URL is not set")
+
+    def get(self, service_id):
+        response = requests.get(
+            self.api_url + "/services/" + service_id,
+            headers={
+                "authorization": "Bearer {}".format(self.access_token)
+            }
+        )
+        self.data = json.loads(response.content)["services"]
```
allait: `response.json()`?

app/main/helpers/service.py
```diff
@@ -0,0 +1,32 @@
...
+        return self
+
+    def set(self, key, value):
+        self.data['key'] = value
```
allait: Quotes typo

quis:
@allait Made some changes based on your suggestions: c247893

TheDoubleK:
@quis ah yes, I can see the error message now.  I'll get the hang of the frontend-build eventually :)

PR #8, opened 2015-02-27
Update to run on port 5004

TheDoubleK:
For development we will want to run multiple apps, so they should each bind to a different port number.


quis:
:+1: Just rebased for you @TheDoubleK, will merge when the tests pass…

PR #10, opened 2015-03-03
Update run scripts

TheDoubleK:
Include required environment variables in run scripts.


quis:
Cheers @thedoublek

:+1: if this is how we're doing it elsewhere

TheDoubleK:
This is how I've done it elsewhere, but there may be other ways of doing it...

PR #12, opened 2015-03-06
Add minimal authentication

quis:
**This is a proposal for what I think is the minimum we need to do to get authentication on the admin app**
image

To test this pull request, run the app after setting the following environment variable:
```
export DM_ADMIN_FRONTEND_PASSWORD_HASH='$p5k2$$salt$lHVacy6sSyQWKa953wuFNZwHD04CtzZb'
```
Then log in as `admin` with password `password`.

**This pull request may well not be enough because**
- I know very little about security and may have exposed some kind of glaring vulnerability here
- This just uses HTTP basic auth—if we want to do sessions/nice login screen then more work would be required

Comments encouraged.

### DIFF
--- a/app/main/__init__.py
+++ b/app/main/__init__.py
@@ -1,9 +1,7 @@
 from flask import Blueprint
-from .helpers.auth import requires_auth
+
 
 main = Blueprint('main', __name__)
-
-main.before_request(requires_auth)
 
 from . import views
 from . import errors
--- a/app/main/__init__.py
+++ b/app/main/__init__.py
@@ -1,6 +1,9 @@
 from flask import Blueprint
+from .helpers.auth import requires_auth
 
 main = Blueprint('main', __name__)
+
+main.before_request(requires_auth)
 
 from . import views
 from . import errors
--- a/tests/app/test_application.py
+++ b/tests/app/test_application.py
@@ -4,7 +4,7 @@ from .helpers import BaseApplicationTest
 class TestApplication(BaseApplicationTest):
     def test_index(self):
         response = self.client.get('/')
-        assert 401 == response.status_code
+        assert 200 == response.status_code
 
     def test_404(self):
         response = self.client.get('/not-found')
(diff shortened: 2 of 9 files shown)


robyoung:
If all routes require authentication you could move the authentication to a before filter which would make it harder to accidentally forget when implementing a new view.
See: https://github.com/alphagov/digitalmarketplace-api/commit/167a05d31cd0a434b3a541c06cb2e7bd780b3c7a

TheDoubleK:
I can't see a .htaccess file ;)

Otherwise, yep I reckon do as Rob says.

quis:
@robyoung That's much neater. Thanks for the tip (and pre-written commit message): a430a5c

tests/app/test_application.py
```diff
@@ -4,7 +4,7 @@
 class TestApplication(BaseApplicationTest):
     def test_index(self):
         response = self.client.get('/')
-        assert 200 == response.status_code
+        assert 401 == response.status_code
```
TheDoubleK: We're going to need to work out how to log the tests in.  But maybe not today.

quis: @TheDoubleK Probably have to disable authentication for the tests. And I'm not sure how we're going to deal with testing this app on production—don't want them editing real services.

app/main/__init__.py
```diff
@@ -1,6 +1,9 @@
 from flask import Blueprint
+from .helpers.auth import requires_auth
 
 main = Blueprint('main', __name__)
 
+main.before_request(requires_auth)
```
TheDoubleK: Can this be wrapped in an ``if not test`` type thing?  There is the Test config that we can use. Think it should work.

quis: @TheDoubleK Figured it out this morning, have a look at 6e370fc and see if that meets your needs.

TheDoubleK:
Nice.
Of course, we now have no way of unit-testing if the authentication works :/

PR #13, opened 2015-03-09
Add S3 document uploads

allait:
Uploads documents to S3, renaming the existing file with an added timestamp. Sets the uploaded file permissions to `'public-read'`.

Uses a local question_id to filename suffix mapping for now, will have to change it to use the ssp-content rules instead.


TheDoubleK:
In the story on Pivotal (https://www.pivotaltracker.com/story/show/88248402) it specifies that the S3 uploads should be done by a shared library:
"Sharing code library: S3 intergration code is shared code"

Because this S3 upload functionality will eventually be required by both the Admin and Supplier apps it is a good opportunity to start a "commons" type library for shared code - I'm not sure the way this is usually done in Python, but it might mean extracting some of the stuff from "helpers" into a new repository, rather than having them in the admin app's codebase.

PR #14, opened 2015-03-10
Add presenter layer

quis:
*As seen in show and tell™*

Before we put service data onto a page, we might want to do two things:
- Pre-process the data from the API into a different format
- Specify a template macro which is suitable for the type of data

This commit looks up both of these things based on the `type` property of the `question` object in the content but defaults to doing no processing if a presenter is not defined.

This means that we have more control over how each row in the summary is displayed, ie we can go:

…from this | …to this
--- | ---
image | image


app/templates/macros/answers.html
```diff
@@ -1,9 +1,35 @@
-{% macro textbox(field, value='') -%}
-  <fieldset class="question" id="{{ field.name }}">
-    <label class="question-heading" for="{{ field.name }}">{{ field.question }}</label>
-    <p class="question-hint">
-      {{ field.hint }}
-    </p>
-    <input type="text" class="text-box" name="{{ field.name }}" id="{{ field.name }}-text-box" value="{{ value }}" />
-  </fieldset>
+{% macro textbox(value='') -%}
+  txtbox
```
allait: Is this right?

quis: Well spotted—fixed in 6f2f07c

allait:
:+1:

PR #15, opened 2015-03-11
Correct copyright notice

TheDoubleK:


LICENCE
```diff
@@ -1,6 +1,4 @@
-The MIT License (MIT)
-
-Copyright (c) 2015 GOV.UK
+Copyright (C) 2015 Crown Copyright (Government Digital Service)
```
tombye: Think the `(C)` should be `(c)` here (according to usage across other GDS projects).

TheDoubleK:
Discussed with @quis and as TM and R for registered trademark, we think the Copyright is probably a capital C.

PR #17, opened 2015-03-16
Generate and document password hashes

quis:
This commit does three main things:
1. Add a script to generate password hashes (`./scripts/generate_password.py`)
2. Improve the strength of the password hashes by doing 10,000 iterations of the hash function
3. Update the documentation with guidance on using the script


allait:
Looks great, thanks Chris! :+1:

PR #21, opened 2015-03-20
Update document mime type

TheDoubleK:
Current document updates in the admin app do not preserve the content type - it is reset to "application/octet-stream".
This sets the correct content type so that browsers know what to do when users access the documents.

I've tested it locally for uploads into the admin-frontend-dev-documents S3 bucket - look in folder '93292' for a document uploaded by this code. (The Properties/Metadata/Content-Type is set to application/pdf)

### DIFF
--- a/app/main/helpers/s3.py
+++ b/app/main/helpers/s3.py
@@ -1,6 +1,7 @@
 import os
 import boto
 import datetime
+import mimetypes
 
 
 class S3(object):
@@ -23,5 +24,6 @@ class S3(object):
             )
 
         key = self.bucket.new_key(full_path)
-        key.set_contents_from_file(file)
+        mimetype, _ = mimetypes.guess_type(key.name)
+        key.set_contents_from_file(file, headers={'Content-Type': mimetype})
         key.set_acl(acl)
--- a/app/main/helpers/s3.py
+++ b/app/main/helpers/s3.py
@@ -25,10 +25,10 @@ class S3(object):
         key = self.bucket.new_key(full_path)
         key.set_contents_from_file(file,
                                    headers={'Content-Type':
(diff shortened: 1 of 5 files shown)


app/main/helpers/s3.py
```diff
@@ -23,5 +24,6 @@ def save(self, path, name, file, acl='public-read'):
             )
 
         key = self.bucket.new_key(full_path)
-        key.set_contents_from_file(file)
+        mimetype, _ = mimetypes.guess_type(key.name)
+        key.set_contents_from_file(file, headers={'Content-Type': mimetype})
```
robyoung: Do we need a test around this?

TheDoubleK:
I've added a test of the detected MIME type for documents.  In order to test it I have made the save() S3 method return the key (previously it didn't return anything).

I'm not sure if that's the best way to test the detected content-type, happy to discuss alternative ways to get at this in the tests if you don't like the look of it.

TheDoubleK:
Merge conflicts fixed with rebase.

app/main/helpers/s3.py
```diff
@@ -21,7 +22,13 @@ def save(self, path, name, file, acl='public-read'):
                 self.bucket_name,
                 full_path
             )
-
         key = self.bucket.new_key(full_path)
-        key.set_contents_from_file(file)
+        key.set_contents_from_file(file,
+                                   headers={'Content-Type':
+                                            self.__get_mimetype(key.name)})
```
allait: Should be `_get_mimetype`. Double-underscore methods get name-mangled and are useful only for things you don't want to be overridden by subclasses.

TheDoubleK:
Updated as per discussion with @allait.

PR #22, opened 2015-03-23
Replaces basic auth with a login page and cookie-based sessions.

quis:
This commit adds a login page, as per [#90204166](https://www.pivotaltracker.com/n/projects/997836/stories/90204166).

It reuses the same authentication, ie it does not do any user management. It does use Flask's session handling. Session lifetime is 1 hour, and is renewed on every request.

It adds a new environment variable (`DM_ADMIN_FRONTEND_COOKIE_SECRET`) which is used to encrypt the cookie. This will need to be set to something secret when deploying the app.

image

image

image

image
image
image


tests/app/helpers.py
```diff
@@ -11,3 +11,13 @@ def setUp(self):
...
         self.s3.stop()
+
+
+class LoggedInApplicationTest(BaseApplicationTest):
+    def setUp(self):
+        super(LoggedInApplicationTest, self).setUp()
+        with self.client.session_transaction() as session:
+            session['username'] = 'admin'
+
+    def tearDown(self):
```
allait: We're not overriding the tearDown from `BaseApplicationTest`, so we can just remove it

quis: @allait Fixed in 4507259

PR #23, opened 2015-03-25
API service update

allait:
Adds a call to the API to update service fields. New field values are looked up in the `Validate.clean_data`, so one of the field validators should populate the value for the field. For documents it's done by `file_can_be_saved`.

Update API is called with current username as update author and 'admin app' as reason. Request is only sent if any of the fields have changed.

Only the fields that have passed validation are saved.

`Validate.errors` are now cached, so the validators won't run more than once.

S3 file upload now takes the current file name as an argument and renames the existing file even if it has a different extension.

### DIFF
--- a/app/main/helpers/s3.py
+++ b/app/main/helpers/s3.py
@@ -38,8 +38,6 @@ class S3(object):
                 existing_path
             )
 
-            self.bucket.delete_key(existing_path)
-
     def _get_mimetype(self, filename):
         mimetype, _ = mimetypes.guess_type(filename)
         return mimetype
--- a/app/main/helpers/s3.py
+++ b/app/main/helpers/s3.py
@@ -14,15 +14,10 @@ class S3(object):
         self.bucket_name = bucket_name
         self.bucket = conn.get_bucket(bucket_name)
 
-    def save(self, path, file, existing_path=None,
-             acl='public-read', move_prefix=None):
+    def save(self, path, file, acl='public-read', move_prefix=None):
         path = path.lstrip('/')
 
-        if existing_path:
-            existing_path = existing_path.lstrip('/')
-            self._move_existing(existing_path, move_prefix)
-        else:
-            self._move_existing(path, move_prefix)
+        self._move_existing(path, move_prefix)
 
         key = self.bucket.new_key(path)
         key.set_contents_from_file(file,
--- a/app/main/helpers/s3.py
+++ b/app/main/helpers/s3.py
@@ -11,14 +11,15 @@ class S3(object):
         self.bucket_name = bucket_name
         self.bucket = conn.get_bucket(bucket_name)
 
-    def save(self, path, file, existing_path=None, acl='public-read'):
+    def save(self, path, file, existing_path=None,
+             acl='public-read', move_prefix=None):
         path = path.lstrip('/')
 
         if existing_path:
             existing_path = existing_path.lstrip('/')
-            self._move_existing(existing_path)
+            self._move_existing(existing_path, move_prefix)
         else:
-            self._move_existing(path)
+            self._move_existing(path, move_prefix)
 
         key = self.bucket.new_key(path)
         key.set_contents_from_file(file,
@@ -27,13 +28,14 @@ class S3(object):
         key.set_acl(acl)
         return key
 
-    def _move_existing(self, existing_path):
-        timestamp = datetime.datetime.utcnow().isoformat()
+    def _move_existing(self, existing_path, move_prefix=None):
+        if move_prefix is None:
+            move_prefix = default_move_prefix()
 
         if self.bucket.get_key(existing_path):
             path, name = os.path.split(existing_path)
             self.bucket.copy_key(
-                os.path.join(path, '{}-{}'.format(timestamp, name)),
+                os.path.join(path, '{}-{}'.format(move_prefix, name)),
                 self.bucket_name,
                 existing_path
             )
@@ -43,3 +45,7 @@ class S3(object):
     def _get_mimetype(self, filename):
         mimetype, _ = mimetypes.guess_type(filename)
         return mimetype
+
+
+def default_move_prefix():
+    return datetime.datetime.utcnow().isoformat()
--- a/app/main/helpers/s3.py
+++ b/app/main/helpers/s3.py
@@ -11,23 +11,34 @@ class S3(object):
         self.bucket_name = bucket_name
         self.bucket = conn.get_bucket(bucket_name)
 
-    def save(self, path, name, file, acl='public-read'):
-        timestamp = datetime.datetime.utcnow().isoformat()
+    def save(self, path, file, existing_path=None, acl='public-read'):
+        path = path.lstrip('/')
 
-        full_path = os.path.join(path, name)
+        if existing_path:
+            existing_path = existing_path.lstrip('/')
+            self._move_existing(existing_path)
+        else:
+            self._move_existing(path)
 
-        if self.bucket.get_key(full_path):
-            self.bucket.copy_key(
-                os.path.join(path, '{}-{}'.format(timestamp, name)),
-                self.bucket_name,
-                full_path
-            )
-        key = self.bucket.new_key(full_path)
+        key = self.bucket.new_key(path)
         key.set_contents_from_file(file,
                                    headers={'Content-Type':
                                             self._get_mimetype(key.name)})
         key.set_acl(acl)
         return key
+
+    def _move_existing(self, existing_path):
+        timestamp = datetime.datetime.utcnow().isoformat()
+
+        if self.bucket.get_key(existing_path):
+            path, name = os.path.split(existing_path)
+            self.bucket.copy_key(
+                os.path.join(path, '{}-{}'.format(timestamp, name)),
+                self.bucket_name,
+                existing_path
+            )
+
+            self.bucket.delete_key(existing_path)
 
     def _get_mimetype(self, filename):
         mimetype, _ = mimetypes.guess_type(filename)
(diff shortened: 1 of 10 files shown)


app/main/helpers/validation_tools.py
```diff
@@ -1,31 +1,40 @@
...
         self.service = service
         self.posted_data = posted_data
         self.uploader = uploader
+        self.clean_data = {}
+        self._errors = None
 
     @property
     def errors(self):
+        if self._errors is not None:
+            return self._errors
```
quis: Nice detail.

app/main/helpers/validation_tools.py
```diff
@@ -33,66 +42,77 @@ def question_errors(self, question_id):
...
-
-    def test(self, question, rule):
+    def test(self, question_id, question, rule):
         if not hasattr(self, rule):
             raise ValueError("Validation rule " + rule + " not found")
-        return getattr(self, rule)(question)
+        return getattr(self, rule)(question_id, question)
 
-    def answer_required(self, question):
+    def answer_required(self, question_id, question):
```
quis: What is `question_id` being used for here?

allait: It's not used in `answer_required`, but it's used in `file_can_be_saved`, and `test` runs every rule the same way

app/main/helpers/validation_tools.py
```diff
@@ -33,66 +42,77 @@ def question_errors(self, question_id):
...
         )
 
+        full_url = urlparse.urljoin(
+            main.config['DOCUMENTS_URL'],
+            file_path
+        )
+
+        self.clean_data[question_id] = full_url
+
         return True
```
quis: Is it possible to return `false` here if `self.uploader.save` fails for any reason? Would be better to show the user an error rather than an exception.

app/main/helpers/validation_tools.py
quis: Ignore the above comment—reading further I can see why you have to pass it even when it's not used.

app/main/helpers/validation_tools.py
allait: That's a great point. I've added a `try.. except` block around `self.uploader.save` and a test to check for validation message in case of an error

quis:
:+1: from me.

app/main/helpers/s3.py
```diff
@@ -1,8 +1,11 @@
 import os
 import boto
+import boto.exception
 import datetime
 import mimetypes
 
+from boto.exception import S3ResponseError  # noqa
```
TheDoubleK: This line isn't long enough to need #noqa - pep8 passes without it.

quis: Oh no, you've discovered the secret @TheDoubleK

allait: It's an unused import, pep8 doesn't check for them, but pyflakes and flake8 do.

TheDoubleK:
When you say "S3 file upload now takes the current file name as an argument and renames the existing file even if it has a different extension.", does this mean that if "something.odt" is uploaded to replace "something.pdf" then it will be renamed to have a .pdf extension? If so that doesn't seem right, it will be served with a misleading name and Content-Type; people won't be able to open the downloaded file easily?

app/main/helpers/validation_tools.py
```diff
@@ -1,31 +1,42 @@
 import os.path
-import time
-import shutil
 
+try:
+    import urlparse
+except ImportError:
+    import urllib.parse as urlparse
```
TheDoubleK: This seems odd - shouldn't installing the requirements ensure that we can import everything we need? i.e. so we shouldn't get ImportErrors?

allait: urlparse is stdlib, but it was moved in python3. This is the most straightforward way to get a 2/3 import that I know of

allait:
>  does this mean that if "something.odt" is uploaded to replace "something.pdf" then it will be renamed to have a .pdf extension?

I probably should've explained this better. The new file will keep the ".odt" extension, but the previous "something.pdf" will be renamed to `<timestamp>-something.pdf` *even though* there's no actual name clash due to different extensions.

@TheDoubleK

app/main/helpers/s3.py
TheDoubleK: So that begs the question... is there a good reason to keep an unused import?

TheDoubleK:
Probably not for this PR, but once G5 services are live we will need some logic in here to not allow updating G5 service documents.

TheDoubleK:
Thanks @allait, yes I worked that out, should have read closer in the first place.

app/main/helpers/s3.py
allait: S3ResponseError is not used in the file itself, but importing it here allows me to import it from `helpers.s3` in `validation_tools.py` and tests. That means we don't have `boto` imported anywhere else and can later replace the exception with our own if we have to.

I could drop the `# noqa` since it doesn't cause issues with our build, but I think it's a good indicator that the fact that the import is not used in the file is what we actually want.

app/main/helpers/validation_tools.py
```diff
@@ -33,66 +44,80 @@ def question_errors(self, question_id):
...
+            question.filename
+        )
 
-        self.uploader.save(
-            'documents/{}'.format(
-                self.service['supplierId'],
-            ),
-            file_name,
-            question
+        if self.service[question_id]:
```
TheDoubleK: Trying to upload a document where one did not previously exist gives an error here:
```
  File "/Users/kevinkeenoy/gdsworkspace/digitalmarketplace-admin-frontend/app/main/helpers/validation_tools.py", line 68, in file_can_be_saved
    if self.service[question_id]:
KeyError: 'sfiaRateDocumentURL'
```
For all lots except SCS an error is the right thing to do here (SCS is the only lot that allows a sfia Rate document), but it should be caught and give a nice error message.

TheDoubleK: And for SCS services it should be possible to add a new sfiaRateDocument

TheDoubleK: I'm happy for "adding new documents" to be another story to be done later though.

quis: @TheDoubleK I _think_ that SFIA rate card was an optional field for all service types. That's what [this line](https://github.com/alphagov/digital-marketplace-ssp-content/blob/6317f59efcce7d7a8ba92fa99cddd476e58506a6/service/pricing.yml#L43) seems to suggest anyway.

app/main/views.py
```diff
@@ -104,21 +104,28 @@ def update(service_id, section):
...
     for question_id in posted_data:
-        if question_id not in errors:
-            service_loader.set(service, question_id, "new value")
-
-    service_loader.post(service)
+        if question_id not in form.errors and question_id in form.clean_data:
+            update[question_id] = form.clean_data[question_id]
+
+    if update:
+        service_loader.post(
```
robyoung: Should we roll back the S3 change if the API update fails? Can we be left in an inconsistent state if we don't?

app/main/views.py
```diff
@@ -104,21 +104,28 @@ def update(service_id, section):
...
-    errors = Validate(content, service, posted_data, s3_uploader).errors
+    form = Validate(content, service, posted_data, s3_uploader)
+    update = {}
 
     for question_id in posted_data:
-        if question_id not in errors:
-            service_loader.set(service, question_id, "new value")
-
-    service_loader.post(service)
+        if question_id not in form.errors and question_id in form.clean_data:
```
robyoung: The side effects of the `errors` property are a bit non-obvious. For example, if these conditions are reversed it no longer works. Could the work of the errors property be done in an explicit validate function (perhaps called from the constructor)? If lazy properties is essential could the `clean_data` attribute be changed to a property that first calls `errors` to preload?

app/main/views.py
quis: If we always timestamp new, rather than previous files (as discussed yesterday, sorry @robyoung ) then inconsistency is unpossible.

robyoung: I like that idea.

app/main/views.py
allait: I was thinking about `validate()`, but I actually quite like the `.errors`. It forces you to deal with errors first before using `cleaned_data`, which is somewhat similar to the Django's `is_valid`.

Reversing the conditions will throw an exception, since clean_data is `None` by default. The same thing would happened if we had the `validate()` method and forgot to call it.

app/main/views.py
allait: In both cases if the API update fails we end up with unused files in S3. It's actually better right now since if you replace a pdf with a pdf the URL stays the same so the API update doesn't actually change anything except creating an archived copy.

At any rate, it seems that it would be hard to consistently recover from this (e.g. if API call failed due to network and so will the S3 delete requests), but we can have a batch task looking for files not referenced in services if this ever becomes a storage/cost problem.

app/main/views.py
robyoung: The bit that I find non-obvious is that a property is doing so much work behind the scenes. I expect a method to potentially change state outside of the method call itself, I wouldn't immediately expect that from a property. It may just be me though.

allait:
I've added new commits to the PR. One fixes the missing 'sfiaRateDocumentURL' service field exception.
Others add timestamps to uploaded file names and clean up unused code/tests from s3 (i.e. removing `existing_path` parameter)

After talking with @minglis we've decided to keep the `_move_existing` for now, so that when the new file will overwrite an existing one (e.g. two file versions uploaded within a minute) the old one is still copied before being replaced.

allait:
Travis is failing because bower can't download stuff from github atm

app/main/helpers/s3.py
TheDoubleK: That all makes sense - yes definitely keep the `#noqa` there or someone will remove it at some point and break things (in all likelihood me).

PR #26, opened 2015-03-31
Edit price, service name and description

quis:
This pull request allows a user of the admin app to edit:
- Service name
- Service description
- Maximum price, minimum price, pricing unit and pricing interval

Pricing is just about the most complex field we had in the SSP so there's quite a lot of code here. The biggest complication is that the API has a key called `priceString` which has to be kept consistent with the `priceMin`, `priceMax`, etc. keys, ie:

On the service page we display | On the edit pricing page we display | What we need to `POST` back to the API
---|---|---
`priceString` | `priceMin`, `priceMax`, `priceUnit`, `priceInterval` | `priceString`, `priceMin`, `priceMax`, `priceUnit`, `priceInterval`
£10 to £20 per instance per day | 10, 20, Instance, Day | £5 to £20 per instance per day, 5, 20, Instance, Day

image

image

image

_N.B. Right now there is [a bug in the API](https://www.pivotaltracker.com/n/projects/997836/stories/91481374) which means that `maxPrice` cannot be cleared once it is set. Fixing this bug should not require any changes to this app. Because it's a fairly unlikely request I don't feel that it should be a blocker for this story._


app/assets/scss/patterns/_pricing.scss
```diff
@@ -0,0 +1,81 @@
+.pricing {
+
+  &-column {
```
tombye: Can we lose this `&-` format? It makes searching for class names you find in the generated CSS difficult.

quis: @tombye 1d9d289

tombye:
:+1:

app/main/views.py
```diff
@@ -120,10 +131,10 @@ def update(service_id, section):
         )
 
     if form.errors:
+        service.update(form.dirty_data)
```
quis: @allait I've added a table to the pull request description which (hopefully) explains it.

allait: @quis thanks Chris! I've looked at the code a bit more and as far as I understand, what happens right now is:
1. All the price fields are posted as 'priceString', ie 'priceString=<min>&priceString=<max>' ... I still don't quite understand why we need `dirty_data` since

app/main/helpers/validation_tools.py
```diff
@@ -16,7 +21,8 @@ def __init__(self, content, service, posted_data, uploader=None):
         self.service = service
         self.posted_data = posted_data
         self.uploader = uploader
-        self.clean_data = None
```
allait: Setting `_data` fields to `None` initially makes dictionary operations fail on them, which means doing something like `service.update(clean_data)` without calling `.errors` first is an explicit error.

quis: @allait That's a cool trick. Changed it back in 14d8e6b

app/main/views.py
quis: @allait Yes, that's exactly how it works. And yes again—trying to keep that logic out of the template was the idea.

allait: And re. `priceString` input names: I like how this means all of the validators are run in order and it's all hidden inside one field. 

But on the other hand this means that all of the Validate test methods are coupled to the `priceString` format (e.g. `max_price_not_a_number` and `min_price_not_a_number` do the same thing, so if the fields were separate we could just have `price_not_a_number`)

The way I've done "composite field" validation before is have steps that would run after individual field validators and have access to the full posted data, so that you could create and join fields like "priceString" there. I have no idea how hard it would be to do this in our case and if it would be better at all.

quis: Right now pricing is the only "composite" field that we have. But this is the second time that the idea of having some kind of post-validation layer has come up. Maybe worth talking about on Tuesday?

PR #28, opened 2015-04-07
Make features and benefits editable

quis:
The rules around editing features and benefits are:
- Maximum of 10 items per list
- Each item can't be more than 10 words in length
- If there is a gap between submitted items, it should be filtered out

### DIFF
--- a/app/main/views.py
+++ b/app/main/views.py
@@ -106,9 +106,12 @@ def update(service_id, section):
 
     # Turn responses which have multiple parts into lists
     for item in request.form.items():
-        item_as_list = request.form.getlist(item[0])
+        key = item[0]
+        item_as_list = request.form.getlist(key)
         if len(item_as_list) > 1:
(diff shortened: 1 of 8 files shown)


quis:
Javascript parts aren't working, closing temporarily…

quis:
Edit: going to raise a separate pull request for adding Javascript enhancement. Review away.

app/main/views.py
```diff
@@ -106,9 +106,12 @@ def update(service_id, section):
 
     # Turn responses which have multiple parts into lists
     for item in request.form.items():
```
allait: `for key in request.form:`?

allait:
Looks great, `request.form` is a nitpick, happy to merge as is :+1:

app/main/views.py
quis: @allait a79948a

PR #29, opened 2015-04-07
Add JS from Digital Marketplace frontend toolkit

quis:
This pull request brings back:
- Word count on the service summary field
- Add/remove item functionality for features and benefits

The tooling is [Gulp include](https://www.npmjs.com/package/gulp-include) which uses [Sprockets](https://github.com/sstephenson/sprockets)-style directives to concatenate multiple Javascript files into one.

It also makes some minor template changes for compatibility with the toolkit Javascript.

words

fb

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -1,4 +1,9 @@
-// Load all the things
+/*
+  The following comments are parsed by Gulp include
+  (https://www.npmjs.com/package/gulp-include) which uses
+  Sprockets-style (https://github.com/sstephenson/sprockets)
+  directives to concatenate multiple Javascript files into one.
+*/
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/application.js
```diff
@@ -0,0 +1,48 @@
+// Load all the things
```
tombye: Instead of this could we add a comment explaining that this uses gulp-include so there is some explanation for this formatting (which doesn't appear anywhere else in the app and is not part of native JS)?

tombye:
Apart from my comment, looks good to me.

app/assets/javascripts/application.js
quis: @tombye a8b20ef

PR #30, opened 2015-04-08
Surface API errors

quis:
If the validations pass but the `post` to the API fails for whatever reason it will appear to the user that their changes have been persisted, when they have, infact, not.

This pull request does the minimum required to indicate to the user that something has gone wrong.

### DIFF
--- a/app/main/views.py
+++ b/app/main/views.py
@@ -128,12 +128,14 @@ def update(service_id, section):
                 update[question_id] = form.clean_data[question_id]
 
     if update:
-        service_loader.post(
+        api_response = service_loader.post(
             service['id'],
             update,
             session['username'],
             'admin app'
         )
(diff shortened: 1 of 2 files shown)


app/main/views.py
```diff
@@ -128,12 +128,14 @@ def update(service_id, section):
...
     if update:
-        service_loader.post(
+        api_response = service_loader.post(
             service['id'],
             update,
             session['username'],
             'admin app'
         )
+        if api_response.status_code is not 200:
+            return "Error: " + str(api_response.content)
```
allait: API response should be a JSON with "error" top-level key, so it might be worth trying to `api_response.json()['error']` and returning content only if that fails

quis:
@allait How about f426c84?

allait:
@quis :+1:

PR #31, opened 2015-04-08
Fix bug with pricing list

quis:
There was a bug when max price was left empty. It meant that max price got replaced with the value of the next item in the pricing list (unit). This is because any list submitted to the app was having its empty items removed.

This commit changes the app to only strip empty items from lists that get validated for their length.


allait:
Looks good. I think we can add a test for `under_10_items` removing the empty values.

quis:
@allait ae4af44

PR #32, opened 2015-04-09
Make one command to initialise the frontend code

tombye:
The current README seems to indicate you just need to run `npm install` to get started with the app. We do list the NPM build commands at the bottom but it's not clear which you need to do before the front-end will be ready.

This creates a command which does just that: `npm run init`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ Install new Python dependencies with pip
 Install frontend dependencies with npm and gulp
 
 ```
-npm install
+npm run init
 ```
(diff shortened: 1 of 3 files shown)


tombye:
This command could easily be confused with `npm init` and it doesn't help that all users will expect to just run `npm install` to get started, as on other projects.

Closing this in favour of better documentation (see https://github.com/alphagov/digitalmarketplace-admin-frontend/pull/33).

PR #36, opened 2015-04-16
Don't show None in the maxPrice field when price is not set

allait:
Would replace 0 with the empty string, but the maxPrice can't be 0


quis:
That's a remarkably simple fix. :+1:

PR #42, opened 2015-04-28
Pc admin aware of api

pcraig3:
Status endpoint for the admin app returns "ok" if the connected to data API.


tests/app/status/test_views.py
```diff
@@ -0,0 +1,32 @@
+from flask import json, Response
+import mock
+
+from ..helpers import BaseApplicationTest
+
+
+class TestSession(BaseApplicationTest):
```
quis: This should probably be named something like `TestStatus`

pcraig3:
`/_status` endpoint now does basically the same thing as the buyer app.

Data API responds normally:
```json
{
  "api_status": {
    "db_version": "10_adding_supplier_to_user_table",
    "status": "ok",
    "version": null
  },
  "status": "ok",
  "version": null
}
```

Data API has no database connection: 
```json
{
  "api_status": {
    "message": "Error connecting to database",
    "status": "error",
    "version": null
  },
  "message": "Error connecting to the (Data) API.",
  "status": "error",
  "version": null
}
```

Data API is down: 
```json
{
  "api_status": null,
  "message": "Error connecting to the (Data) API.",
  "status": "error",
  "version": null
}
```

app/status/views.py
```diff
@@ -1,9 +1,46 @@
...
+    service_loader = ServiceLoader(
+        main.config['API_URL'],
+        main.config['API_AUTH_TOKEN']
+    )
+
+    api_response = utils.return_response_from_api_status_call(
+        service_loader.status
+    )
+
+    apis_wot_got_errors = []
```
TheDoubleK: The variable name doesn't really feel consistent with the rest of our codebase.  It is boring I know, but something like `apis_with_errors` would be better.

TheDoubleK: Sorry.

app/status/views.py
```diff
@@ -1,9 +1,46 @@
...
+
+    if api_response is None or api_response.status_code is not 200:
```
allait: Should be `!= 200`. `is` is an identity comparison, it shouldn't be used for integers/strings.

TheDoubleK: Should use != for integer comparison - 'is' behaves strangely for int in Python (it will work in this case for a 200, but not for, say a 404). We should consistently use == and != for int comparison in all our code.
See here for more details: http://stackoverflow.com/questions/306313/pythons-is-operator-behaves-unexpectedly-with-integers

app/status/views.py
```diff
@@ -1,9 +1,46 @@
...
+from ..main.helpers.service import ServiceLoader
+from ..main import main
 
 
 @status.route('/_status')
 def status():
-    return jsonify(status="ok", version=utils.get_version_label())
+
+    # ServiceLoader is the only thing that actually connects to the API
+    service_loader = ServiceLoader(
```
allait: ServiceLoader is going away in #43, maybe it's worth rebasing on top of that branch. This also means that we need a new place for the `status` method, which also exists in supplier-frontend `api_client` and as `get_api_status` in buyer-frontend `models.py`. I think it's worth replacing all of them with a `dmutils` base client method right away

app/status/views.py
pcraig3: No worries! Easily done.

app/status/views.py
quis: #43 is merged now—suspect that's why this branch will no longer merge.

pcraig3:
(Data) API Client now imported from `dmutils`

requirements_for_test.txt
```diff
@@ -3,3 +3,4 @@
 pep8==1.5.7
 nose==1.3.4
 mock==1.0.1
+requests-mock==0.6.0
```
allait: Missing a newline

app/status/utils.py
```diff
@@ -1,4 +1,5 @@
 import os
+from requests.exceptions import ConnectionError
```
allait: It'd be better to rely on exceptions from the `dmutils.apiclient`, since requests should be an implementation detail at this point. It could be as simple as importing the `requests.exceptions.ConnectionError` in `dmutils.apiclient` and using `from apiclient import ConnectionError` here. But I guess we can change this later, since updating the utils version across repositories is a bit of a chore right now.

PR #43, opened 2015-04-29
Extract data api

robyoung:
Depends on https://github.com/alphagov/digitalmarketplace-utils/pull/8


quis:
I've fixed the problem that Travis was having, but this still won't build until alphagov/digitalmarketplace-utils#8 is merged.

quis:
Going to close and reopen this to make Travis try again…

PR #51, opened 2015-05-11
Move status endpoint to /admin/_status

allait:


TheDoubleK:
This will need a corresponding pull-request in the smoketests to use this new endpoint.

allait:
Yeah, I know. _status smoketests are broken right now since the per-frontend domains do not exist anymore.

PR #52, opened 2015-05-11
Handle 404 from API when getting service

robyoung:


app/main/views.py
```diff
@@ -58,19 +58,22 @@ def find():
...
+                value, content.get_question(key)
+            )
+
+        template_data = get_template_data({
+            "sections": content.sections,
+            "service_data": presented_service_data,
+            "service_id": service_id
+        })
+        return render_template("view_service.html", **template_data)
+    except APIError as e:
```
allait: Unless I'm missing something I think we can just wrap the `service_data = data_api_client.get_service(service_id)` line

PR #53, opened 2015-05-11
[#94078878] Not able to save features and benefits with only one feature or one benefit.

robyoung:
https://www.pivotaltracker.com/story/show/94078878

This also fixes how the API client response is being used. The client now returns the entire response object. This feels a bit clunky but it is how the API is currently returning. I'm going to look at returning a response object that gives access to the metadata and links while giving access to the data in a more natural way.

### DIFF
--- a/app/main/views.py
+++ b/app/main/views.py
@@ -100,7 +100,7 @@ def update(service_id, section):
     # Turn responses which have multiple parts into lists
     for key in request.form:
         item_as_list = request.form.getlist(key)
-        if len(item_as_list) > 1:
+        if content.get_question(key)['type'] == 'list':
             posted_data[key] = item_as_list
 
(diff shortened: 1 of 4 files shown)


app/main/views.py
```diff
@@ -100,7 +100,7 @@ def update(service_id, section):
     # Turn responses which have multiple parts into lists
     for key in request.form:
         item_as_list = request.form.getlist(key)
-        if len(item_as_list) > 1:
+        if content.get_question(key)['type'] == 'list':
```
quis: This change breaks `pricing` type questions. It will also future-break series of checkboxes elements which get `POST`ed as lists.

robyoung:
Closing while I address @quis's comment.

PR #54, opened 2015-05-13
[Fixes 94480244] 500 error on G5 service

robyoung:
Only try to display answers if there are any. We were getting index out
of range errors because there were no features or benefits.

### DIFF
--- a/app/templates/macros/answers.html
+++ b/app/templates/macros/answers.html
@@ -58,7 +58,7 @@
         </li>
       {% endfor %}
     </ul>
-  {% else %}
+  {% elif value|length == 1 %}
     {{ value.0 }}
   {% endif %}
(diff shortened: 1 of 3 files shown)


TheDoubleK:
I can't reproduce the bug locally at all (I do have the service data imported).  I'm not sure why.  (Although I notice that it's one of the very few exported services where 'lot' is missing - don't think this has anything to do with the bug here, but a coincidence.)

I'm going to merge on the basis that the code makes sense and the tests run and nothing breaks when I run the updated code.  Fingers crossed it will fix the bug in preview.

TheDoubleK:
OK, I've seen the bug using a different G5 service ID (one with a lot but no features and benefits).  Confirm that it's fixed by this PR.

PR #58, opened 2015-05-14
[Fixes #94078878] Features and benefits edit

robyoung:
In the previous attempt at this I hadn't actually tested what was being added into the page. I have added a couple of additional tests to cover this.


TheDoubleK:
The linking to pivotal doesn't work both ways, so it would still be nice to have a pull request message that explains what the pull request does.

Copy and pasting the ticket number from here to search pivotal feels like unnecessary hassle.

PR #59, opened 2015-05-14
Admin deactivate a service

TheDoubleK:
See this story: https://www.pivotaltracker.com/story/show/92216650
And also: https://www.pivotaltracker.com/story/show/94082572

Adds radio buttons at the bottom of the admin service page to change the status of a service.
To go from deactivated to published takes two clicks.

I have checked the API code and services are removed from the search index when their status changes from 'published', so there is no extra work to be done for the second story above. (See relevant API code for updating a status here: https://github.com/alphagov/digitalmarketplace-api/blob/master/app/main/views/services.py#L338)

### DIFF
--- a/app/main/helpers/presenters.py
+++ b/app/main/helpers/presenters.py
@@ -18,7 +18,7 @@ class Presenters(object):
             return value
 
     def _service_id(self, value):
-        if re.findall("[a-zA-Z]", value):
+        if re.findall("[a-zA-Z]", str(value)):
             return [value]
         else:
             return re.findall("....", str(value))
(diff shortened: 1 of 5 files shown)


app/main/helpers/presenters.py
```diff
@@ -18,7 +18,7 @@ def present(self, value, question_content):
             return value
 
     def _service_id(self, value):
-        if re.findall("[a-zA-Z]", value):
+        if re.findall("[a-zA-Z]", str(value)):
             return [value]
         else:
             return re.findall("....", str(value))
```
robyoung: Is this what we want? Given the string `qwertyuiop` it will return `['qwert', 'yuiop']`

quis: This is the desired behaviour.

TheDoubleK:
Merged with master because the rebase was killing me... fixing the same thing for every commit.

quis:
This will break if we change the database to speak public/private/removed. But I don't think there's a clean way around that. So I'm going to merge it :+1:

PR #60, opened 2015-05-15
Update dmutils and replace APIError with HTTPError

allait:
Use the new HTTPError exception instead of the base APIError one.

Requires https://github.com/alphagov/digitalmarketplace-utils/pull/36

### DIFF
--- a/tests/app/main/test_views.py
+++ b/tests/app/main/test_views.py
@@ -6,7 +6,7 @@ except ImportError:
     from io import BytesIO as StringIO
 
 import mock
-from dmutils.apiclient import APIError
+from dmutils.apiclient import HTTPError, REQUEST_ERROR_MESSAGE
 
 from ..helpers import BaseApplicationTest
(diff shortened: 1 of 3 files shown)


tests/app/main/test_views.py
```diff
@@ -6,7 +6,7 @@
     from io import BytesIO as StringIO
 
 import mock
-from dmutils.apiclient import APIError
+from dmutils.apiclient import HTTPError, REQUEST_ERROR_MESSAGE
```
TheDoubleK: I have always been told not to use constants from the code you're testing in tests - the tests should have the expected values in the test package (the idea being that if someone messes up the message in the main code then the test will fail because the message is wrong).

allait: That's a good point, but I think it applies to the dmutils tests more than it applies here. If we decide to change the default error message I think it shouldn't make the _app_ tests fail.

TheDoubleK: That's very true.

PR #61, opened 2015-05-19
No trailing slashes on URLs

quis:
Duplicates https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/52


robyoung:
Could we have a test for the redirecting?

quis:
@robyoung Test added in 6e39de6

PR #62, opened 2015-05-19
[Fixes #94162812] Make view service link relative

robyoung:
The link was always pointing to production and was pointing to `/service/<id>` rather than `/services/<id>`.


quis:
This story is marked as blocked because the production buyers' app is still Grails and making this link relative won't point at the Grails app. Do we:
- think it's now unblocked somehow
- not care at the moment
?

robyoung:
That is a very good point. I hadn't spotted that. @andrewchong wasn't aware of that update when I chatted to him earlier. I'm going to close for now, until we can catch up with @andrewchong.

PR #64, opened 2015-06-08
Incorporate Feature Flags.

pcraig3:
The upgraded `dmutils` sets up FeatureFlags (among others) in its `init_app` function, and includes the `get_version_label` and `get_flags` methods.

The `init_app` function also sets up `logging`, `config`, and `proxy_fix`, so we don't have to import them into our api anymore.


pcraig3:
Updated to require as-yet unreleased version of dmutils, so it's going to break until [this pull request](https://github.com/alphagov/digitalmarketplace-utils/pull/54) is merged.

TheDoubleK:
:+1: good to merge once conflicts are resolved.

app/__init__.py
```diff
@@ -19,13 +20,14 @@ def create_app(config_name):
...
-    data_api_client.init_app(application)
+    init_app(
+        application,
+        configs[config_name],
+        bootstrap=bootstrap,
+        data_api_client=data_api_client,
+        feature_flags=feature_flags
+    )
+    # Should be incorporated into digitalmarketplace-utils as well
     csrf.init_app(application)
```
pcraig3: This is a bit sucky -- basically shows how brittle our [`init_app`](https://github.com/alphagov/digitalmarketplace-utils/blob/master/dmutils/__init__.py#L8) function is. 
Would it be better to generalize `init_app()` in dmutils so that it could take a dictionary of objects and call `.init_app(application)` on each of them?

robyoung: I'm not sure we have to worry too much about removing duplication here. Not all applications need `csrf` and all the `csrf` setup is done in `csrf.init_app` so making each app that does require it call `csrf.init_app` seems better than passing it into another function.

PR #65, opened 2015-06-08
Add CSRF tokens to all forms that are POSTed to.

TheDoubleK:
Addresses this bug: https://www.pivotaltracker.com/story/show/96434676

This uses the WTForms CSRF protection to check that all forms that are POSTed have a valid CSRF token. This is the same method as used in the Supplier app. It just uses the basic Form() object as we aren't using WTForms for the rest of the validation.

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -2,6 +2,7 @@ from datetime import timedelta
 
 from flask import Flask, request, redirect
 from flask.ext.bootstrap import Bootstrap
+from flask_wtf.csrf import CsrfProtect
 from dmutils import apiclient, config, logging
 
 from config import configs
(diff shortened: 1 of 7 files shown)


TheDoubleK:
One more thing to fix...

allait:
I've done this slightly differently in the supplier frontend: https://github.com/alphagov/digitalmarketplace-supplier-frontend/commit/5b62bf8071ebb9cc44724585ae29f61bb9fa4071.

This enabled CSRF for all POST requests, even those that don't use the WTF `Form` objects. This means that the only change needed is adding the csrf input field to the form HTML with ` <input type="hidden" name="csrf_token" value="{{ csrf_token() }}" />` and dropping the field before validating request data.

The benefit is that it makes it hard to accidentally skip a POST form when adding Form objects. But  I haven't looked into whether it allows custom error messages. I'm not entirely sure we need them though: CSRF mismatch happens when someone deliberately tries to mess with the form or if the token timed out when the page was open for too long (1hr by default, but I think it's configurable).

TheDoubleK:
Changed to structure suggested by @allait - think this is ready to go now.

PR #67, opened 2015-06-10
Make Gulpfile generic

quis:
This copies the Gulpfile from the suppliers app, which is now generic, and makes the necessary changes to the setup of this app so that it works.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -62,6 +62,7 @@ npm-debug.log
 app/assets/scss/toolkit
 app/templates/govuk
 app/templates/toolkit
+app/content
 
 # Front end generated assets
 app/static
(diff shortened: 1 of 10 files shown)


TheDoubleK:
I have a feeling that the gulpfile change in the suppliers app may have caused the app to break in preview (e.g. see: https://preview.development.digitalmarketplace.service.gov.uk/suppliers/services).  Let's hold off merging this until we get the supplier app fixed and find if there is a problem with the generic version.

allait:
Is this ok to merge? Admin app is currently broken since there was a merge after the new AWS stuff went out.

quis:
@allait I've assigned it to @tombye so he can check that the comments he left on the supplier Gulpfile have also been addressed in this app.

tombye:
:+1:

PR #68, opened 2015-06-11
Deny third-party sites from embedding in an iframe

pcraig3:
Added an `X-Frame-Options` line in our header which denies embedding our site in an iframe by other sites.
All very well and good for this app, but we'll also have to duplicate the code in other apps we want to share this behaviour.
Things to consider:
- We *could* set this at the server level if we wanted to make
this change more uniform.
- [The alphagov/whitehall team have set their public-facing headers
for their content to `ALLOWALL`](https://github.com/alphagov/whitehall/commit/9f8c7ccd1fdf7695e66b1340cb6b89ca0e134f9a), but @minglis says we're not going to allow this at present.

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -45,4 +45,9 @@ def create_app(config_name):
         if request.path != '/' and request.path.endswith('/'):
             return redirect(request.path[:-1], code=301)
 
+    @application.after_request
+    def add_header(response):
+        response.headers['X-Frame-Options'] = 'DENY'
+        return response
+
     return application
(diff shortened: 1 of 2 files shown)


pcraig3:
If we want to replicate the behaviour in all apps, we can put the decorated method into dm-utils in the `init_app` function.

Then it's maybe less clear what's being set when the app is created though.

TheDoubleK:
Tested and works.

Maybe we should also put this on the supplier app - buyer app doesn't matter so much at the moment as it's read-only.

pcraig3:
@minglis thinks we should have it in all of our apps, so I'm putting it in utils and then ([once that's merged](https://github.com/alphagov/digitalmarketplace-admin-frontend/pull/68))  we can removed the logic but keep the test.

PR #73, opened 2015-06-16
Pen test adding headers

minglis:
Adds no cache headers to admin app


robyoung:
The comment on the first commit states that it adds a dependency on a more recent version of utils but it doesn't.

robyoung:
@TheDoubleK's [comment on the other pull request](https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/70#issuecomment-112461019) applies here as well.

PR #75, opened 2015-06-16
Prohibit empty username or password.

pcraig3:
App will reject empty usernames or passwords, tests have been extended to reflect this.
Also isolated the header verification test to bring it more in line with what's happening in the [buyer app](https://github.com/alphagov/digitalmarketplace-buyer-frontend/blob/master/tests/test_application.py#L26) and the [supplier app](https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/master/tests/app/test_application.py#L53).

### DIFF
--- a/tests/app/main/test_views.py
+++ b/tests/app/main/test_views.py
@@ -13,6 +13,26 @@ from ..helpers import LoggedInApplicationTest
 
 
 class TestSession(BaseApplicationTest):
+
+    def _login(self, username, password, is_authenticated=True):
+        post_response = self.client.post('/admin/login', data=dict(
+            username=username,
+            password=password
+        ))
+
+        get_response = self.client.get('/admin')
+
+        if is_authenticated:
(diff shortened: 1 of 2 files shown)


tests/app/main/test_views.py
```diff
@@ -22,29 +42,25 @@ def test_url_with_non_canonical_trailing_slash(self):
...
-        response = self.client.post('/admin/login', data=dict(
-            username="admin",
-            password="wrong"
-        ))
-        self.assertEquals(200, response.status_code)
+    def test_invalid_logins(self):
+        invalid_logins = [
+            ("admin", "wrong"),
+            ("adminadmin", ""),
+            ("", "adminadmin")
```
quis: Do we care that `("ad", "minadmin")` will still work?

TheDoubleK: I think we can live with it until we wire in the proper API auth - coming soon!

PR #76, opened 2015-06-16
Make use of new dmutils features

quis:
Namely:

Add present_all method to presenters
https://github.com/alphagov/digitalmarketplace-utils/pull/61

Make content loader handle dependencies…
https://github.com/alphagov/digitalmarketplace-utils/pull/62

…which means that the app can update to the latest content structure, as
changed in: https://github.com/alphagov/digital-marketplace-ssp-content/pull/42

This pull request also makes _some_ use of the shared templates in the toolkit, but not all of them, because the admin app is a bit more comprehensive than the supplier at this point.

### DIFF
--- a/app/main/views.py
+++ b/app/main/views.py
@@ -67,16 +67,10 @@ def view(service_id):
     except HTTPError:
         flash({'api_error': service_id}, 'error')
         return redirect(url_for('.index'))
-
-    presented_service_data = {}
-    for key, value in service_data.items():
-        presented_service_data[key] = presenters.present(
-            value, content.get_question(key)
-        )
-
+    content.filter(service_data)
     template_data = get_template_data({
         "sections": content.sections,
-        "service_data": presented_service_data,
+        "service_data": presenters.present_all(service_data, content),
         "service_id": service_id
     })
     return render_template("view_service.html", **template_data)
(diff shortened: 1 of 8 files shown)


app/templates/index.html
```diff
@@ -7,28 +7,30 @@
...
-        {% endfor %}
-    </div>
-{% endif %}
-{% endwith %}
-    
-  {{ page_heading("Find a service") }}
+        {%
+          with
+          message = displayed_message,
+          type = "error" if category == 'error' else "success"
```
robyoung: This should be `type = "destructive" ...`

app/main/views.py
```diff
@@ -67,16 +67,10 @@ def view(service_id):
...
         flash({'api_error': service_id}, 'error')
         return redirect(url_for('.index'))
-
-    presented_service_data = {}
-    for key, value in service_data.items():
-        presented_service_data[key] = presenters.present(
-            value, content.get_question(key)
-        )
-
+    content.filter(service_data)
```
robyoung: We just discussed this. It's not thread safe because it changes the state of the content loader, which is shared.

quis:
Closing this until issues are resolved.

PR #77, opened 2015-06-17
Add audit events

minglis:
Adds a "service updates" page to the admin app

- this is the first example of using audit events
- uses only service updates, no other audit events are involved - rendering logic depends on this
- shows a page to list events
- page has a "filter" fields which allow searching audit events by "acknowledged"  and by date
- date searching is by day (format YYYY-MM-DD)
- users can acknowdege an event which marks the fact that user did it in the database, via the API.

- UI probably is not perfect (!) 

Delivers on:
https://www.pivotaltracker.com/story/show/95431796

Depends on:
https://github.com/alphagov/digitalmarketplace-api/pull/149
and
https://github.com/alphagov/digitalmarketplace-utils/pull/68


app/assets/scss/_pagination.scss
```diff
@@ -0,0 +1,79 @@
+nav[role=navigation] {
+display: block;
+}
+
+.previous-next-navigation {
```
robyoung: It looks like this copied straight from the buyer frontend. Should it be in the frontend toolkit?

app/main/forms.py
```diff
@@ -0,0 +1,36 @@
...
+class ServiceUpdateAuditEventsForm(Form):
+    audit_date = DateField(
+        'Audit Date',
+        format='%Y-%m-%d',
+        validators=[Optional()])
+    acknowledged = StringField(
+        'acknowledged',
+        default="false",
+        validators=[
+            Regexp('^(all|true|false)$'),
```
robyoung: Should this be a [`RadioField`](http://wtforms.readthedocs.org/en/latest/fields.html#wtforms.fields.RadioField)?

app/assets/scss/_pagination.scss
quis: It should, at some point.

app/main/service_update_audits.py
```diff
@@ -0,0 +1,62 @@
...
+            "service_update_audits.html",
+            today=form.format_date_for_display(),
+            acknowledged=form.default_acknowledged(),
+            audit_events=audit_events['auditEvents'],
+            form=form,
+            **get_template_data())
+    else:
+        return render_template(
+            "service_update_audits.html",
+            today=datetime.now().strftime("%d/%m/%Y"),
```
robyoung: I think the formatting of the date should be done in the template with the newly created `dateformat` template filter.

app/templates/_proposition_header.html
```diff
@@ -9,7 +9,8 @@
     {% if authenticated %}
       <ul id="proposition-links">
         <li>
-          <a href="{{ url_for('.logout') }}">Log out</a>
+            <a href="{{ url_for('.service_update_audits') }}">Service Updates</a>
+            <a href="{{ url_for('.logout') }}">Log out</a>
```
robyoung: Should these links be in separate list item elements?

app/templates/service_update_audits.html
```diff
@@ -0,0 +1,154 @@
+{% from 'macros/page_heading.html' import page_heading %}
+
+{% extends "_base_page.html" %}
+{% block page_title %}
+Digital Marketplace admin
+{% endblock %}
```
robyoung: Do you think we should set this in `_base_page.html`? It looks like it would be the same title for all pages in the admin app.

app/templates/service_update_audits.html
```diff
@@ -0,0 +1,154 @@
...
+                        <label for="acknowledged-3">
+                            <input name="acknowledged" value="false" id="acknowledged-3" type="radio"
+                                   aria-controls=""
+                                   {% if acknowledged=="false" %} checked {% endif %}>
+                            Not acknowledged
+                        </label>
+
+                    </div>
+                </div>
+            </div>
```
robyoung: I'm not sure if it's a big deal but I think the preferred way of using the frontend toolkit is via the provided templates now, so that markup changes can be rolled out more easily. cc/ @tombye @quis

app/templates/service_update_audits.html
```diff
@@ -0,0 +1,154 @@
...
+
+                    <li class="next">
+                        <a href="">
+                            Next <span class="visuallyhidden">page</span>
+                            <span class="page-numbers">2 of 3</span>
+                        </a>
+                    </li>
+                </ul>
+            </nav>
+            -->
```
robyoung: My little brain doesn't understand. I'm assuming this is necessary because it's the only place the pagination navigation is used but it's commented out. How does it work?

app/__init__.py
```diff
@@ -47,4 +47,19 @@ def remove_trailing_slash():
         if request.path != '/' and request.path.endswith('/'):
             return redirect(request.path[:-1], code=301)
 
+    @application.after_request
```
allait: dmutils should already be doing this for all apps: https://github.com/alphagov/digitalmarketplace-utils/blob/master/dmutils/__init__.py#L45

tests/app/main/test_service_update_audits.py
```diff
@@ -0,0 +1,344 @@
...
+
+        data_api_client.find_audit_events.assert_called()
+
+    @mock.patch('app.main.service_update_audits.data_api_client')
+    def test_should_render_correct_form_defaults(self, data_api_client):
+        response = self.client.get('/admin/service-updates')
+        self.assertEquals(200, response.status_code)
+
+        self.assertIn(
+            '<input class="filter-field-text" id="audit_date" name="audit_date" type="text" value="">',  # noqa
```
allait: You could use `lxml`instead of string comparisons for cases like this. I think some of the other tests might already doing this, since it's listed in the `requirements_for_test`

app/main/forms.py
minglis: Had lots of issues with rendering the Radio Fields, so opted for toolkit HTML and String field to back it.

allait: I think `AnyOf` validator would fit here better

app/main/service_update_audits.py
minglis: Added a new formatted to handle formatting date objects - previous formatted only string representations of dates.

app/templates/service_update_audits.html
minglis: Sorry I put his in to ensure we had all the CSS, put I haven't implemented pagination yet. Commented out rather than deleted. Bad patter I know.

minglis: But it took a while to get all the CSS to align properly...

minglis: And once branch is deleted....

tests/app/main/test_service_update_audits.py
minglis: Not seen that, will look at it.

app/__init__.py
minglis: Removed

PR #78, opened 2015-06-18
/admin/services should 404 with no service ID

quis:
…rather than return a 500 error.

Fulfils: https://www.pivotaltracker.com/story/show/96881132

### DIFF
--- a/app/main/views.py
+++ b/app/main/views.py
@@ -53,7 +53,7 @@ def logout():
 @main.route('/services', methods=['GET'])
 def find():
     if request.args.get("service_id") is None:
-        abort(404)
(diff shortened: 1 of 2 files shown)


TheDoubleK:
Would it be nicer to have the form reload with a "Couldn't find that service ID" flashed message? Or probably not a priority for now?

quis:
This is only for when no service ID is given. But I agree that just showing 'page not found' isn't very helpful. Have changed it to re-show the 'enter service ID' form if you just click the search button without entering anything.

TheDoubleK:
Ah yep, sorry.  Still nicer like this :+1:

PR #80, opened 2015-06-24
Fix disappearing fields bug

quis:
Brings in https://github.com/alphagov/digitalmarketplace-utils/pull/74 as well as some other fixes to the utils.

This fixes a bug where filtering of the content persisted between requests, causing some fields which should have been shown to start disappearing (terminationCost was the example found by the smoke tests).

### DIFF
--- a/app/main/views.py
+++ b/app/main/views.py
@@ -5,16 +5,17 @@ from dmutils.apiclient import HTTPError
 from .. import data_api_client
 from . import main
 from dmutils.validation import Validate
-from dmutils.content_loader import ContentLoader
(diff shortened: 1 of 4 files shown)


allait:
Could we add a test for the scenario that shows how the issue occurred? Something like:

1. Request a SaaS service
2. Check that terminationCost is not shown
3. Request SCS service
4. Check that terminationCost is shown.

quis:
@allait Turned your `<ol>` into code: b0044d1

allait:
@quis :+1: It's good to merge, just needs a rebase.

quis:
@allait Rebased

PR #81, opened 2015-06-25
Refactor views into a package

robyoung:
This better matches the other applications and makes it more obvious
what a modules contain views.


quis:
:+1: To self-merge once the tests pass

PR #83, opened 2015-06-30
[#95620350] Use api for authentication

robyoung:
Depends on https://github.com/alphagov/digitalmarketplace-utils/pull/97.

This changes the admin frontend to use the data API for user authentication with Flask-Login.

The `User` model has moved to digitalmarketplace-utils as part of this.


TODO.md
```diff
@@ -0,0 +1 @@
+- the authenticated template flag has been removed
```
allait: Was this supposed to be committed?

robyoung: Whoopsie!

allait:
:+1: from me, but since this is security-related I think we should have more people look through it

allait:
Old envrionment variables and config keys can be probably removed now

app/__init__.py
```diff
@@ -30,21 +33,19 @@ def create_app(config_name):
...
         bootstrap=bootstrap,
         data_api_client=data_api_client,
-        feature_flags=feature_flags
+        feature_flags=feature_flags,
+        login_manager=login_manager
     )
     # Should be incorporated into digitalmarketplace-utils as well
     csrf.init_app(application)
 
+    application.permanent_session_lifetime = timedelta(hours=1)
```
TheDoubleK: An hour is probably fine here too, but might we want admin app to be shorter than for suppliers?  I guess if it is locked down to AH and only Nicole as a user this will be fine though.

app/templates/login.html
```diff
@@ -8,15 +8,26 @@
...
-      type = "success"
-    %}
-      {% include "toolkit/notification-banner.html" %}
-    {% endwith %}
-  {% endif %}
+  {% with messages = get_flashed_messages(with_categories=true) %}
+    {% if messages %}
+      {% for category, message in messages %}
+        {% if message == 'no_account' %}
+          {% set displayed_message = "Sorry, we couldn't find a admin account with that username" %}
```
TheDoubleK: Should be "an admin account"

app/main/views/login.py
```diff
@@ -1,32 +1,56 @@
...
+            ), 403
 
-    return render_template("login.html", **get_template_data({
-        "error": "Could not log in",
-        "previous_responses": request.form
-    }))
+        user = User.from_json(user_json)
+        login_user(user)
+
+        if next_url and next_url.startswith('/admin'):
```
TheDoubleK: This doesn't seem to work - it just takes me to the "Find a service" page no matter what is in as `next`.

TheDoubleK: Ah - it needs to be set in the login form for it to be passed in to the POST, as here: https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/ba5af6be0397972d737ace6bd3fef98db14a4d27/app/main/views/login.py#L22

app/templates/login.html
```diff
@@ -25,17 +36,44 @@
...
-      {% include "toolkit/forms/validation.html" %}
-    {% endwith %}
-  {% endif %}
-  <form autocomplete="off" action="{{ url_for('.login') }}" method="post" class="question">
-    <div style="display:none;"><input id="csrf_token" name="csrf_token" type="hidden" value="{{ csrf_token() }}"></div>
-    <label class="question-heading" for="username">Username</label>
-    <input autocomplete="off" type="text" name="username" id="username" class="text-box" value="{{previous_responses.username}}">
-    <label class="question-heading" for="password">Password</label>
-    <input autocomplete="off" type="password" name="password" id="password" class="text-box">
+  <form autocomplete="off" action="{{ url_for('.render_login') }}" method="post">
```
TheDoubleK: Or if you leave the form action empty here then I think it might post including the GET parameters.  But I'm not 100% sure that will work.

PR #85, opened 2015-07-06
[#97091228] Add diff page

pcraig3:
This pull request (nearly) completes the "[Front-end diff page](https://www.pivotaltracker.com/story/show/97091228)" story on Pivotal.   There's, as far as I know, no capacity for an admin user (Nicole) to add a note to an acknowledged change, nor to revert one.

Admin users can see a line-by-line word-highlighted diff of which changes were made and when they were made.  It's also possible to change the URL to get diffs for any set of revisions (ie, it would be possible to see the first-ever revision compared to the current one).

Works reasonably on mobile, but probably you don't want to be doing that too often.

Also changed a few instances of `.now()` to `.utcnow()`, and put a few of the date/time strings we were using throughout the app in the `__init__.py` file

### Notes

* Whitespace changes don't show up on the diff page (though they will be saved to the database).
* It's possible to create an audit event by clicking "Save" without actually changing any content.  Two revisions would exist, but they wouldn't have any content changes, and the resulting diff diff page would just have an empty space under the 'Service revisions' header.  It's technically accurate but it probably isn't the behaviour we want.
* Anything wrong with the `archive_id`s in the url (if they don't exist, if they refer to different services, if they're in reverse chronological order) results in a [404](https://github.com/alphagov/digitalmarketplace-admin-frontend/blob/add-diff-page/app/main/views/services.py#L211).
* Potentially we should move the DATETIME strings to utils.
* <del>'View service' button on mobile version is noticeably off-centre.</del>
* <del>Structure of DiffTools class seems a bit weird to me. Sound off if you have opinions.</del>
* <del>Was having problems with isolating testing logic in `tests/app/test_diff_tools.py`</del>, but since changing the structure of DiffTools, I've rewritten the tests.
* Finally, it's not possible to do much actionable stuff from the diff page as an admin user.  You can 'View' the service or change its status, but it's not possible to (a) acknowledge the audit event, or (b) revert the change in any way.  Maybe we should add an 'Acknowledge' button to the page or something.

***

This is a screenshot for the diff page after having made the following changes:

##### Service features 
1. Removed the word "Configurations"
2. Changed "Application" to "Apple Pie"
3. Added "Three new words" to the end of an existing line.
4. Removed all previous content and added new content.

##### Service benefits 
1. Added a period to the end of the line.
2. Added a new line.

*Large screen version*

screen shot 2015-07-06 at 11 00 56

*Small screen version*
*(Updated to show new 'View service' button.  Past screenshot [here](https://cloud.githubusercontent.com/assets/2454380/8520060/14aba228-23cf-11e5-8a1d-dd2438661b6a.png))*

screen shot 2015-07-06 at 14 30 50

***

[>> Story on Pivotal](https://www.pivotaltracker.com/story/show/97091228)


app/__init__.py
```diff
@@ -56,16 +59,16 @@ def remove_trailing_slash():
...
     def timeformat(value):
         return datetime.strptime(
-            value, formats.DATETIME_FORMAT).strftime('%H:%M:%S')
+            value, formats.DATETIME_FORMAT).strftime(DISPLAY_TIME_FORMAT)
 
     @application.template_filter('dateformat')
     def dateformat(value):
         return datetime.strptime(
-            value, formats.DATETIME_FORMAT).strftime('%d/%m/%Y')
+            value, formats.DATETIME_FORMAT).strftime(DISPLAY_DATE_FORMAT)
```
allait: API times are in UTC, should we convert them to GMT/BST for display?

app/main/helpers/diff_tools.py
```diff
@@ -0,0 +1,275 @@
+import difflib
+from abc import ABCMeta, abstractmethod
+from flask import Markup
```
allait: Since `Markup` objects are rendered as-is they open up a possibility for an XSS or HTML injection attack if they include any user input. For diffs, that means that if the new service version contains any HTML tags they will be rendered in the output.

To prevent this we need to either call `flask.escape` manually on any part of the user data that gets added to the `Markup` objects.

pcraig3: Good catch!  [Think I've fixed it](https://github.com/alphagov/digitalmarketplace-admin-frontend/commit/b05546df932097a8a90eb241adceb2b984b4869f).

app/main/helpers/diff_tools.py
```diff
@@ -0,0 +1,276 @@
...
+
+        lines = {
+            'revision_1': [],
+            'revision_2': []
+        }
+
+        # Make sure lists of strings are same length
+        # Pad the shorter list with empty strings if necessary
+        # Also means pair of lists in returned dictionary are of equal length
+        revision_1, revision_2 = self._pad_lists_to_max_length(
```
allait: You could use [itertools.izip_longest](https://docs.python.org/2/library/itertools.html#itertools.izip_longest) instead of a padding method

app/main/helpers/diff_tools.py
```diff
@@ -0,0 +1,276 @@
...
+
+    def _split_words_diff(self, words_diff):
+        """
+        Accepts a list of words that have been diffed
+        Returns two lists: one with removals ('-') and one with additions ('+')
+        Ignores detail ('?') lines
+
+        In:  ['  Hi', '- there', '+ there!', '?      +\n']
+        Out: (['  Hi', '- there'], ['  Hi', '+ there!'])
+        """
```
allait:
```python
revision_1 = [w for w in words_diff if self._get_word_type(word) in ['unchanged', 'removal']]
revision_2 = [w for w in words_diff if self._get_word_type(word) in ['unchanged', 'addition']]

return revision_1, revision_2
```

app/main/views/services.py
```diff
@@ -105,10 +106,77 @@ def edit(service_id, section):
...
+                service_data_revision_1, service_data_revision_2):
+            raise ValueError
+
+        service_data = data_api_client.get_service(
+            service_data_revision_1['id'])['services']
+
+    except (HTTPError, KeyError, ValueError):
+        return abort(404)
+
+    content = service_content.get_builder().filter(service_data)
```
allait: This will list sections that the admin user can edit, not the ones that could've been changed by the supplier. The admin manifest is probably a superset of the supplier one, so this shouldn't be an issue, but I wonder if we should make a note of it somewhere.

PR #96, opened 2015-07-23
[#98947448] Fix diff tool bug.

pcraig3:
If one of the keys was non-existent, the diff page would break.

```python
# Would both throw a KeyErrors
revision_1 = { 'serviceName': "Paul's Service" }
revision_2 = {}

...

revision_1 = {}
revision_2 = { 'serviceFeatures': [ "Fixes your computer", "Makes you tea"] }
```
New update makes sure that the diff page handles the case where a key exists in one revision and not the other.

There's nothing in the interface differentiating between a new key being added and a new value being added, but I don't think that's actually a meaningful difference.

[>> Story on Pivotal](https://www.pivotaltracker.com/story/show/98947448)

***

#### Key (containing a list) exists in revision 2, not revision 1
screen shot 2015-07-23 at 09 45 43

####  Key (containing a string) exists in revision 1, not revision 2
screen shot 2015-07-23 at 09 46 18

***

##### Note: 
If a searched-for key doesn't exist in *either* revision, two empty values are assigned but they won't show up in the diff page because they're not different from each other.

### DIFF
--- a/app/main/helpers/diff_tools.py
+++ b/app/main/helpers/diff_tools.py
@@ -27,8 +27,8 @@ def get_diffs_from_service_data(
     for section in sections:
         for question in section['questions']:
             revisions_are_valid = True
-            question_revision_1 = revision_1.get(question['id'], None)
-            question_revision_2 = revision_2.get(question['id'], None)
+            question_revision_1 = revision_1.get(question['id'], [])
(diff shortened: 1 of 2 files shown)


quis:
:+1:

PR #98, opened 2015-07-27
Add Gulp task to copy over GOVUK toolkit images

tombye:
To answer a request for the gulpfile to be aligned across our front-end apps.

https://github.com/alphagov/digitalmarketplace-buyer-frontend/commit/fd86738d38240e825ba40011f678e0112a0473cc#commitcomment-12323019

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -190,6 +190,15 @@ gulp.task(
 );
 
 gulp.task(
+  'copy:govuk_toolkit_assets:images',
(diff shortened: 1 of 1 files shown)


tombye:
This is currently being broken by the issue fixed in https://github.com/alphagov/digitalmarketplace-admin-frontend/pull/99

(Merge that when it runs green and this should also run green.)

quis:
:+1:

PR #99, opened 2015-07-27
Fix order of arguments in forms macros

tombye:
The order of arguments in all the forms macros was causing this error in the tests:

```
File "app/templates/macros/forms.html", line 29, in template
    {% macro textbox(field, value='', service_data) -%}
TemplateSyntaxError: non-default argument follows default argument
```

### DIFF
--- a/app/templates/edit_section.html
+++ b/app/templates/edit_section.html
@@ -60,7 +60,7 @@
             {{ forms.validation_message(errors[question.id ], question.id) }}
           {% endif %}
 
-          {{ forms[question.type](question, service_data[question.id], service_data) }}
+          {{ forms[question.type](question, service_data, service_data[question.id]) }}
 
         {{ forms.question_wrapper_close(question) }}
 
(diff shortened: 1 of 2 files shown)


allait:
Can we add a default value to the `service_data` argument instead, since it's not really used by most macros. Also the `key, object, value` order is slightly unusual.

tombye:
I've re-ordered the arguments to make the order `key, value, object`.

I looked at leaving `service_data` out of the macros that didn't use it but this caused type errors in jinja. Looking into this it seems unnamed arguments sent in are only added to the `kwargs` and `varargs` variables if either of them are used in the macro (see these points in the source code: catch_kwargs: [L426](https://github.com/mitsuhiko/jinja2/blob/46acbf02ed9ab58c7a92553c95790e75bdc3c930/jinja2/runtime.py#L426), [L431](https://github.com/mitsuhiko/jinja2/blob/46acbf02ed9ab58c7a92553c95790e75bdc3c930/jinja2/runtime.py#L431)). Also see http://stackoverflow.com/questions/13944751/args-kwargs-in-jinja2-macros#answers-header for an example of this usage.

allait:
:+1:

PR #100, opened 2015-07-27
Add utils-based cache busting of assets

quis:
Brings in https://github.com/alphagov/digitalmarketplace-utils/pull/102

Fixes the _breaking changes_ introduced by: https://github.com/alphagov/digitalmarketplace-utils/pull/101


allait:
:+1:

PR #101, opened 2015-08-04
Invite supplier user

minglis:
Added invite user to suppliers functionality

- new email template: basically same as in suppler app with different user
- updates to users page to do form / form valuation for email to send
- uses updates to utils for sending emails
- similar functionality to supplier app for actual sending

Notes:
- Needs new beanstalk properties - Mandrill Key, Key and Salt for token generation
- MUST be in AWS before shipping
- Needs merge of Utils code (https://github.com/alphagov/digitalmarketplace-utils/pull/111)


app/main/views/errors.py
```diff
@@ -1,9 +1,9 @@
 from flask import render_template
```
robyoung: Why was this moved into `views`? It makes the layout different to the other frontend apps and they're a bit distinct from the other modules in the `views` package.

app/templates/view_supplier_users.html
```diff
@@ -5,129 +5,164 @@
 {% extends "_base_page.html" %}
 
 {% block page_title %}
-  {{ supplier.name }} – Digital Marketplace admin
```
robyoung: I think indentation has been mangled in this file.

minglis: pycharm hates the templates/html. Often trips me up. Should be a little better now.

app/main/views/errors.py
minglis: Sorry debugging an import issue. Moved it back.

PR #106, opened 2015-08-18
[#100210316] Add an audit event for inviting users

robyoung:
See: https://www.pivotaltracker.com/story/show/100210316

Log when users are invited by admin users as well. I have decided to not log anything specific to identify the action as coming from the admin application. This can be determined from the role of the user performing the action. It is very unlikely that roles would change from 'supplier' to 'admin'.


TheDoubleK:
:+1:

PR #107, opened 2015-08-21
Pass AuditType objects to data API client

robyoung:
Since 6.0.0 [1] the data API client expects AuditTypes instances.

[1]
https://github.com/alphagov/digitalmarketplace-utils/compare/5.3.1...6.0.0


allait:
:+1:

PR #108, opened 2015-08-24
Admin-ccs role

allait:
Part of [#101410132](https://www.pivotaltracker.com/story/show/101410132)

#### Add 403 error handler and 403 page template
Shows a message when a user tries to access a page/view that requires
a different role.

#### Add `role_required` view decorator to check current_user role
Returns 403 if user has none of the roles listed in the decorator
arguments.

#### Add role_required('admin') decorator to non-read-only views
Limits users with 'admin-ccs' role to read-only actions by wrapping
POST and form-rendering views with `role_required('admin')`

#### Hide buttons and links to restricted actions based on user role
Adds `current_user.has_role` checks to the templates to hide service
section "Edit" links; status change, audit event "Acknowledge" and
user "Activate"/"Deactivate"/"Unlock" buttons.

### DIFF
--- a/app/main/auth.py
+++ b/app/main/auth.py
@@ -1,0 +1,30 @@
+from functools import wraps
+
+from flask import abort
+from flask_login import current_user
+
+
+def role_required(*roles):
+    """Ensure that logged in user has one of the required roles.
+
+    Return 403 if the user doesn't have a required role.
+
+    Should be applied before the `@login_required` decorator:
+
+        @login_required
+        @role_required('admin', 'admin-ccs')
+        def view():
+            ...
+
+    """
+
+    def role_decorator(func):
+        @wraps(func)
+        def decorated_view(*args, **kwargs):
+            if not any(current_user.has_role(role) for role in roles):
+                return abort(403, "One of {} roles required".format(", ".join(roles)))
+            return func(*args, **kwargs)
+
+        return decorated_view
+
+    return role_decorator
--- a/app/templates/errors/403.html
+++ b/app/templates/errors/403.html
@@ -1,0 +1,14 @@
+{% extends "_base_page.html" %}
+
+{% block page_title %}Access denied - 403 – Digital Marketplace{% endblock %}
+
+{% block content %}
+  <div class="page-container">
+    {% with heading = "You don’t have permission to perform this action" %}
+      {% include 'toolkit/page-heading.html' %}
+    {% endwith %}
+    <p>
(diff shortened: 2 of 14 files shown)


app/main/auth.py
```diff
@@ -0,0 +1,30 @@
...
+
+def role_required(*roles):
+    """Ensure that logged in user has one of the required roles.
+
+    Return 403 if the user doesn't have a required role.
+
+    Should be applied before the `@login_required` decorator:
+
+        @login_required
+        @role_required('admin', 'admin-ccs')
```
TheDoubleK: I'm not sure about specifying 'CCS' in the role name - it might be nicer to have the role describe what permissions the role gives, rather than who the users with that role are, e.g. 'admin-readonly' or 'admin-view' (I guess we may have non-ccs people with this role in future?)

app/templates/errors/403.html
```diff
@@ -0,0 +1,14 @@
+{% extends "_base_page.html" %}
+
+{% block page_title %}Access denied - 403 – Digital Marketplace{% endblock %}
+
+{% block content %}
+  <div class="page-container">
+    {% with heading = "You don’t have permission to view this page" %}
```
TheDoubleK: This might be better as "you don't have permission to perform this action" - as the restricted users will hit this most often after trying to press buttons they shouldn't, rather than trying to view pages.

TheDoubleK:
On the "Find servcies by supplier ID" page the link to the service says "Edit" even though restricted users can't edit it - could maybe change the link text to "View details" or something similar.
screen shot 2015-08-25 at 10 52 18

app/templates/view_service.html
```diff
@@ -89,6 +89,7 @@
         {% endcall %}
       {% endfor %}
 
+      {% if current_user.has_role('admin') %}
```
TheDoubleK: The `compare_revisions.html` template could also have this check around the "Update status" button to hide it.

screen shot 2015-08-25 at 11 05 31

app/main/auth.py
allait: I'm not sure about the name either, but couldn't come up with anything better in the API PR. It's readonly right now, but that's more likely to change than the "CSS" part, which would make `admin-readonly` much more misleading.

allait:
Added missing role checks around revision status update and "Edit" link (replaced with "Details", which already had a condition attached) and changed the 403 heading to "... perform this action"

TheDoubleK:
:+1:

PR #109, opened 2015-08-24
Add G-Cloud 7 statistics page

quis:
This PR takes the data produced by the API in… alphagov/digitalmarketplace-api#211

…and puts it, tabulated, onto the page, so that we can know:
- services broken down by status
- services broken down by lot
- suppliers broken down by how far through the process they are
- users broken down by how recently they logged in

This is a first step towards:
- showing the data over time
- transforming the data into graphs
- making a 'big screen' view

--

image

### DIFF
--- a/app/main/__init__.py
+++ b/app/main/__init__.py
@@ -3,7 +3,7 @@ from flask import Blueprint
 
 main = Blueprint('main', __name__)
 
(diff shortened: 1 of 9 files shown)


allait:
This is ace, I really like `label_and_count`! :+1:

PR #110, opened 2015-08-25
[#101410262] Add search by supplier name

robyoung:
https://www.pivotaltracker.com/story/show/101410262

This is required to be able to find suppliers that do not yet have any services.

screen shot 2015-08-25 at 11 00 49


TheDoubleK:
:+1: looks good to me.

PR #111, opened 2015-08-25
Add big screen view for statistics page

quis:
This is something we can try out and see if it's useful. Will probably be better once we have graphs.

big-screen-2


allait:
:+1:

PR #112, opened 2015-08-27
Add JSON Apache access logs

robyoung:
This also changes the 4xx alarm to only be based on 400-403 errors.

### DIFF
--- a/.ebextensions/cwl-logs-apache-access.config
+++ b/.ebextensions/cwl-logs-apache-access.config
@@ -29,10 +29,14 @@ Mappings:
   CWLogs:
     AccessLogs:
       LogFile: "/var/log/httpd/cwl_access_log"
+      JSONLogFile: "/var/log/httpd/cwl_access_log.json"
       TimestampFormat: "%d/%b/%Y:%H:%M:%S %z"
       LogGroupName: {"Fn::GetOptionSetting": {"OptionName": "LogGroupName"}}
+      JSONLogGroupName: {"Fn::GetOptionSetting": {"OptionName": "JSONLogGroupName"}}
       ApplicationName: {"Fn::GetOptionSetting": {"OptionName": "ApplicationName"}}
     FilterPatterns:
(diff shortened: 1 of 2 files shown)


.ebextensions/cwl-logs-apache-access.config
```diff
@@ -29,10 +29,14 @@ Mappings:
   CWLogs:
     AccessLogs:
       LogFile: "/var/log/httpd/cwl_access_log"
+      JSONLogFile: "/var/log/httpd/cwl_access_log.json"
       TimestampFormat: "%d/%b/%Y:%H:%M:%S %z"
       LogGroupName: {"Fn::GetOptionSetting": {"OptionName": "LogGroupName"}}
+      JSONLogGroupName: {"Fn::GetOptionSetting": {"OptionName": "JSONLogGroupName"}}
       ApplicationName: {"Fn::GetOptionSetting": {"OptionName": "ApplicationName"}}
     FilterPatterns:
+      Http400To403MetricFilter: "{ $.logType = apache-access && $.application = api && $.status >= 400 && $.status <= 403 }"
```
TheDoubleK: We might also want to alert on 408 (Request Timeout) - could indicate networking problems?  And possibly also 405 (Method not allowed) - could indicate malicious users?

TheDoubleK:
It makes sense to exclude 404s from the alerts, but we might also want to alert on 408 (Request Timeout) - could indicate networking problems? And possibly also 405 (Method not allowed) - could indicate malicious users?

robyoung:
Yes, we should review whether these alarms are useful and what might be more useful. This change is bringing admin app in line with other applications. I'm not sure which way round makes more sense, making the alarms consistent or making them useful. It would be useful to have a chat about it if you've got a few minutes this afternoon?

minglis:
Wonder whether there's a couple of alerts? A 400/405 one for bad requests, a 401-403 one for failed logins and the like, 408 and associated could be something else?

Right now might be better to have them all in line, and in "tech debt" sprint fix 'em up?

TheDoubleK:
Sounds reasonable :+1:

PR #113, opened 2015-08-27
[#100305120] Remove local dates for dmutils dates

pcraig3:
Got rid of all in-app date formats and presentation logic for their equivalents in dmutils.
This means that our times are now displayed in B<sub>ritish</sub>S<sub>tandard</sub>T<sub>ime</sub> rather than UTC.

I also switched up a few `assertTrue`s for `assertIn`s because the error messages you get back are more helpful.

[>> Story on Pivotal](https://www.pivotaltracker.com/story/show/100305120)


quis:
:+1:

PR #114, opened 2015-08-27
Add placeholder text to service update search

pcraig3:
Was asked today how to search for audit events with specific dates in the admin app and I had to check to code to find out.
Adding placeholder text as a stopgap solution until we decide to overhaul this search page.


TheDoubleK:
:+1: 
+2 for the branch name.

minglis:
👍

quis:
Placeholder text is evil, but as a stopgap it should still go out.

PR #115, opened 2015-08-27
Update MI page to work with declaration status

quis:
Declaration status has changed from being true false to:

- `null` if declaration wasn't started (first page wasn't saved)
- `"started"` if declaration was started but not complete
- `"complete"` if declaration is complete

This commit doesn't expose this difference to the user (we can consider this later) but keeps the stats working.

### DIFF
--- a/app/main/helpers/sum_counts.py
+++ b/app/main/helpers/sum_counts.py
@@ -9,7 +9,7 @@ def _sum_counts(stats, filter_by=None, sum_by='count'):
     return sum(
         statistic[sum_by] for statistic in stats
         if not filter_by or all(
-            statistic.get(key) == value
+            statistic.get(key) in value if isinstance(value, list) else statistic.get(key) == value
             for key, value in filter_by.items()
         )
     )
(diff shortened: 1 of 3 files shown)


app/main/helpers/sum_counts.py
```diff
@@ -9,7 +9,7 @@ def _sum_counts(stats, filter_by=None, sum_by='count'):
     return sum(
         statistic[sum_by] for statistic in stats
         if not filter_by or all(
-            statistic.get(key) == value
+            statistic.get(key) in value if isinstance(value, list) else statistic.get(key) == value
```
TheDoubleK: I know this is probably "Pythonic", but it might be clearer to split this logic up a bit if possible? It's making my brain hurt.

app/main/views/stats.py
```diff
@@ -41,19 +41,19 @@ def view_statistics(framework_slug):
...
                 'has_completed_services': False
             },
             'completed_services_only': {
-                'has_made_declaration': False,
+                'declaration_status': [None, 'started'],
                 'has_completed_services': True
             },
             'valid_submission': {
-                'has_made_declaration': True,
+                'has_made_declaration': 'complete',
```
TheDoubleK: Should this one also be `declaration_status` rather than `has_made_declaration`?

quis:
@TheDoubleK Is 5f230b64c497f2b9984ccdbb8ba980222fa8800c any better?

TheDoubleK:
Marginally ;)
:+1:

PR #116, opened 2015-09-01
Display changes in framework statistics over time

quis:
47690-st60

# Use audit events rather than calling API directly

Previously the stats page called the API on every page load. It could only get the latest data.

This commit changes the stats page to get its data from audit events instead, where the stats are/will be snapshotted every hour, as per: alphagov/digitalmarketplace-api#222

TL;DR:
- less expensive API calls
- potential for showing change in data over time

# Get basic charts on the page

This commit creates charts using [C3](http://c3js.org/). Chosen because it is
- [MIT licensed](https://github.com/masayuki0812/c3/blob/master/LICENSE) (unline some other JS charting libraries)
- D3 based, which means the results look similar to Performance Platform

The code written in this commit extracts data from the tables on the statistics page and then passes it to C3 to generate the charts.

# Refine appearance and behaviour of charts

Including:
- some non-canon date formatting
- using a time series for the x axis, so that the points are properly spaced, even if they don't get captured at regular intervals
- using stacked area charts for most of the data (this seemed to make sense last time)
- some variations (colour, tooltip) depending on whether the page is being viewed in big screen mode or not

### DIFF
--- a/app/assets/javascripts/_tables-to-charts.js
+++ b/app/assets/javascripts/_tables-to-charts.js
@@ -2,16 +2,29 @@
 
   GOVUK.GDM.tablesToCharts = function() {
 
+    var isBigScreen = $("#framework-statistics-big-screen").length;
+
     $(".framework-statistics table")
       .each(function(tableIndex) {
 
-        var columns = [];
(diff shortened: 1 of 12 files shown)


allait:
:+1:

PR #119, opened 2015-09-02
Make big screen view clearer

quis:
screen shot 2015-09-02 at 07 37 31

**Add key to big screen view**

Because big screen view doesn't have tooltip, it's hard to identify which line is which in the charts.

This commit adds a colour-coded key to the table headings to relate the numbers to the lines more clearly.

--

**Relabel users table to be clearer**

### DIFF
--- a/app/templates/view_statistics.html
+++ b/app/templates/view_statistics.html
@@ -115,16 +115,16 @@
       {% endcall %}
     {% endcall %}
 
-    {{ summary.heading("Users") }}
+    {{ summary.heading("Users by last login time") }}
     {% call(item) summary.list_table(
       users,
       empty_message="Data not available",
       caption="Users",
       field_headings=[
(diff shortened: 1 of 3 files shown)


app/templates/view_statistics.html
```diff
@@ -115,16 +115,16 @@
...
     {% call(item) summary.list_table(
       users,
       empty_message="Data not available",
       caption="Users",
       field_headings=[
         summary.hidden_field_heading("Date and time"),
-        "Never logged in",
-        "Not logged in this week",
-        "Logged in this week",
+        "Before Python",
```
TheDoubleK: Will "Before Python" be meaningful to CCS?

quis: @TheDoubleK Changed to 'Before July' in 8808068

PR #120, opened 2015-09-02
Fix mixed-up lots on graphs

quis:
- don't confuse SaaS and PaaS
- match the order of the lots in the Digital Marketplace buyer frontend


TheDoubleK:
:+1:

PR #121, opened 2015-09-03
Allow up to 1260 historic data points on graphs

quis:
(35 days * 24 snapshots per day) * 50% for safety = 1260

Depends on
https://github.com/alphagov/digitalmarketplace-api/pull/237
https://github.com/alphagov/digitalmarketplace-utils/pull/156


allait:
:+1:

PR #122, opened 2015-09-03
Add basic prev/next navigation for service updates

quis:
--

screen shot 2015-09-03 at 17 13 35

- uses design pattern from the toolkit
- preserves filter options when navigating from page to page


TheDoubleK:
When clicking the "Acknowledge" button for an event on a page >1 it returns you to the first page.  Would be nice to have the redirect from the ack preserve the page number.

app/templates/service_update_audits.html
```diff
@@ -139,26 +139,19 @@ <h3 class="validation-masthead-heading" id="validation-masthead-heading">
...
-                        <a href="">
-                            Previous <span class="visuallyhidden">page</span>
-                            <span class="page-numbers">1 of 3</span>
-                        </a>
-                    </li>
-
-                    <li class="next">
-                        <a href="">
-                            Next <span class="visuallyhidden">page</span>
-                            <span class="page-numbers">2 of 3</span>
```
TheDoubleK: Do we mind not having the "Page 2 of 3" etc. any more?  Or did this never work in the first place anyway?

tests/app/main/views/test_service_updates.py
```diff
@@ -339,3 +343,22 @@ def test_should_call_api_ack_audit_event(self, data_api_client):
...
+
+        data_api_client.find_audit_events.assert_called_with(
+            page=5,
+            audit_type=AuditTypes.update_service,
+            acknowledged='false',
+            audit_date=None
+            )
+
+    @mock.patch('app.main.views.service_updates.data_api_client')
+    def test_should_not_pass_valid_page_argument_to_api(self, data_api_client):
```
TheDoubleK: Should this be `should_not_pass_invalid...`?

app/templates/service_update_audits.html
quis: I don't think we have any way of knowing how many pages there are in total. But we can definitely tell the user what the next/previous page is:

514e3bf

image

tests/app/main/views/test_service_updates.py
quis: Good point – 9405c95

quis:
@TheDoubleK The preserving page after acknowledging thing is tricky. It would then be possible to click the button and be redirected to a page which no longer has any events on it because you've shortened the list. Can you think of any way around that?

TheDoubleK:
Turns out I was just stalling so I wouldn't need to fix merge conflicts on my branch :trollface: 

Let's put this in as it is and we can fix the return-to-page issue if it becomes a problem for Nicole. 
:+1:

PR #124, opened 2015-09-07
Fix logging and metric filters

robyoung:
## Change application name in Apache logs
This should be admin-frontend rather than admin so that it gets picked up by the metric filters correctly.

## Remove metric filters for logs
These are now managed by dmaws

## Add login success tracking
This allows us to have rate based cloudwatch alarm.


TheDoubleK:
:+1: 
This is my kind of pull-request:
screen shot 2015-09-07 at 16 17 21

robyoung:
:grin:

PR #125, opened 2015-09-07
[#102567802] New role that can only view G-Gloud 7 stats

TheDoubleK:
See this story: https://www.pivotaltracker.com/story/show/102567802

This restricts access for users with role `admin-ccs-sourcing` to only the statistics page.  Their dashboard page when logging in looks like this:
screen shot 2015-09-07 at 15 39 07

This PR would depend on this going in first: https://github.com/alphagov/digitalmarketplace-api/pull/240 if we currently had any users with the now-deprecated role `admin-ccs`, but we don't - so either way round should be fine.

### DIFF
--- a/app/assets/scss/application.scss
+++ b/app/assets/scss/application.scss
@@ -8,18 +8,19 @@
 // Digital Marketplace front end toolkit components
 $path: "/admin/static/images/";
 @import "toolkit/breadcrumb";
-@import "toolkit/page-headings";
-@import "toolkit/forms/summary";
-@import "toolkit/forms/questions";
-@import "toolkit/forms/hint";
-@import "toolkit/forms/textboxes";
-@import "toolkit/forms/word-counter";
-@import "toolkit/forms/list-entry";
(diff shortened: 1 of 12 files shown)


quis:
I think we do have an `admin-ccs` user, the one we use for the big screen.

Also, inb4 @superroz, lowercase 's' in statistics.

quis:
I would like to see the role decorator [on that stats page](https://github.com/alphagov/digitalmarketplace-admin-frontend/blob/additional-ccs-view-stats-role/app/main/views/stats.py#L17) as well. Can't hurt to be explicit in case we add extra roles in the future who _shouldn't_ have access to the stats.

TheDoubleK:
Can't we use Title Case for Titles? Whatever :)

TheDoubleK:
Fixed up. There were a lot of endpoints with no roles specified before, but yes it makes sense to add to all of them now there's just that last one without any.  Added it to the dashboard page too for good measure.

quis:
:+1:

PR #126, opened 2015-09-14
Allow an admin user to change a supplier’s name

quis:
# Add a 'change name' link to the list of suppliers

image

# Add a new route, and a corresponding page with a form to enter a new supplier name

image

# Add a `post` endpoint for the new route which makes an API call to change the supplier’s name

image

--

There's no length limit on the new name. I think we can trust our admin users for now?

--

Completes [story 103088044](https://www.pivotaltracker.com/story/show/103088044).

### DIFF
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -25,6 +25,33 @@ def find_suppliers():
     )
 
 
+@main.route('/suppliers/<string:supplier_id>/edit/name', methods=['GET'])
+@login_required
+@role_required('admin', 'admin-ccs-category')
+def edit_supplier_name(supplier_id):
+
+    supplier = get_supplier(supplier_id)
+
+    return render_template(
+        "edit_supplier_name.html",
+        supplier=supplier["suppliers"],
+        **get_template_data()
+    )
+
+
+@main.route('/suppliers/<string:supplier_id>/edit/name', methods=['POST'])
+def update_supplier_name(supplier_id):
+
+    supplier = get_supplier(supplier_id)
+    new_supplier_name = request.form.get('new_supplier_name', '')
+
+    data_api_client.update_supplier(
+        supplier['suppliers']['id'], {'name': new_supplier_name}, current_user.email_address
+    )
+
+    return redirect(url_for('.find_suppliers', supplier_name_prefix=new_supplier_name[:1]))
+
+
 @main.route('/suppliers/users', methods=['GET'])
 @login_required
 @role_required('admin', 'admin-ccs-category')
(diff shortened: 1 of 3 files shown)


app/main/views/suppliers.py
```diff
@@ -25,6 +25,33 @@ def find_suppliers():
     )
 
 
+@main.route('/suppliers/<string:supplier_id>/edit/name', methods=['GET'])
+@login_required
+@role_required('admin', 'admin-ccs-category')
```
TheDoubleK: `admin-ccs-category` is a read-only role so shouldn't be able to do this.

app/main/views/suppliers.py
```diff
@@ -25,6 +25,33 @@ def find_suppliers():
...
+    supplier = get_supplier(supplier_id)
+
+    return render_template(
+        "edit_supplier_name.html",
+        supplier=supplier["suppliers"],
+        **get_template_data()
+    )
+
+
+@main.route('/suppliers/<string:supplier_id>/edit/name', methods=['POST'])
```
TheDoubleK: I think the `POST` also needs login and role_required decorators.

app/templates/edit_supplier_name.html
```diff
@@ -0,0 +1,45 @@
...
+      },
+      {
+        "label": supplier.name
+      }
+    ]
+  %}
+    {% include "toolkit/breadcrumb.html" %}
+  {% endwith %}
+  <div class="page-container">
+    {% with heading = "Change supplier name" %}
```
TheDoubleK: It might be nice to have the old supplier name shown somewhere other than in the breadcrumb - it took me a while to spot it there.

app/templates/view_suppliers.html
```diff
@@ -29,19 +29,18 @@
...
-
-      {{ summary.field_name(item.name) }}
-
-      {{ summary.service_link("Users", url_for(".find_supplier_users", supplier_id=item.id)) }}
-      {{ summary.service_link("Services", url_for(".find_supplier_services", supplier_id=item.id)) }}
-
-      {% endcall %}
+        {% call summary.row() %}
+          {{ summary.field_name(item.name) }}
+          {{ summary.edit_link("Change name", url_for(".edit_supplier_name", supplier_id=item.id)) }}
```
TheDoubleK: It would be good to only show the `Change name` to `admin` users - at the moment `admin-ccs-category` users can also see this page.  I think they should be able to browse it - and click through to see the list of users & services - but not to make any changes.

app/main/views/suppliers.py
quis: @TheDoubleK Well spotted—removed for debugging and restored in 55d7647 (which also addresses the `admin-ccs-category` thing

app/templates/edit_supplier_name.html
quis: It's shown inside the textbox…

app/templates/view_suppliers.html
quis: Good point. Addressed in 502f21b.

app/templates/edit_supplier_name.html
TheDoubleK: Ah yes, OK that's fine then.

TheDoubleK:
:+1:

PR #127, opened 2015-09-14
Adding search by email to the admin app

minglis:
- adds new search by email on index page
- new users controller to handle looking up by email
- re-use of suppliers-users table to show user
- links to supplier page for supplier users
- allows finding ALL user types
- Adds unlock/activate buttons
- uses a hidden field for the return link so can share the change status code and return to correct page.

### DIFF
--- a/app/main/views/users.py
+++ b/app/main/views/users.py
@@ -31,4 +31,3 @@ def find_user_by_email_address():
             users=list(),
             email_address=None,
             **get_template_data()), 404
-
--- a/app/main/views/users.py
+++ b/app/main/views/users.py
@@ -1,0 +1,34 @@
+from flask import render_template, request
+from flask_login import login_required, current_user, flash
+
+from .. import main
+from . import get_template_data
+from ... import data_api_client
+from ..auth import role_required
+
+
+@main.route('/users', methods=['GET'])
+@login_required
+@role_required('admin', 'admin-ccs-category')
+def find_user_by_email_address():
+    template = "view_users.html"
+    users = None
+
+    email_address = request.args.get("email_address", None)
+    if email_address:
+        users = data_api_client.get_user(email_address=request.args.get("email_address"))
+
+    if users:
+        return render_template(
+            template,
+            users=[users['users']],
+            email_address=request.args.get("email_address"),
+            **get_template_data())
+    else:
+        flash('no_users', 'error')
+        return render_template(
+            template,
+            users=list(),
+            email_address=None,
+            **get_template_data()), 404
+
--- a/app/templates/view_users.html
+++ b/app/templates/view_users.html
@@ -63,7 +63,7 @@
 
         {{ summary.field_name(item.name) }}
 
-        {{ summary.field_name(item.role) }}
+        {{ summary.text(item.role) }}
 
         {% call summary.field() %}
             {% if item.role == 'supplier' %}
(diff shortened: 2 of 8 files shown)


TheDoubleK:
The `Role` column seems too wide here - would be better to have `Supplier` as the wide column.
screen shot 2015-09-14 at 14 16 39

app/main/views/users.py
```diff
@@ -0,0 +1,33 @@
...
+@main.route('/users', methods=['GET'])
+@login_required
+@role_required('admin', 'admin-ccs-category')
+def find_user_by_email_address():
+    template = "view_users.html"
+    users = None
+
+    email_address = request.args.get("email_address", None)
+    if email_address:
+        users = data_api_client.get_user(email_address=request.args.get("email_address"))
```
TheDoubleK: Being picky, this could be singular `user` because there's only ever one user.

app/templates/view_users.html
```diff
@@ -0,0 +1,163 @@
...
+              {% endif %}
+            {% endif %}
+        {% endcall %}
+
+        {% endcall %}
+        {% endcall %}
+    <form action="{{ url_for('.find_user_by_email_address') }}" method="get" class="question">
+      <label class="question-heading" for="email_address">Find users by email address</label>
+      <p class='hint'>
+        eg john@smokeyjoes.com
```
TheDoubleK: screen shot 2015-09-14 at 14 35 23

app/templates/view_users.html
```diff
@@ -0,0 +1,163 @@
...
+                'Locked',
+                summary.hidden_field_heading("Change status")
+            ],
+            field_headings_visible=True)
+        %}
+        {% call summary.row() %}
+
+        {{ summary.field_name(item.name) }}
+
+        {{ summary.field_name(item.role) }}
```
TheDoubleK: In terms of the wide `Role` column, I think it's because it's getting `class=summary-item-field-first` (as does `name`) - but it shouldn't have the `first` on the end.  I have no idea how you'd change this though.

app/templates/view_users.html
TheDoubleK: Alternately: https://www.youtube.com/watch?v=fH_oTTWSm1M

quis: > As described in RFC 2606 and RFC 6761, a number of domains such as example.com and example.org are maintained for documentation purposes.

http://www.iana.org/domains/reserved

tests/app/main/views/test_users.py
```diff
@@ -0,0 +1,171 @@
...
+            '//input[@class="button-secondary"]')[0].attrib['value']
+        unlock_link = document.xpath(
+            '//tr[@class="summary-item-row"]//td/span/form')[0]
+        return_link = document.xpath(
+            '//tr[@class="summary-item-row"]//td/span/form/input')[1]
+        self.assertEquals('/admin/suppliers/users/999/unlock', unlock_link.attrib['action'])
+        self.assertEquals('Unlock', unlock_button)
+        self.assertEquals('/admin/users?email_address=test.user%40sme.com', return_link.attrib['value'])
+
+    def test_should_show_deactivate_button(self, data_api_client):
```
TheDoubleK: Would it be overkill to test that the button isn't shown to `admin-ccs-category` users? (Given that we don't currently have anyone with that role it might be).

TheDoubleK:
Very minor comments.  Tested and all works - good to go if we don't care about these small things.

app/templates/view_users.html
quis: You'd wanna change `summary.field_name(…)` to `summary.text(…)`.

quis:
Checked this out and tried it locally. Looks alright to me. I addressed a couple of little things in 4bb8d6e and 341c1fe and would be in favour of getting this deployed now :+1:

PR #128, opened 2015-09-14
Button added to remove users from suppliers

minglis:
- on supplier user page there is now a 'remove from supplier' button
- this will change the role of that user to buyer
- redirects back to the page

Requires this API update:
- [x] https://github.com/alphagov/digitalmarketplace-api/pull/244/files


TheDoubleK:
Should admin also be able to add users to a supplier without going through the send-invitation route?

I'm thinking of an "Undo" for this really.

TheDoubleK:
Tested and works :+1: once the required API change is in.

PR #129, opened 2015-09-22
Add grids to form content

tombye:
Wraps all form content in a two-thirds grid so their width is dictated by the grid instead of part of their styling.

Part of https://www.pivotaltracker.com/story/show/103093496

Brings in this work from the DM Toolkit: https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/158

I've also done some work to make the use of grids a bit more uniform.


app/assets/javascripts/application.js
```diff
@@ -22,6 +22,14 @@
 
   var module;
 
+  if(typeof console === 'undefined') {
```
quis: If we're going to put this in every app, could it be in the toolkit?

app/assets/scss/views/_service_updates.scss
```diff
@@ -0,0 +1,19 @@
+.govuk-option-select {
```
quis: Why isn't this most/all of this not going into the toolkit pattern?

tombye: That's a good point. I took these styles from the buyers app so they are shared so suitable for pushing to the toolkit.

I'll raise a pull request to that end (which will involve moving the buyers app to a toolkit version too) though I don't see why it should hold up this one.

app/assets/javascripts/application.js
tombye: That depends on whether the module loader should go in because it's the only JavaScript that uses the `console` object. Thoughts?

quis: Should we standardise on having the module loader in every app?

tombye:
Closing this until the new styles can be brought in from the DM Toolkit

tombye:
Re-opening now the styles are in.

app/assets/javascripts/application.js
tombye: I've made a pull request to add it into the DM Toolkit (see https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/164).

I'll raise another pull request to make this app use that version once merged.

app/assets/scss/views/_service_updates.scss
tombye: I've made a pull request which adds this text field into the DM Toolkit (see https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/163).

I'll raise another pull request to make this app use that version once merged.

quis:
Just waiting for the integration of:
- https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/163
- https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/164
and I'm happy with this

PR #130, opened 2015-09-24
Update to latest utils version

TheDoubleK:
Update methods no longer take a `reason` parameter.

It's not desperate to make this change now, but I think this is the only place other than the supplier app (updated here: https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/293) that still passes a reason in to updates, so makes sense to tidy it all up at the same time.

Depends on:
- [x] https://github.com/alphagov/digitalmarketplace-utils/pull/166/files


quis:
:+1: when the version is ready.

PR #132, opened 2015-10-05
Fix view service table columns

robyoung:
There was a miss-copy when moving this over from the supplier frontend.
The field macros already call `summary.field` so wrapping the call in another `summary.field` creates invalid HTML that browsers interpret as another column.


allait:
:+1:

PR #133, opened 2015-10-06
Use response headers to log requestId from apache

allait:
Request headers are only set on requests to the API from the frontend
apps, but response headers are set by a WSGI middleware for all
responses.


robyoung:
:+1:

PR #134, opened 2015-10-08
[#103325898] Strip leading and trailing whitespace from form fields.

pcraig3:
I've subclassed the WTF `StringField` class with one that preprocesses all user-inputted values.  Specifically, all submitted values will be stripped of leading and trailing whitespace.

What happens if I'm trying to login with my valid email and password:

Submitted values | Processed values | Result
------------ | ------------ | -------------
`'email@email.com'`, `'password'` | `'email@email.com'`, `'password'` | :white_check_mark:
`' email@email.com '`, `'password'` | `'email@email.com'`, `'password'` | :white_check_mark:
`'email@email.com'`, `' password '` | `'email@email.com'`, `' password '` | :x:
`' email@email.com '`, `' password '` | `'email@email.com'`, `' password '` | :x:

Also used when inviting a user.

[>> Story in Pivotal](https://www.pivotaltracker.com/story/show/103325898)

-

I haven't replaced the StringField in the `ServiceUpdateAuditEventsForm` because that one gets its input from a set of radio buttons, and if you're monkeying around with the `value` attribute in the HTML, then you shouldn't be.

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -4,6 +4,7 @@ from wtforms.validators import Optional, AnyOf, DataRequired, \
     Email
 from datetime import datetime
 from dmutils.formats import DATE_FORMAT
+from dmutils.forms import StripWhitespaceStringField
 
 
 class ServiceUpdateAuditEventsForm(Form):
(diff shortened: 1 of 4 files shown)


robyoung:
:+1:

PR #136, opened 2015-10-22
Lowercase lot updates

allait:
#### Show full lot name in service list tables
Instead of showing slug abbreviation display the full lot name
in service lists.

#### Update framework stats page with lowercase lots


app/templates/view_statistics.html
```diff
@@ -83,10 +83,10 @@
         {% call summary.field(first=True) %}
           {{ item.created_at[:19]|replace('T', ' ')|replace('-', '/') }}
         {% endcall %}
-        {{ summary.text(item.SaaS) }}
-        {{ summary.text(item.PaaS) }}
-        {{ summary.text(item.IaaS) }}
-        {{ summary.text(item.SCS) }}
+        {{ summary.text(item.saas) }}
+        {{ summary.text(item.paap) }}
```
quis: Platform as a Platform?

app/main/views/stats.py
```diff
@@ -51,10 +51,10 @@ def view_statistics(framework_slug):
...
         }),
         services_by_lot=format_snapshots(snapshots, 'services', {
-            'IaaS': {'lot': 'IaaS'},
-            'PaaS': {'lot': 'PaaS'},
-            'SaaS': {'lot': 'SaaS'},
-            'SCS':  {'lot': 'SCS'},
+            'iaas': {'lot': 'iaas'},
+            'paas': {'lot': 'paas'},
+            'saas': {'lot': 'saas'},
+            'scs':  {'lot': 'scs'},
```
quis: I think you'd need to make the comparison case-insensitive [here](https://github.com/alphagov/digitalmarketplace-admin-frontend/blob/lowercase-lot-updates/app/main/helpers/sum_counts.py#L22) for this to work, because the snapshots will still contain the camelCase lots.

Not sure if we care about the historic G7 data at this point.

quis: image

app/templates/view_statistics.html
allait: Oops :sweat_smile: Fixed now.

allait:
@quis added both lot slugs to `services_by_lot` (turns out it's much easier than I thought it would be thanks to the list check, `format_snapshots` is pretty neat!)

I had to clear and migrate my local DB so don't have any graphs, could you check that it fixes it for snapshots?

quis:
:+1: :shipit:

PR #137, opened 2015-10-22
Watch for changes to content

quis:
Implements https://github.com/alphagov/digitalmarketplace-utils/pull/193

Adds https://pypi.python.org/pypi/watchdog as a requirement to make watching for file changes use system events, which is faster

### DIFF
--- a/application.py
+++ b/application.py
@@ -3,12 +3,13 @@
 import os
 from app import create_app
 from flask.ext.script import Manager, Server
+from dmutils import init_manager
 
 application = create_app(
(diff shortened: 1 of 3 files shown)


quis:
# Have flask init init the manager
Depends on and implements alphagov/digitalmarketplace-utils#195

quis:
# Bring in 11.1.1 of utils

To allow us to use the version of `init_manager` which returns a `Manager` instance.

robyoung:
:+1:

PR #142, opened 2015-11-07
Only get the latest stats while framework is open

quis:
To keep the statistics fresh, we append the latest numbers to the table every time the page is loaded.

These numbers won’t change after close of submissions, so we should only do it when the framework is in an 'open' state.

image

### DIFF
--- a/app/main/views/stats.py
+++ b/app/main/views/stats.py
@@ -25,7 +25,7 @@ def view_statistics(framework_slug):
             per_page=1260
         )['auditEvents']
         framework = data_api_client.get_framework(framework_slug)['frameworks']
-        if framework['status'] is 'open':
+        if framework['status'] == 'open':
             snapshots.append(
                 {
                     'data': data_api_client.get_framework_stats(framework_slug),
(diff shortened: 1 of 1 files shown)


app/main/views/stats.py
```diff
@@ -24,13 +24,14 @@ def view_statistics(framework_slug):
...
         )['auditEvents']
-        latest = data_api_client.get_framework_stats(framework_slug)
-        snapshots.append(
-            {
-                'data': latest,
-                'createdAt': datetime.utcnow().strftime(DATETIME_FORMAT)
-            }
-        )
+        framework = data_api_client.get_framework(framework_slug)['frameworks']
+        if framework['status'] is 'open':
```
TheDoubleK: We should use `==` for comparing values of things (strings) - `is` tests for identity and seems like it is only generally used for `is None`.  See http://stackoverflow.com/questions/1504717/why-does-comparing-strings-in-python-using-either-or-is-sometimes-produce

Otherwise good to go I think.

allait:
:+1:

PR #143, opened 2015-11-10
Update content.

pcraig3:
Update [digitalmarketplace-frameworks](https://github.com/alphagov/digitalmarketplace-frameworks) content, and update the admin app so it's consistent with the new changes.

The only thing that's changed is that `maxLengthInWords` has now become `max_length_in_words`, so pretty small pull request.


quis:
:+1:

PR #144, opened 2015-11-11
Fix bug in agreements list

robyoung:
The slug key in the framework response is 'slug'
The slug key in the supplier framework response is 'frameworkSlug'


TheDoubleK:
:+1:

PR #145, opened 2015-11-17
Multi-extension agreement download

robyoung:
This change adds support for framework agreements potentially having different extensions introduced in https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/340

Because the full path is no longer known, instead the prefix is provided and the most recent matching file is returned. This works because the agreement file names do not overlap and the timestamp is added to the start of the archive file name. If the supplier first uploads a PDF and then uploads a JPEG the JPEG will be returned here because it is the most recent matching upload.

### DIFF
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -72,7 +72,11 @@ def download_agreement_file(supplier_id, framework_slug, document_name):
         abort(404)
     legal_supplier_name = supplier_framework['declaration']['SQ1-1a']
     agreements_bucket = s3.S3(current_app.config['DM_AGREEMENTS_BUCKET'])
-    path = get_agreement_document_path(framework_slug, supplier_id, legal_supplier_name, document_name)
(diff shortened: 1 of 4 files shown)


app/main/views/suppliers.py
```diff
@@ -72,7 +72,11 @@ def download_agreement_file(supplier_id, framework_slug, document_name):
         abort(404)
     legal_supplier_name = supplier_framework['declaration']['SQ1-1a']
     agreements_bucket = s3.S3(current_app.config['DM_AGREEMENTS_BUCKET'])
-    path = get_agreement_document_path(framework_slug, supplier_id, legal_supplier_name, document_name)
+    prefix = get_agreement_document_path(framework_slug, supplier_id, legal_supplier_name, document_name)
+    agreement_documents = agreements_bucket.list(prefix=prefix)
+    if not len(agreement_documents):
```
TheDoubleK: Could this just be `if not agreement_documents:`?

TheDoubleK:
Minor pedantic comment that is not essential to fix; otherwise looks good :+1:

PR #148, opened 2015-11-26
Add coverage and requirements

quis:


allait:
Admin app is still using `nosetests`, I don't think it recognises `--cov` arguments. We could try replacing it with `py.test` in `scripts/run_tests.py` to see if that works out of the box.

quis:
@allait edc8e22 
> Also switches from Nosetests to Pytest, because Nose doesn’t recognise `--cov` arguments.

quis:
94.098% :tada:

TheDoubleK:
Nice coverage.

allait:
:+1:

PR #151, opened 2015-12-04
Limit stats audit events to current framework

robyoung:
When loading the stored stats from previous days on the stats page we need to limit it to the current framework.


TheDoubleK:
Looks like it will work :+1:

robyoung:
I've tested it locally against the prod API.

PR #152, opened 2015-12-07
Make stats page lot-agnostic

quis:
Previous this was hard-coded to the G-Cloud lots.

image

Would appreciate someone testing this with their local data and seeing if it looks reasonable…


TheDoubleK:
We could add a link for DOS stats to the front page:
screen shot 2015-12-07 at 14 37 19

TheDoubleK:
I don't get a graph line at all running it locally, but my local data might be crappy?
screen shot 2015-12-07 at 14 38 45

allait:
:+1:

PR #153, opened 2015-12-07
Only count complete services in lot breakdown

quis:
Because you automatically get a draft just for clicking into ‘outcomes’, ‘specialists’ or ‘research participants’ the numbers aren’t very meaningful.

I think it’s better to only count services that suppliers have marked as complete.


ralph-hawkins:
Yeah, agree

carolinegreen:
That does make more sense.

allait:
:+1:

PR #154, opened 2015-12-09
[#109804168] Count suppliers who have started their declaration as "interested"

TheDoubleK:
See: https://www.pivotaltracker.com/story/show/109804168

Suppliers who had started their declaration but not completed any services were not being shown in any of the supplier counts on the the statistics dashboard.

This includes them in the count of "interested only" suppliers.


quis:
Nice work tracking this down. :+1:.

PR #155, opened 2015-12-17
Enable admin to post new files

tommorris:
Pivotal: [#107188452](https://www.pivotaltracker.com/story/show/107188452)

As an admin user I want to upload new clarification answers in pdf, new ITT packs in zip format and their accompanying email in PDFs from the admin app so that I don't have to get a developer involved each time.

### DIFF
--- a/app/main/views/communications.py
+++ b/app/main/views/communications.py
@@ -78,7 +78,6 @@ def upload_communication(framework_slug):
             flash('itt_pack', 'upload_communication')
 
     if len(errors) > 0:
-        print errors
         for category, message in errors.items():
             flash(category, message)
     return redirect(url_for('.manage_communications', framework_slug=framework_slug))
--- a/app/main/views/communications.py
+++ b/app/main/views/communications.py
@@ -1,0 +1,84 @@
+from flask import render_template, redirect, url_for, current_app, \
+    request, flash
+from flask_login import login_required
+
+from .. import main
+from ..auth import role_required
+from ... import data_api_client
+from . import get_template_data
+
+from dmutils import s3
+from dmutils.documents import file_is_pdf, file_is_zip
+
+
+def _get_path(framework_slug, path):
+    return '{}/communications/{}'.format(framework_slug, path)
+
+
+def _get_itt_pack_path(framework_slug):
+    return '{0}/communications/{0}-itt-pack.zip'.format(framework_slug)
+
+
+@main.route('/communications/<framework_slug>', methods=['GET'])
+@login_required
+@role_required('admin')
+def manage_communications(framework_slug):
+    communications_bucket = s3.S3(current_app.config['DM_COMMUNICATIONS_BUCKET'])
+    framework = data_api_client.get_framework(framework_slug)['frameworks']
+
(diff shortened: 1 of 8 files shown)


app/templates/manage_communications.html
```diff
@@ -0,0 +1,106 @@
...
+    ]
+%}
+{% include "toolkit/breadcrumb.html" %}
+{% endwith %}
+
+<div class="page-container">
+{% with messages = get_flashed_messages(with_categories=true) %}
+{% if messages %}
+{% for category, message in messages %}
+{% if category == 'upload_communication' %}
```
quis: This should use [the notification banner template](http://alphagov.github.io/digitalmarketplace-frontend-toolkit/notification-banner.html) from the toolkit, rather than being hard-coded HTML.

app/templates/manage_communications.html
```diff
@@ -0,0 +1,106 @@
...
+    {% if message == 'itt_pack' %}
+      ITT pack was uploaded.
+    {% elif message == 'clarification' %}
+      New clarification was uploaded.
+    {% elif message == 'communication' %}
+      New communication was uploaded.
+    {% endif %}
+    </p>
+  </div>
+{% elif category == 'not_pdf' %}
```
quis: These `not_*` messages should use [the validation masthead pattern](http://alphagov.github.io/digitalmarketplace-frontend-toolkit/forms/validation.html), linking to the appropriate form controls below.

tommorris: @quis fixed!

app/templates/manage_communications.html
tommorris: @quis: fixed!

quis:
Plus one for the front end bits.

robyoung:
I don't think that the filetype checking should be done here, it should be done in [`dmutils`](https://github.com/alphagov/digitalmarketplace-utils/blob/master/dmutils/documents.py) so that we can be consistent with how we do this between applications. Also, while checking the filetype by it's contents is useful we should be using existing libraries like [python-magic](https://github.com/ahupp/python-magic) rather than doing it ourselves as it's notoriously finicky. I think we talked about not doing it for this story and instead tackling it in it's own right separately.

tommorris:
@robyoung I can get rid of it if you'd prefer, but it seemed easier to chuck it in and then replace it at some point in the future.

app/main/views/communications.py
```diff
@@ -0,0 +1,84 @@
...
+
+@main.route('/communications/<framework_slug>', methods=['POST'])
+@login_required
+@role_required('admin')
+def upload_communication(framework_slug):
+    communications_bucket = s3.S3(current_app.config['DM_COMMUNICATIONS_BUCKET'])
+    framework = data_api_client.get_framework(framework_slug)['frameworks']
+    errors = {}
+
+    if 'communication' in request.files and request.files['communication'].filename is not u'':
```
robyoung: This could be `if request.files.get('communication')` as the `FileStorage` object evaluates to `False` if there is no file.

robyoung:
Needs a rebase but looks good :+1:

tommorris:
@robyoung done.

PR #156, opened 2015-12-18
Update bower to stop it complaining

tombye:
When it complains, it tells you to update the
global install. This assumes your installation is
global which ours isn't.

image

This message is therefore dangerous to follow.

Following it and updating bower is sensible anyway.


robyoung:
:+1:

PR #157, opened 2015-12-21
Use same layout container as supplier + buyer apps

quis:
https://www.pivotaltracker.com/story/show/110394970

> The admin app is the only one which uses a different layout system to generate its pages.
>
> For example, each page on the supplier and buyer apps wrap their content in a `<div>` with id of wrapper'. The admin app instead uses a `<div>` with class of `page-container`.
>
> This reduces the amount of shared CSS we can use on the admin app and will cause problems, the more modules we use from the toolkit.

This commit fixes this.

### DIFF
--- a/app/assets/scss/_layout.scss
+++ b/app/assets/scss/_layout.scss
@@ -5,7 +5,7 @@
   background: $red;
 }
 
-.page-container {
+#wrapper {
   @extend %site-width-container;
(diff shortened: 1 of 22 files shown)


quis:
Sorry for the                                                               whitespace changes

tombye:
:+1:

PR #158, opened 2015-12-22
Make index links explicit again

robyoung:
In an earlier commit I grouped together links on the admin index page and made the title's just the framework name. In hindsight this was a bad idea because it means that the links don't fully describe what they are going to. This is apparent in the functional tests.


quis:
:+1:

PR #159, opened 2015-12-22
Upgrade frameworks and toolkit

robyoung:
This brings in the latest frontend toolkit and frameworks to make the pricing work again. This was broken by an earlier update to the utils package. This is a massive jump in versions but the only place that should be affected is the service editing and that has been tested.


pcraig3:
Noticed two things: one being a functional problem that's not related to your pull request and the other being a layout thing.

##### 1. Bucket names

Initially, I wasn't able to do any editing because the bucket name doesn't follow the same pattern as other bucket names.  Our default bucket name is [`admin-frontend-dev-documents`](https://github.com/alphagov/digitalmarketplace-admin-frontend/blob/master/scripts/run_app.sh#L13) (as set in the `run_app.sh` script), which doesn't gel with the logic that generates a [`bucket_short_name`](https://github.com/alphagov/digitalmarketplace-utils/blob/master/dmutils/s3.py#L31).  I get that it's not really to do with this story, but it *does* need fixing.

##### 2. Layout thing

More straightforward.  Looks like the column width is smaller and the hint text doesn't have the right colour.
*(Maybe this is part of @quis's story?)*

###### Pricing in supplier app
screen shot 2015-12-22 at 11 00 47

###### Pricing in admin app
screen shot 2015-12-22 at 11 01 16

pcraig3:
:+1: to fixing the layout problem in the near future.

PR #160, opened 2015-12-22
Make index links explicit again

robyoung:
In an earlier commit I grouped together links on the admin index page and made the title's just the framework name. In hindsight this was a bad idea because it means that the links don't fully describe what they are going to. This is apparent in the functional tests.

This is a copy of https://github.com/alphagov/digitalmarketplace-admin-frontend/pull/158 which got clobbered by https://github.com/alphagov/digitalmarketplace-admin-frontend/pull/157


pcraig3:
:+1:

PR #161, opened 2015-12-23
Page for changes to service status

quis:
_Rebased on top of https://github.com/alphagov/digitalmarketplace-admin-frontend/pull/164. Changed introduced in this PR: https://github.com/alphagov/digitalmarketplace-admin-frontend/compare/add-dmapiclient-package...service-status-updates_

***

~~This is kinda hacky but I want to see if it’s the right approach before cleaning it up.~~

Shows a page of audit events for change of service status. One page per day. Pagination to move to the next/previous days that have events.

I think this is easier to understand than the weird date input filter thing.

screen shot 2015-12-23 at 15 25 49

screen shot 2015-12-23 at 15 25 33

### DIFF
--- a/app/main/views/service_updates.py
+++ b/app/main/views/service_updates.py
@@ -1,6 +1,4 @@
-import re
-from datetime import datetime
-from ordered_set import OrderedSet
+from datetime import datetime, timedelta
 
 from flask import request, render_template, redirect, url_for, abort
 from flask_login import login_required, current_user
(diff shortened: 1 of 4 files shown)


pcraig3:
It's not really clear what the old status used to be, but it's also not really clear if that's anything worth knowing.

Way better than the other "Service Updates" layout though.

quis:
## Make pagination work for days with lots of events …

If a day had lots of events they would need to be shown across multiple pages (because the API, currently, only returns 100 events).

This modifies the pagination to work for both days and pages within days. So the journey now looks like:

Wednesday → Thursday → Thursday page 2  → Thursday page 3 → Friday

It does mean that you can’t skip days with no events when paginating, but this would have fallen apart on days with many events in the previous version anyway.

quis:
This will work better with https://github.com/alphagov/digitalmarketplace-api/pull/315 (but it’s not a dealbreaker).

quis:
## Change description of status to public/unavailable

In discussion with @cathrooney we’ve decided that the statuses should be:
- ‘unavailable’ if either an admin or a supplier has removed/unpublished/turned off a service
- ‘public’ if the service has been published/made available/turned on

***

image

quis:
## Map both "enabled" and "disabled" to "removed"

The two states are now just:
  - live
  - removed

quis:
## Show status updates by latest first

Implements: alphagov/digitalmarketplace-api#315

app/templates/_proposition_header.html
```diff
@@ -10,7 +10,10 @@
       <ul id="proposition-links">
         {% if current_user.has_role('admin') or current_user.has_role('admin-ccs-category') %}
         <li>
-            <a href="{{ url_for('main.service_update_audits') }}">Service Updates</a>
+            <a href="{{ url_for('main.service_update_audits') }}">Service updates</a>
+        </li>
+        <li>
+            <a href="{{ url_for('main.service_status_update_audits') }}">Service status changes</a>
```
TheDoubleK: I'm not sure what qualifies a page as getting it's own top-level navigation rather than another item in the huge list of things on the main admin page?  This is fine here as far as I'm concerned, but we should take a proper look at what goes where in the Admin app some time soon.

TheDoubleK:
Tested locally and works fine.

I like the added ability to see into the future... do we care about this?
screen shot 2016-01-06 at 14 47 13

app/templates/service_status_update_audits.html
```diff
@@ -0,0 +1,83 @@
...
+
+  {%
+    with
+    previous_page = {
+      "url": url_for(".service_status_update_audits", day=day, page=page-1),
+      "title": "Page {}".format(page - 1),
+      "label": "of {}".format(day_as_datetime|dateformat)
+    } if previous_page else {
+      "url": url_for(".service_status_update_audits", day=next_day.strftime('%Y-%m-%d')),
+      "title": "Next day",
```
TheDoubleK: I would probably have put the arrows the other way round here: `Next day` to the right and `Previous day` to the left.

quis:
@TheDoubleK 

Agree on the navigation. Needs looking at. For this PR I was just following the convention set by the service updates page.

Fixed the future in 820edb3

Conventionally ‘next’ goes into the past, eg from the blog:
image

TheDoubleK:
But that's only because blogs are designed to be read arse-upwards - if I'm reading a book then the "next page" goes into the future. But anyway, it's fine.

:+1: after a rebase.

app/main/views/service_updates.py
```diff
@@ -13,6 +13,44 @@
 from . import get_template_data
 
 
+@main.route('/service-status-updates', methods=['GET'])
+@main.route('/service-status-updates/<day>', methods=['GET'])
+@main.route('/service-status-updates/<day>/page-<int:page>', methods=['GET'])
```
allait: I know I'm very late to this, but why are day and page not query parameters? That's what we use for services and it would also allow pagination across dates.

PR #163, opened 2015-12-29
Add reset password link to login page

pcraig3:
Copied the markup from the supplier template.

[>> Story on Pivotal](https://www.pivotaltracker.com/story/show/110492418)

***

screen shot 2015-12-29 at 11 33 37

### DIFF
--- a/app/templates/login.html
+++ b/app/templates/login.html
@@ -74,13 +74,18 @@
                 </div>
             {% endif %}
 
-            <button class="button-save">Log in</button>
+            {% with type = "save", label = "Log in" %}
+              {% include "toolkit/button.html" %}
+            {% endwith %}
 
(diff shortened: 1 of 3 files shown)


app/templates/login.html
```diff
@@ -75,6 +75,12 @@
             {% endif %}
 
             <button class="button-save">Log in</button>
+
+            <header class="page-heading-smaller">
```
allait: `<header>` doesn't really fit here, we probably should be using something else both here and in the supplier app

quis:
cc @allait 

### Use 'secondary action link' pattern from DM toolkit
Implements alphagov/digitalmarketplace-frontend-toolkit#209

allait:
:+1:

PR #167, opened 2016-01-12
Remove a few lines of unneeded code

pcraig3:
Since we're no longer checking the status of the desired framework, we don't need to get the framework back from the API.


pcraig3:
The stuff I removed doesn't get tested. :/

allait:
:+1:

PR #168, opened 2016-01-14
[#111504612] Fix service link on status list

robyoung:
https://www.pivotaltracker.com/story/show/111504612


tombye:
Looks good to me :+1:

PR #169, opened 2016-01-14
Fix the uploaded supplier pack file name

allait:
supplier-frontend expects the supplier pack file name to be
`<slug>-supplier-pack.zip`, which doesn't match the file name
generated by the admin app on upload.

https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/master/app/main/views/frameworks.py#L86


TheDoubleK:
:+1:

PR #171, opened 2016-01-18
Don't allow viewing supplier declaration for open frameworks

allait:
Supplier declarations shouldn't be viewed by anyone while the
framework is open for submissions.


TheDoubleK:
Nice catch - looks good.
I'm not sure why the coveralls still hasn't gone green - it looks like it ran already?
:+1: when the merge button will work.

allait:
Looks like coveralls is lagging a bit (http://status.coveralls.io). Merging without it.

PR #174, opened 2016-01-20
Upload countersigned agreements

carolinegreen:
Add configuration for agreements buckets.  List latest countersigned agreement, and add ability to upload or remove countersigned agreement.

Version bump for digitalmarketplace-apiclient.

Countersigned agreement
screen shot 2016-01-21 at 16 50 45

Successful upload 
screen shot 2016-01-21 at 16 54 48

Unsuccessful upload
screen shot 2016-01-21 at 16 56 09

Confirmation screen for file deletion
screen shot 2016-01-21 at 16 58 01

No agreements have been uploaded
screen shot 2016-01-21 at 16 59 27

### DIFF
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -1,5 +1,6 @@
 from flask import render_template, request, redirect, url_for, abort, current_app
 from flask_login import login_required, current_user, flash
+from dateutil.parser import parse as parse_date
 
 from .. import main
 from . import get_template_data
@@ -9,8 +10,10 @@ from ..auth import role_required
 from dmapiclient import HTTPError, APIError
 from dmapiclient.audit import AuditTypes
 from dmutils.email import send_email, generate_token, MandrillException
-from dmutils.documents import get_signed_url, get_agreement_document_path
+from dmutils.documents import get_signed_url, get_agreement_document_path, file_is_pdf, \
+    get_countersigned_agreement_document_path
 from dmutils import s3
+from dmutils.formats import datetimeformat
 
 
 @main.route('/suppliers', methods=['GET'])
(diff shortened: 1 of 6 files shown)


robyoung:
Could we get some screenshots of how this looks?

carolinegreen:
Yes! Forgot about that!

robyoung:
Also, it would be good to make a note of the fact that it depends on https://github.com/alphagov/digitalmarketplace-api/pull/330

app/main/views/suppliers.py
```diff
@@ -86,6 +88,119 @@ def download_agreement_file(supplier_id, framework_slug, document_name):
...
+    agreements_bucket = s3.S3(current_app.config['DM_AGREEMENTS_BUCKET'])
+    errors = {}
+
+    if request.files.get('countersigned_agreement'):
+        the_file = request.files['countersigned_agreement']
+        if not file_is_pdf(the_file):
+            errors['countersigned_agreement'] = 'not_pdf'
+
+        if 'countersigned_agreement' not in errors.keys():
+            filename = '{}/agreements/{}/{}-countersigned-framework-agreement.pdf'.format(
```
robyoung: This could use the [`get_countersigned_agreement_document_path`](https://github.com/alphagov/digitalmarketplace-utils/blob/master/dmutils/documents.py#L207) function.

app/main/views/suppliers.py
```diff
@@ -86,6 +88,119 @@ def download_agreement_file(supplier_id, framework_slug, document_name):
     return redirect(url)
 
 
+@main.route('/suppliers/<supplier_id>/countersigned-agreements/<framework_slug>/'
+            '<document_name>', methods=['GET'])
+@login_required
+@role_required('admin-ccs-sourcing')
+def list_countersigned_agreement_file(supplier_id, framework_slug, document_name):
```
robyoung: I think this will only ever show either zero or one countersigned agreements. I think that the table is a good way of displaying it but there seems to be logic in here and in download countersigned agreement file that assumes there can be multiple.

app/main/views/suppliers.py
```diff
@@ -86,6 +88,119 @@ def download_agreement_file(supplier_id, framework_slug, document_name):
...
+
+
+@main.route('/suppliers/<supplier_id>/countersigned-agreements_download/<framework_slug>/'
+            '<document_name>', methods=['GET'])
+@login_required
+@role_required('admin-ccs-sourcing')
+def download_countersigned_agreement_file(supplier_id, framework_slug, document_name):
+    agreements_bucket = s3.S3(current_app.config['DM_AGREEMENTS_BUCKET'])
+    agreements_bucket.list('{}/agreements/{}'.format(
+        framework_slug, supplier_id))
```
robyoung: I think this is a testing leftover, the list isn't put anywhere.

app/main/views/suppliers.py
```diff
@@ -86,6 +88,119 @@ def download_agreement_file(supplier_id, framework_slug, document_name):
...
+@role_required('admin-ccs-sourcing')
+def remove_countersigned_agreement_file(supplier_id, framework_slug, document_name):
+    agreements_bucket = s3.S3(current_app.config['DM_AGREEMENTS_BUCKET'])
+    document = '{}/agreements/{}/{}-{}.pdf'.format(framework_slug, supplier_id, supplier_id, document_name)
+
+    if request.method == 'GET':
+        flash('countersigned_agreement', 'remove_countersigned_agreement')
+
+    if request.method == 'POST':
+        document_deleted = agreements_bucket.bucket.delete_key(key_name=document)
```
robyoung: Do we need to keep a record of the signed agreements? If so, I think we should add a method to [`dmutils.s3.S3`](https://github.com/alphagov/digitalmarketplace-utils/blob/master/dmutils/s3.py#L23) to delete that does something like `S3._move_existing`.

app/main/views/suppliers.py
carolinegreen: There are multiple saved, but I think that the requirement was that only the latest one was shown.  That's why I'm not showing all of them.

app/main/views/suppliers.py
carolinegreen: Well a copy is already kept as it works the same way as the agreements.

app/templates/suppliers/upload_countersigned_agreement.html
```diff
@@ -0,0 +1,113 @@
...
+    {%
+      with
+      type = "save",
+      label = "Upload file"
+    %}
+      {% include "toolkit/button.html" %}
+    {% endwith %}
+  </form>
+
+{% endblock %}
```
robyoung: Brrp, needs a newline.

app/main/views/suppliers.py
robyoung: But isn't the copy made of the file that was there beforehand? I don't think  that there is another copy of the file being deleted here.

app/main/views/suppliers.py
robyoung: They are there for archiving, so that we can manually roll back if someone really messes up. I think for the purpose of this interface we can probably assume there is only one. I'm not 100% sure on this though, it would be worth speaking to @cathrooney to see if they need to see the history of the file. In any case, the prefix we have here won't match those archive files as they start with the timestamp.

app/main/views/suppliers.py
carolinegreen: No there isn't, the timestamped copy is kept.

app/main/views/suppliers.py
carolinegreen: That appeared to be the way the agreements bucket was working though.  For example in the method download an agreement file, which I didn't write it uses the list method to access one file.

app/main/views/suppliers.py
allait: I think Rob's point was that the timestamped copy is a copy of the file being replaced by the new version. So the most recent file doesn't have a backup and if it's removed without creating one then there's no copy left. At least that's how I remember it working.

I'm also not sure about using `.bucket` attributes outside the `S3` class. It exposes the boto interface directly and it'd be better to wrap everything we need into a helper method.

app/main/views/suppliers.py
```diff
@@ -86,6 +89,118 @@ def download_agreement_file(supplier_id, framework_slug, document_name):
...
+@login_required
+@role_required('admin-ccs-sourcing')
+def list_countersigned_agreement_file(supplier_id, framework_slug, document_name):
+    supplier = data_api_client.get_supplier(supplier_id)['suppliers']
+    framework = data_api_client.get_framework(framework_slug)['frameworks']
+    agreements_bucket = s3.S3(current_app.config['DM_AGREEMENTS_BUCKET'])
+    path = get_countersigned_agreement_document_path(framework_slug, supplier_id)
+    countersigned_agreement_document = agreements_bucket.get_key(path)
+    if countersigned_agreement_document:
+        print('countersigned_agreement_document', countersigned_agreement_document)
```
robyoung: Here and line 107 have stray print statements

app/main/views/suppliers.py
```diff
@@ -86,6 +89,116 @@ def download_agreement_file(supplier_id, framework_slug, document_name):
     return redirect(url)
 
 
+@main.route('/suppliers/<supplier_id>/countersigned-agreements/<framework_slug>/'
+            '<document_name>', methods=['GET'])
```
pcraig3: Is the `document_name` necessary?  Seems like it's hardcoded everywhere and since the route already contains `countersigned-agreements` and the framework slug, it might be okay to lose it.

app/main/views/suppliers.py
```diff
@@ -86,6 +89,116 @@ def download_agreement_file(supplier_id, framework_slug, document_name):
...
+
+    return redirect(url_for(
+        '.list_countersigned_agreement_file',
+        supplier_id=supplier_id,
+        framework_slug=framework_slug,
+        document_name=document_name)
+    )
+
+
+@main.route('/suppliers/<supplier_id>/countersigned-agreements_download/<framework_slug>/'
```
pcraig3: Not sure if we use underscores in any of our other urls. Could they be dashes instead?

app/main/views/suppliers.py
```diff
@@ -86,6 +89,116 @@ def download_agreement_file(supplier_id, framework_slug, document_name):
...
+
+@main.route('/suppliers/<supplier_id>/countersigned-agreements_remove/<framework_slug>/'
+            '<document_name>', methods=['GET', 'POST'])
+@login_required
+@role_required('admin-ccs-sourcing')
+def remove_countersigned_agreement_file(supplier_id, framework_slug, document_name):
+    agreements_bucket = s3.S3(current_app.config['DM_AGREEMENTS_BUCKET'])
+    document = get_countersigned_agreement_document_path(framework_slug, supplier_id)
+
+    if request.method == 'GET':
```
pcraig3: This seems like a really good way of doing the 'are you sure' confirmation thing.

app/main/views/suppliers.py
carolinegreen: Yeah that was a leftover not removed.

app/main/views/suppliers.py
carolinegreen: Yeah shouldn't be that.

pcraig3:
Looks good :+1:

PR #175, opened 2016-02-04
Add environment variables for communications and agreements buckets

tombye:
The admin app uses them but they were absent until now.


robyoung:
:+1:

PR #176, opened 2016-02-05
Remove supplier name from agreement paths

allait:
Updates utils to use new agreement path helpers.
Removes a separate endpoint for downloading countersigned agreements:
now that the paths for signed and countersigned agreements have the
same structure there's no need for a separate view function.


pcraig3:
:+1: Looks good!

PR #177, opened 2016-02-05
Fix buyer app service links

robyoung:
The previous path was deprecated and has recently been removed from the buyer app.

:sparkles: This has been tested locally with the functional tests :sparkles:


TheDoubleK:
Looks great :+1: when Travis pulls his finger out.

PR #178, opened 2016-02-05
Add link to DOS agreements page

robyoung:
The page already exists and can be viewed for any framework so the simplest possible solution is to just add a link to the admin home page.

## Before
screen shot 2016-02-05 at 14 58 41

## After
screen shot 2016-02-05 at 14 58 56

### DIFF
--- a/app/templates/index.html
+++ b/app/templates/index.html
@@ -119,9 +119,14 @@
     {% endwith %}
     {% with items = [
         {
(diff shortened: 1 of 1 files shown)


TheDoubleK:
At some point the admin interface needs a bit of love, but in terms of what we have right now I think this just grand and does everything needed :+1:

robyoung:
This has had the :+1: from @cathrooney too

PR #179, opened 2016-02-08
Fail on error exit status in build

robyoung:


allait:
:+1:, coveralls is being weird again

PR #180, opened 2016-02-12
Standardise errorhandler names to match other apps

pcraig3:
There were a few errorhandler names in the admin app that were wrong, based on precedents set in the other frontend apps. 

Now they're not.

### DIFF
--- a/app/main/errors.py
+++ b/app/main/errors.py
@@ -4,7 +4,7 @@ from dmapiclient import APIError
 
 
 @main.app_errorhandler(APIError)
-def api_error(e):
+def api_error_handler(e):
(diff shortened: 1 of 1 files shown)


pcraig3:
> Coverage decreased (-0.0%) to 89.828%

TheDoubleK:
If you were feeling really conscientious you *could* add tests for the 400 and 500 error handlers (these are the ones missing tests).  But it's not like there's any logic in them, so :+1: as is if you have better things to do.

pcraig3:
No logic in (doing) them right now, so I'm gonna merge it.
Still an improvement.

PR #181, opened 2016-02-12
Introduce Makefile

robyoung:
This follows on from the changes introcuded in https://github.com/alphagov/digitalmarketplace-api/pull/340

- Move environment default config values into config.py
- Introduce Makefile
- Replace run_app.sh and run_tests.sh with appropriate make calls


allait:
:+1:

PR #182, opened 2016-02-16
[#113436157] Add 'Admin home' breadcrumb link to pages without it

pcraig3:
This pull request fixes a bug whereby a couple of pages didn't have breadcrumb links back to the Admin home page.

Spent a while thinking about a smarter way to do this, but couldn't figure it out without adding a bunch of logic to a "breadcrumb-building" template or some other form of indirection likely to age badly.  So copy+pasted it in.

[>> Story on Pivotal](https://www.pivotaltracker.com/story/show/113436157)


robyoung:
:+1:

PR #183, opened 2016-02-17
Remove obsolete macros from admin app

pcraig3:
While [adding "Admin home" breadcrumbs to a few of the pages missing it](https://github.com/alphagov/digitalmarketplace-admin-frontend/pull/182), I noticed that a few pages were still pulling in the `breadcrumb.html` macro in their headers, but not doing anything with it.
A similar thing was happening with the `page_heading.html` macro.  

These macros have since been replaced by their equivalents in the frontend toolkit, so I've:
- removed the imports
- swapped over templates still using the macros to using the toolkit
- deleted the macro files

Took a bit of a shortcut with some of the markup in the `compare-revisions.html` template, but I figured it was better to do that rather than (a) keep a bunch of old code or (b) try to more correctly overhaul the markup in a template no one's using.

***

#### Compare revisions (before)
screen shot 2016-02-17 at 11 55 27

#### Compare revisions (after)
screen shot 2016-02-17 at 11 51 48


TheDoubleK:
Could use the a "temporary message" element from the toolkit if we wanted to keep the blue box, but I think it looks fine with the link.

TheDoubleK:
:+1:

pcraig3:
The blue links are dead, Kev.  Now it's all about "View service page on the Digital Marketplace".

screen shot 2016-02-18 at 10 14 04

(taken from the edit service view in the supplier app.)

PR #185, opened 2016-02-29
Fix logging errors

allait:
### Update dmutils to 17.3.3 to fix logging exceptions 
Fixes log formatting errors in JSON logger

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ install:
   - npm install
   - npm run frontend-build:production
 script:
-  - ./scripts/run_tests.sh --cov=app --cov-report=term-missing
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -9,7 +9,7 @@ install:
   - npm install
   - npm run frontend-build:production
 script:
-  - ./scripts/run_tests.sh --cov=app --cov-report=term-missing
+  - PYTEST_ARGS='--cov=app --cov-report=term-missing' make test
```
TheDoubleK: Do these not need the `"@$" argument too?

allait: This is the script run by travis, I don't think there are any additional arguments passed to it, so this should be fine.

TheDoubleK:
:+1:

PR #186, opened 2016-03-02
Don't inject the template data explicitly

TheDoubleK:
This is handled by `init_app` in utils now.

Pages look right locally and tests pass, so I believe this works.


TheDoubleK:
Coveralls is lying.

How can coverage drop when the only change is to remove a method but not remove any tests? There is one less thing to be tested but the same amount of tests, so coverage can only have gone up.

allait:
:+1:

PR #189, opened 2016-03-08
Use the latest dmutils.

pcraig3:
Updated to the new dmutils and fixed the calls to `.get_error_messages()` that would have otherwise been broken.


TheDoubleK:
:+1:

PR #193, opened 2016-05-09
Change supplier link following user search

TheDoubleK:
Fixes this bug: https://www.pivotaltracker.com/story/show/118839501

When searching for a user by email address it is more convenient for admin users to go to a page where they can access both services and user accounts for that supplier. (Currently the link takes them to the page with users for the supplier.)

I have demo'd this to Nicole and it made her happy :)


allait:
👍

PR #194, opened 2016-05-10
Change to using the new dmcontent package

Wynndow:
The content loader has been extracted from dmutils to its own pacakge, `dmcontent`.

This pull request updates the admin-frontend to use the new package.

It updates the requirements to import the new package, as well as a new version of `dmutils` which no longer includes the moved content loader.

It also slightly updates three tests to drop leading 0s for dates, as version 20.0.0 of `dmutils` does not use them.


allait:
👍

PR #195, opened 2016-05-10
Allow admin to move users between supplier accounts

TheDoubleK:
Fixes this bug: https://www.pivotaltracker.com/story/show/119279671

Currently the "Remove user from supplier" button doesn't work, because before we had actual buyer accounts users were ported to a "buyer" role, which now fails unless the user happened to have an email address that is allowed for our buyer accounts, and anyway is not the expected behaviour any more.

This removes the broken "Remove from supplier" button, and instead adds a form field that allows an admin user to move existing supplier users from one supplier to another.

:sparkles:  Demo'd to Nicole and she likes it. :sparkles:

~~It's probably going to break the functional tests though - fix for those coming soon.~~
Functional test fix should be merged just *before* this: 
 - [ ] https://github.com/alphagov/digitalmarketplace-functional-tests/pull/195

screen shot 2016-05-10 at 17 51 46

### DIFF
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -4,7 +4,7 @@ from dateutil.parser import parse as parse_date
 
 from .. import main
 from ... import data_api_client, content_loader
-from ..forms import EmailAddressForm
+from ..forms import EmailAddressForm, MoveUserForm
 from ..auth import role_required
 from dmapiclient import HTTPError, APIError
 from dmapiclient.audit import AuditTypes
@@ -268,6 +268,7 @@ def update_supplier_name(supplier_id):
 @role_required('admin', 'admin-ccs-category')
 def find_supplier_users():
     form = EmailAddressForm()
+    form2 = MoveUserForm()
 
     if not request.args.get('supplier_id'):
         abort(404)
@@ -279,6 +280,7 @@ def find_supplier_users():
         "view_supplier_users.html",
         users=users["users"],
         form=form,
+        form2=form2,
         supplier=supplier["suppliers"]
     )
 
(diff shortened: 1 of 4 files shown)


app/main/views/suppliers.py
```diff
@@ -268,6 +268,7 @@ def update_supplier_name(supplier_id):
 @role_required('admin', 'admin-ccs-category')
 def find_supplier_users():
     form = EmailAddressForm()
+    form2 = MoveUserForm()
```
pcraig3: Do we care about the variable names for the forms?
<sub>I don't.</sub>

app/main/views/suppliers.py
```diff
@@ -313,12 +315,50 @@ def deactivate_user(user_id):
...
+                updater=current_user.email_address
+            )
+            flash("user_moved", "success")
+        else:
+            flash("user_not_moved", "error")
+        return redirect(url_for('.find_supplier_users', supplier_id=supplier_id))
+    else:
+        return render_template(
+            "view_supplier_users.html",
+            form=form,
```
pcraig3: I don't know if we ever do this, but if you just initialized the form here (ie, `form=EmailAddressForm()`), then it would be clear that we don't use it in this route at all.  
By initializing it at the beginning, it seems like it might be being used in the view somewhere.

app/templates/view_supplier_users.html
```diff
@@ -176,7 +162,7 @@
                       <div class="question">
                           {{ form.email_address.label(class="question-heading-with-hint") }}
                           <p class="hint">
-                              Enter the email address you of the user you wish to add
+                              Enter the email address you of the person you wish to invite
```
pcraig3: Nice.

app/main/views/suppliers.py
TheDoubleK: <sub>I don't either</sub>

TheDoubleK: <sub><sub>...but I'm going to fix it anyway</sub></sub>

app/main/views/suppliers.py
TheDoubleK: Yes that's much nicer - fixed.

pcraig3:
👍

TheDoubleK:
Thanks - please can I get a thumbs-up on the functional test change too and will merge together?

PR #197, opened 2016-06-13
Bump govuk template

tombye:
Bumps the GOVUK template and fixes the problem this causes (see https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/338).

https://www.pivotaltracker.com/story/show/120364875


pcraig3:
```
      ------
    < legit! >
      ------
        / 
       / 
```
✊  😃  👍

PR #200, opened 2016-08-03
Admin edit address and duns

TheDoubleK:
Fixes this bug with declaration editing in the admin app: https://www.pivotaltracker.com/story/show/126406697

The first problem was that the frameworks version hadn't been bumped for an age, so the app didn't know about the DUNs number question at all and thought the address question was a text field when it has been split into a three-part multiquestion.

Once the new content was pulled in I needed to add a form macro for editing multiquestions.  This is simpler than the ones used in other apps, as we don't need to show question hints, etc. for admin editing.

Finally, the summary page wasn't quite doing things right in terms of the content loader - it was rendering questions from the content with values pulled directly from the declaration object without calling `.summary` which ties together questions with the answers given. Adding that call meant that the address multiquestion shows up OK in the summary page.

Summary table with address and DUNs number displayed
------------------------------------------------------------------
screen shot 2016-08-03 at 17 38 16

New form fields on the edit page
-------------------------------------
screen shot 2016-08-03 at 17 38 25

screen shot 2016-08-03 at 17 38 33

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -5,7 +5,7 @@
     "jquery": "1.11.2",
     "digitalmarketplace-frontend-toolkit": "https://github.com/alphagov/digitalmarketplace-frontend-toolkit.git#v15.13.1",
     "govuk_template": "https://github.com/alphagov/govuk_template/releases/download/v0.17.3/jinja_govuk_template-0.17.3.tgz",
-    "digitalmarketplace-frameworks": "https://github.com/alphagov/digitalmarketplace-frameworks.git#v1.1.0",
+    "digitalmarketplace-frameworks": "https://github.com/alphagov/digitalmarketplace-frameworks.git#v2.1.2",
     "hogan": "3.0.2",
     "c3": "0.4.10"
   }
--- a/app/main/views/services.py
+++ b/app/main/views/services.py
@@ -49,7 +49,7 @@ def view(service_id):
         return redirect(url_for('.index'))
 
     service_data['priceString'] = format_service_price(service_data)
-    content = content_loader.get_builder('g-cloud-6', 'edit_service_as_admin').filter(service_data)
+    content = content_loader.get_manifest('g-cloud-6', 'edit_service_as_admin').filter(service_data)
 
     return render_template(
         "view_service.html",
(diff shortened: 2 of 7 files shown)


TheDoubleK:
screen shot 2016-08-03 at 17 44 36
:100: :100:

bower.json
```diff
@@ -5,7 +5,7 @@
     "jquery": "1.11.2",
     "digitalmarketplace-frontend-toolkit": "https://github.com/alphagov/digitalmarketplace-frontend-toolkit.git#v15.13.1",
     "govuk_template": "https://github.com/alphagov/govuk_template/releases/download/v0.17.3/jinja_govuk_template-0.17.3.tgz",
-    "digitalmarketplace-frameworks": "https://github.com/alphagov/digitalmarketplace-frameworks.git#v1.1.0",
+    "digitalmarketplace-frameworks": "https://github.com/alphagov/digitalmarketplace-frameworks.git#v2.1.2",
```
idavidmcdonald: This bump for the frameworks requires the latest version of the content loader (whilst this change you're making wouldn't break anything, this requirement isn't really clear though. It might be worth us adding a change log to the frameworks repo to tell people to do this rather than hope they realise).

idavidmcdonald:
image
Is the "[[tradingStatus]]" in your screenshot anything to worry about?

bower.json
TheDoubleK: Nice spot - I should have noticed the major version change.

idavidmcdonald:
image

Do we want to have "Building and street" etc in here? Surely the user just wants to see the plain address without the subtitles (which are not easy to differentiate at the moment either). Don't know if this is existing behaviour that has previously been OK'ed or not?

TheDoubleK:
The `[[tradingStatus]]` isn't something to worry about, in that it's how it is at the moment. It isn't ideal, but admin should know what it refers to.

We don't use the logic of question numbering in the admin app that we have in the submissions part of the supplier app, so there are no question numbers to cross-reference here (which is what the `[[ ]]` items are).

TheDoubleK:
It's using the toolkit pattern we have for displaying multiquestions in summary tables.  For addresses it might make sense not to show what the fields are, but in the general case for multiquestions this won't work - e.g. if it just displayed "10% 20% 70%" or "Yes No Yes Yes No" as values without field names it would be pretty unhelpful as a summary.

I don't think we should make a special case for address.

idavidmcdonald:
👍

PR #201, opened 2016-08-05
Admin can extract buyer contact details and their briefs

Wynndow:
For [this](https://www.pivotaltracker.com/story/show/124104161) story on Pivotal.

Admin's need to be able to extract a list of buyers with their contact details and their briefs.

This PR adds a link to the download user list page which hooks up to a new endpoint on the api. Api PR is [here](https://github.com/alphagov/digitalmarketplace-api/pull/430) and apiclient PR is [here](https://github.com/alphagov/digitalmarketplace-apiclient/pull/32)

Apiclient PR will need merging for tests to go green.

download_buyer_csv

PR #202, opened 2016-08-09
Add link for admin's to download a csv of buyers details and their briefs

Wynndow:
Part of [this](https://www.pivotaltracker.com/story/show/124104161) story on Pivotal.

This PR supersedes [this](https://github.com/alphagov/digitalmarketplace-admin-frontend/pull/201) PR for the same story.

For [this](https://www.pivotaltracker.com/story/show/124104161) story on Pivotal.

Admin's need to be able to extract a list of buyers with their contact details and their briefs.

This PR adds a link to the download user list page which hits the api at /briefs and /users and combines the results into a csv for download.. Api PR is [here](https://github.com/alphagov/digitalmarketplace-api/pull/431) and apiclient PR is [here](https://github.com/alphagov/digitalmarketplace-apiclient/pull/33)

Apiclient PR will need merging for tests to go green.

download_buyer_csv


app/main/helpers/csv_generator.py
```diff
@@ -0,0 +1,22 @@
+import unicodecsv
+
+
+def iter_csv(rows, headers):
```
idavidmcdonald: Good work on moving this into here. 

You'll notice we have a very similar function in the [buyer frontend](https://github.com/alphagov/digitalmarketplace-buyer-frontend/blob/5b1f1640166b13f47fc8feacb52bc4a09c677302/app/buyers/views/buyers.py#L317). The question now is should this be moved into dm-utils? I'd probably say no for the moment just so we don't creep the scope of this story but maybe something for the future.

app/main/views/users.py
```diff
@@ -65,32 +66,46 @@ def download_users(framework_slug):
...
-            yield line.read()
 
+@main.route('/users/download/buyers', methods=['GET'])
+@login_required
+@role_required('admin')
+def download_buyers_and_briefs():
+    buyers_list = [user for user in data_api_client.find_users_iter(role='buyer')]
+    briefs_list = [brief for brief in data_api_client.find_briefs_iter(with_users=True)]
+
+    for buyer in buyers_list:
```
idavidmcdonald: With all these nested for loops, do we have any idea if performance will be OK (if it is even a concern)?

tests/app/main/views/test_users.py
```diff
@@ -283,3 +284,155 @@ def test_download_csv(self, data_api_client):
...
+                }]
+            }
+        ]
+
+        response = self.client.get('/admin/users/download/buyers')
+        rows = [line.split(",") for line in response.get_data(as_text=True).splitlines()]
+        buyer = rows[1]
+
+        assert response.status_code == 200
+        assert buyer == [u'Chris', u'chris@gov.uk', u'01234567891',
```
idavidmcdonald: Could be good to assert the values of the header row too (probably just in one of your tests, no need for overkill).

tests/app/main/views/test_users.py
```diff
@@ -283,3 +284,155 @@ def test_download_csv(self, data_api_client):
...
+                }]
+            }
+        ]
+
+        response = self.client.get('/admin/users/download/buyers')
+
+        assert response.mimetype == 'text/csv'
+
+    def test_filename_includes_a_timestamp(self, data_api_client):
+        with mock.patch('app.main.views.users.datetime') as mock_date:
```
idavidmcdonald: I think an alternative here would be to use `freeze_time`. You can see an example [here](https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/0578507db1d2cfb23b55f177f4d74f1de69bd3ce/tests/app/main/test_services.py#L798). I think your current solution is still good though but good to be aware of freeze_time.

app/main/helpers/csv_generator.py
Wynndow: I feel like I should, but i'm not sure what best practice is....

app/main/views/users.py
Wynndow: Yeah, it could take a while! I'm not sure how to test it though with the quantity of data we have on production.

It's not going to get used very often, and only by our admin's, so i don't think it being super quick is a massive issue.

I'm going to look into how to stress test it.

Wynndow: I've run the function with 10'000 users and 10'000 briefs in a test environment and it took just under 40 seconds. I don't know how this stacks up agains the numbers on production but I suspect it's ok?

tests/app/main/views/test_users.py
Wynndow: Nice! Will remember that one!

tests/app/main/views/test_users.py
Wynndow: Good call, have included.

app/main/views/users.py
```diff
@@ -65,32 +66,61 @@ def download_users(framework_slug):
...
+        "emailAddress",
+        "phoneNumber",
+        "createdAt",
+        "briefs"
+    ]
+
+    buyers_dict = {}
+    for buyer in buyers:
+        buyer['briefs'] = []
+        buyers_dict[buyer['id']] = buyer
```
pcraig3: 💯  💯 

This is much nicer than the double `for` loop that was here before.

app/main/views/users.py
```diff
@@ -65,32 +66,61 @@ def download_users(framework_slug):
...
+
+    for brief in briefs:
+        for user in brief['users']:
+            brief_string = '{} - {}'.format(brief['title'], brief['status'])
```
pcraig3: Is this all that we want to know about the briefs themselves?  There's no `id` needed?

pcraig3:
Also, branch naming skillz: 💯

app/main/views/users.py
Wynndow: Yeah, Alexey showed me the light. Also I learnt a lot about generators which was nice.

app/main/views/users.py
Wynndow: Yup. Story just asks for what you see here. Sarah seems happy with it.

pcraig3:
I think this is 👍  to go.

PR #203, opened 2016-08-22
Show briefs as open instead of live in csv

Wynndow:
Part of [this](https://www.pivotaltracker.com/story/show/124104161) story on Pivotal.

Briefs were being listed in the csv with their status as whatever their
status in the database is set at. To be more consistent with the rest of
the digitalmarketplace, if a brief's status is live it should be listed
as open in the csv.


TheDoubleK:
:+1:

PR #204, opened 2016-08-23
[#127807225] Add `variations_agreed` to framework users csv download

TheDoubleK:
Completes this story: https://www.pivotaltracker.com/story/show/127807225

Depends on API change here: 
 - [ ] https://github.com/alphagov/digitalmarketplace-api/pull/436

### DIFF
--- a/tests/app/main/views/test_users.py
+++ b/tests/app/main/views/test_users.py
@@ -263,7 +263,8 @@ class TestUsersExport(LoggedInApplicationTest):
             "framework_agreement": False,
             "supplier_id": 1,
             "user_email": "test.user@sme.com",
-            "user_name": "Tess User"
+            "user_name": "Test User",
+            "variations_agreed": "var1"
         }]
 
         response = self._return_user_export_response(data_api_client, framework, users)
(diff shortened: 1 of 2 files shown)


tests/app/main/views/test_users.py
```diff
@@ -245,6 +245,8 @@ def _assert_things_about_user_export(self, response, users):
             assert sorted(list(users[0].keys())) == sorted(rows[0])
 
             for index, user in enumerate(users):
+                print("LHS: {}".format(sorted([six.text_type(val) for val in user.values()])))
```
Wynndow: Do these prints serve any purpose? They don't assert anything and they don't print out in my tests if run locally?

TheDoubleK: No that was some debug stuff I forgot to remove - they do print out if the test fails, I'll drop them now.

TheDoubleK: Needed a rebase anyway so I squashed this into previous commit.

Wynndow:
Looks good to me! 👍

PR #205, opened 2016-08-23
Import api-client 6.3.3

Wynndow:
There was an issue with requests to the api from iter methods changing
the http protocol from http to https. The update on the apiclient fixes
it.


allait:
👍

PR #206, opened 2016-08-25
[#127982249]agreement listing: use new template for g-cloud-8 based on mockup

risicle:
Story https://www.pivotaltracker.com/story/show/127982249

### DIFF
--- a/app/templates/_view_agreements_results.html
+++ b/app/templates/_view_agreements_results.html
@@ -1,0 +1,12 @@
+{% for supplier_framework in supplier_frameworks %}
+<div class="search-result">
+  <h2 class="search-result-title">
+    <a href="{{ url_for('.view_signed_agreement', supplier_id=supplier_framework.supplierId, framework_slug=supplier_framework.frameworkSlug) }}">{{ supplier_framework.supplierName }}</a>
+  </h2>
+  <ul class="search-result-important-metadata">
+    <li class="search-result-metadata-item">
+      Submitted: {{ supplier_framework.agreementReturnedAt }}
+    </li>
+  </ul>
+</div>
+{% endfor %}
--- a/app/templates/_view_agreements_summary.html
+++ b/app/templates/_view_agreements_summary.html
@@ -1,0 +1,8 @@
+{% if supplier_frameworks %}
+{% set content %}
+  <em class="search-summary-count">{{ supplier_frameworks|length }}</em>
+  {{ pluralize(supplier_frameworks|length, "agreement", "agreements") }}
+  <em>waiting for countersigning</em>
+{% endset %}
+{% include "toolkit/search-summary.html" %}
+{% endif %}
--- a/app/templates/view_agreements_g8.html
+++ b/app/templates/view_agreements_g8.html
@@ -36,13 +36,13 @@
   </div>
 
   <div class="grid-row">
-      <section class="column-one-third search-page-filters">
+      <div class="column-one-third search-page-filters">
 
-      </section>
-      <section class="column-two-thirds" aria-label="Framework agreements">
+      </div>
+      <div class="column-two-thirds">
           {% include "_view_agreements_summary.html" %}
           {% include "_view_agreements_results.html" %}
-      </section>
+      </div>
   </div>
 
 {% endblock %}
--- a/app/templates/view_agreements_g8.html
+++ b/app/templates/view_agreements_g8.html
@@ -1,0 +1,48 @@
+{% import "toolkit/summary-table.html" as summary %}
+
+{% extends "_base_page.html" %}
+
+{% block breadcrumb %}
+  {%
+      with items = [
+          {
+              "link": url_for('.index'),
+              "label": "Admin home"
+          }
+      ]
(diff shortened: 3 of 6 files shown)


risicle:
Rebased to be sitting atop #207, waiting for that to go in.

TheDoubleK:
Needs another rebase (against master now), but looking good.

I think it needs some space added below the explanatory text, looks a bit snug at the moment.
screen shot 2016-08-26 at 13 52 48

app/templates/_view_agreements_results.html
```diff
@@ -0,0 +1,12 @@
+{% for supplier_framework in supplier_frameworks %}
+<div class="search-result">
+  <h2 class="search-result-title">
+    <a href="{{ url_for('.view_signed_agreement', supplier_id=supplier_framework.supplierId, framework_slug=supplier_framework.frameworkSlug) }}">{{ supplier_framework.supplierName }}</a>
+  </h2>
+  <ul class="search-result-important-metadata">
+    <li class="search-result-metadata-item">
+      Submitted: {{ supplier_framework.agreementReturnedAt }}
```
TheDoubleK: I can see it's working on the page alright, but how does the timestamp get converted into a nice human-friendly string without having ` |datetimeformat` after it?

risicle: `agreements.py:23` does it in the view (I didn't change it)

risicle:
And when it comes to the layout, the lefthand column is there, but doesn't display as it doesn't have any content yet.

app/templates/_view_agreements_results.html
TheDoubleK: Ah yep, I see it now. Most other places we tend to do the formatting in the template I think, not sure why this got done in the view.

TheDoubleK:
Yep I saw that empty column there.

There is a style `marketplace-paragraph` that I think adds padding (or maybe margin?) at the bottom of some text - see, e.g. the top paragraph here: https://www.digitalmarketplace.service.gov.uk/buyers/frameworks/digital-outcomes-and-specialists/requirements/digital-specialists

I'm not sure if it's in the toolkit or if it's a custom style in the buyer frontend though.  Either way, if you can pull it in here it should do the job.

risicle:
Ah good I was wondering what the "right" way of doing that was.

TheDoubleK:
I'm not the best person to say what's "right", but at least I know how it's been done before :)

risicle:
Hmm problem is that style's explicitly in e.g. the buyer app's `application.scss`, not present in the admin one.

TheDoubleK:
I'd just copy the style over to the bottom of `application.scss` in this app.  But others might have a better solution.

risicle:
Done and done.

ralph-hawkins:
I'd like to talk to someone about the layout of this before it goes live please.

app/templates/view_agreements_g8.html
```diff
@@ -0,0 +1,48 @@
...
+        </p>
+      </div>
+    </div>
+  </div>
+
+  <div class="grid-row">
+      <section class="column-one-third search-page-filters">
+
+      </section>
+      <section class="column-two-thirds" aria-label="Framework agreements">
```
tombye: I don't think we need the `aria-label` attribute here because the search results are the main content in the page. 

To explain, I added it to the G-Cloud search page because the search results followed the filters in the page order, meaning they took a while to get to if going through the page in a linear fashion. Giving the search results region an `aria-label` meant they had a human-readable name to identify them so users of assistive tech' could find and jump to them easily.

We don't have the same problems in this page so I think the attribute may confuse matters.

risicle:
bunch of @tombye things addressed now

app/templates/_view_agreements_summary.html
```diff
@@ -0,0 +1,8 @@
+{% if supplier_frameworks %}
+{% set content %}
+  <span class="search-summary-count">{{ supplier_frameworks|length }}</span>
+  {{ pluralize(supplier_frameworks|length, "agreement", "agreements") }}
+  <strong>waiting for countersigning</strong>
```
tombye: These `<strong>` and `<span>` tags should be `<em>`s to match those used in the buyer app:

https://github.com/alphagov/digitalmarketplace-buyer-frontend/blob/1de609c5cb69bd3e1cdc8138f1bd5f78895a86a5/app/presenters/search_summary.py#L10

app/templates/view_agreements_g8.html
tombye: Update: after talking to @risicle the first column will have some content in it but this will be quite minimal. That being the case, it's still a good idea to remove the `aria-label` as it still won't be that hard to get to the search results.

That being the case we should also swap the `section` tags out for `div`s. `section` tags define regions in the document (see https://www.w3.org/TR/html-aam-1.0/#el-section) which we have established we don't need. `div`s have no role so can be used for visuals alone (see https://www.w3.org/TR/html-aam-1.0/#el-div).

tombye:
👍  for the front-end stuff in this.

PR #207, opened 2016-08-26
[#127982711] Admin views signed agreement page

TheDoubleK:
For this story: https://www.pivotaltracker.com/story/show/127982711

First version of agreement signing page, buttons and lot details to be added by later stories.

~~For embedding PDFs I discussed with @pcraig3 and we agreed that hard-coding `height="905"` is a reasonable approach.  Reasoning as follows: the 100% width of the column is 640px, and the aspect ratio of A4 is 1.414.  640*1.414 is (almost) 905.~~
That logic all still stands, but the way PDF embedding looks it required a small amount of scroll bar so I've increased the height to 930px. (Testing it it seems like 925px would probably do, but it doesn't hurt to give it the extra 5px). [Screenshot now updated to no scroll bar version]

## Supplier has returned an image file (png in this case)
countersign g-cloud 8 agreement-img

## Supplier has returned a PDF
screen shot 2016-08-26 at 11 38 23


app/templates/suppliers/view_signed_agreement.html
```diff
@@ -0,0 +1,104 @@
...
+{% block breadcrumb %}
+    {%
+    with items = [
+        {
+            "link": url_for('.index'),
+            "label": "Admin home"
+        },
+        {
+            "link": url_for('.list_agreements', framework_slug=framework.slug),
+            "label": "{} agreements".format(framework.name)
```
allait: Trying to `.format` a string with a unicode string breaks in Python 2. I think we should start using `"%s agreements" | format(framework.name)` for things like this, even though we don't have any non-latin characters in our framework names right now.

allait:
Looks good 👍

app/templates/suppliers/view_signed_agreement.html
```diff
@@ -0,0 +1,104 @@
...
+            {%
+            with
+                text = supplier_framework.declaration.companyRegistrationNumber,
+                link = "https://beta.companieshouse.gov.uk/company/{}".format(supplier_framework.declaration.companyRegistrationNumber)
+            %}
+            {% include "toolkit/external-link.html" %}
+            {% endwith %}
+        </div>
+        
+        <h2>Appointment is to</h2>
```
allait: I think we can hide this block for now

PR #208, opened 2016-08-26
Break long emails

TheDoubleK:
This is an issue raised by @andrewchong when testing the page.  I'm not sure which looks worse to be honest, but at least we can see it here for discussion and put it in if we like the broken version more.

## Before
screen shot 2016-08-26 at 15 13 05

## After
screen shot 2016-08-26 at 15 14 57


risicle:
Yup I think it's (a little) better :+1:

PR #209, opened 2016-09-05
[#128842227]view_signed_agreement: show lots supplier has applied for

risicle:
This is waiting on a decision about what `status` `Service`s to take into account when aggregating for the lots (see `TODO` comment in code).

Don't imagine we should be counting `DraftService`s too should we?

### DIFF
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -1,3 +1,5 @@
+from collections import OrderedDict
+
 from flask import render_template, request, redirect, url_for, abort, current_app
 from flask_login import login_required, current_user, flash
 from dateutil.parser import parse as parse_date
@@ -87,6 +89,16 @@ def view_signed_agreement(supplier_id, framework_slug):
     if not supplier_framework.get('agreementReturned'):
         abort(404)
 
+    # build an OrderedDict of applied-for lotSlug against lotName, ordered by lotSlug
+    lot_slugs_names = OrderedDict(sorted(
+        (service["lotSlug"], service["lotName"],)
+        for service in data_api_client.find_services_iter(
+            supplier_id=supplier_id,
+            framework=framework_slug,
+            )
+        )
+    )
+
     agreements_bucket = s3.S3(current_app.config['DM_AGREEMENTS_BUCKET'])
     prefix = get_agreement_document_path(framework_slug, supplier_id, SIGNED_AGREEMENT_PREFIX)
     agreement_documents = agreements_bucket.list(prefix=prefix)
@@ -101,6 +113,7 @@ def view_signed_agreement(supplier_id, framework_slug):
         supplier=supplier,
         framework=framework,
         supplier_framework=supplier_framework,
+        lot_slugs_names=lot_slugs_names,
         agreement_url=url,
         agreement_ext=agreement_documents[-1]['ext']
     )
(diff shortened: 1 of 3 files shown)


app/main/views/suppliers.py
```diff
@@ -87,6 +89,17 @@ def view_signed_agreement(supplier_id, framework_slug):
     if not supplier_framework.get('agreementReturned'):
         abort(404)
 
+    # build an OrderedDict of applied-for lotSlug against lotName, ordered by lotSlug
+    lot_slugs_names = OrderedDict(sorted(
+        (service["lotSlug"], service["lotName"],)
+        for service in data_api_client.find_services_iter(
+            supplier_id=supplier_id,
+            framework=framework_slug,
+            # TODO filter by status?
```
TheDoubleK: This `TODO` should come out of the code if we're not going to do it.

app/main/views/suppliers.py
```diff
@@ -87,6 +89,17 @@ def view_signed_agreement(supplier_id, framework_slug):
     if not supplier_framework.get('agreementReturned'):
         abort(404)
 
+    # build an OrderedDict of applied-for lotSlug against lotName, ordered by lotSlug
+    lot_slugs_names = OrderedDict(sorted(
```
TheDoubleK: Yep this will give the right order for lots, namely:
Lot 1: Infrastructure as a Service (IaaS)
Lot 2: Platform as a Service (PaaS)
Lot 3: Software as a Service (SaaS)
Lot 4: Specialist Cloud Services (SCS)

Not sure if this is a lucky coincidence or if the lots were just numbered in alphabetical order by CCS in the first place.

tests/app/main/views/test_suppliers.py
```diff
@@ -1068,6 +1091,9 @@ def test_should_show_agreement_details_on_page(self, s3, data_api_client):
         assert len(document.xpath('//li[contains(text(), "DG9 0QG")]')) == 1
         # Company number - linked
         assert len(document.xpath('//a[@href="https://beta.companieshouse.gov.uk/company/1234456"][contains(text(), "1234456")]')) == 1  # noqa
+        # Lots
+        assert len(document.xpath('//li[contains(text(), "Lettuce & cucumber")]')) == 1
```
TheDoubleK: If there's a simple way to test the displayed order of lots it might be good to test that too, but not if it's a complete headache.

TheDoubleK:
Couple of very minor comments, but should be pretty much good to go after a rebase.

app/main/views/suppliers.py
risicle: Oh shag yes

tests/app/main/views/test_suppliers.py
risicle: Yeah I was thinking about this - my first instinct was to try and find a way of getting lxml to tell me what character an element was found at but no dice... I'll think...

TheDoubleK:
I don't think testing the order is worth sweating about :+1:

risicle:
@tombye is it ok that I added a class to the html here just for the purposes of being able to target it in one of the tests would you say?

risicle:
Ok @allait says it's not worth testing for order if we'd have to put a special class in for the purpose. Removing test.

PR #211, opened 2016-09-07
Add opportunity location to buyer users download

TheDoubleK:
It is useful to know the location of opportunities (e.g. for user research recruitment), so we can include them in the buyer users export here.

### DIFF
--- a/app/main/views/users.py
+++ b/app/main/views/users.py
@@ -104,7 +104,11 @@ def download_buyers_and_briefs():
 
     for brief in briefs:
         for user in brief['users']:
-            brief_string = '{} - {}'.format(brief['title'], 'open' if brief['status'] == 'live' else brief['status'])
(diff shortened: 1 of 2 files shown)


risicle:
Only thing I would say is if you're adding fields to character (hyphen, comma...) separated data outputs, adding it as the _final_ column usually makes it more backwards-compatible for any consuming code there may be (consider fetching a subfield through `data.split("-")[1]`)

TheDoubleK:
It's a reasonable point, but there is certainly no consuming code anywhere for this, and I think it makes more logical sense to have the location before the status.

Also consider `data.split("-")[-1]`, which would continue to work with it this way round - any change is potentially a breaking change for someone, but in this case we know nothing is reading this programmatically anyway.

risicle:
Fair 'nuf :+1:

PR #212, opened 2016-09-12
[#129915689]download_buyers_and_briefs: redesign to output rows of buyer-brief combinations instead of briefs aggregated to buyer rows…

risicle:
Story https://www.pivotaltracker.com/story/show/129915689

Tried to implement this in a nice declarative, functional way.

Ended up using a standard `Writer` in the end instead of a `DictWriter` as I decided the overhead of creating & merging all the intermediate dicts wasn't worth it.

### DIFF
--- a/app/main/views/users.py
+++ b/app/main/views/users.py
@@ -105,9 +105,13 @@ def download_buyers_and_briefs():
         "name",
         "emailAddress",
         "phoneNumber",
-        "createdAt",
     )
-    # no user_generated_fields as of yet
+    user_generated_fields = OrderedDict((
+        (
+            "createdAtDate",
+            lambda brief: brief.get("createdAt", "").partition("T")[0],  # cheap truncation of iso timestamp
+        ),
+    ))
     brief_verbatim_fields = (
         "title",
         "location",
@@ -118,8 +122,9 @@ def download_buyers_and_briefs():
             lambda brief: "open" if brief.get("status") == "live" else brief.get("status", ""),
         ),
         (
-            "applicationsClosedAtIfClosed",
-            lambda brief: brief.get("applicationsClosedAt", "") if brief.get("status") == "closed" else "",
+            "applicationsClosedAtDateIfClosed",
+            lambda brief:
+                (brief.get("applicationsClosedAt", "") if brief.get("status") == "closed" else "").partition("T")[0],
         ),
     ))
 
@@ -128,7 +133,7 @@ def download_buyers_and_briefs():
             # header row
             tuple(chain(
                 ("user.{}".format(field_name) for field_name in user_verbatim_fields),
-                # no user_generated_fields as of yet
+                ("user.{}".format(field_name) for field_name in iterkeys(user_generated_fields)),
                 ("brief.{}".format(field_name) for field_name in brief_verbatim_fields),
                 ("brief.{}".format(field_name) for field_name in iterkeys(brief_generated_fields)),
             )),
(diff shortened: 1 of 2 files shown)


app/main/views/users.py
```diff
@@ -87,42 +93,74 @@ def download_users(framework_slug):
...
-    briefs = (brief for brief in data_api_client.find_briefs_iter(with_users=True))
-    buyer_headings = [
+    users = {user["id"]: dict(user, briefs=[]) for user in data_api_client.find_users_iter(role="buyer")}
+
+    # join users with briefs (a "hash join")
+    for brief in data_api_client.find_briefs_iter(with_users=True):
+        for user in brief["users"]:
+            users[user["id"]]["briefs"].append(brief)
+
+    timestamp = datetime.utcnow().strftime('%Y%m%dT%H%M%S')
```
pcraig3: Small point, but could you not use the `DATETIME_FORMAT` we have in [utils](https://github.com/alphagov/digitalmarketplace-utils/blob/master/dmutils/formats.py)?

app/main/views/users.py
```diff
@@ -87,42 +93,74 @@ def download_users(framework_slug):
...
-                brief['title'],
-                brief.get('location'),
-                'open' if brief['status'] == 'live' else brief['status'],
+    )
+    user_generated_fields = OrderedDict((
+        (
+            "createdAtDate",
+            lambda brief: brief.get("createdAt", "").partition("T")[0],  # cheap truncation of iso timestamp
+        ),
+    ))
```
pcraig3: I think this is really cool.

app/main/views/users.py
```diff
@@ -87,42 +93,74 @@ def download_users(framework_slug):
...
+                (brief.get(field_name, "") for field_name in brief_verbatim_fields),
+                (func(brief) for func in itervalues(brief_generated_fields)),
+            ))
+            for user, brief in chain.from_iterable(
+                (
+                    (user, brief) for brief in
+                    (user["briefs"] or ({},))   # if user has an empty seq of briefs add a single fake blank
+                                                # one ("outer join" behaviour)
+                )
+                for user in sorted(itervalues(users), key=lambda user: user["name"])
```
pcraig3: These nested `for`s that generate everything is a bit overwhelming, but once you figure out to work from the bottom up, it seems like it's probably okay.

app/main/views/users.py
risicle: This is inherited from previous code - I looked at whether we could do it less arbitrarily and I think I concluded we couldn't. It's not using `DATETIME_FORMAT` because it needs to be used for a filename and therefore restricts its "funny characters". Should possibly add a comment about this.

app/main/views/users.py
risicle: I agree. This is the "normal" way of doing things with functional programming. In a way, the loops are inside out, and I think I could have made it clearer through formatting had pep8 wanted to play ball. _I_ think it's much clearer about where all your data dependencies are and the declarative approach makes it significantly harder to screw up.

risicle: One thing it *does* make harder is commenting - not everywhere you want to leave a comment is a convenient/syntactically valid place to leave a comment, because it's all technically inside one expression.

app/main/views/users.py
risicle: I think _you're_ really cool.

Hm.

app/main/views/users.py
pcraig3:
```python
(
  {
    [
      # yeah, no kidding
    ]
  }
)
```

app/main/views/users.py
pcraig3: Oh right, so you don't want to use hyphens or dots or anything.

Fair enough.

TheDoubleK:
@pcraig3 I think it still needs a separate thumbs-up comment to go green on the 4th wall.

Also, not sure why we aren't seeing the new "rebase and merge" button...  seems like it would be a good thing: https://github.com/blog/2243-rebase-and-merge-pull-requests

TheDoubleK:
Checked repository settings and looks like it should be there:
screen shot 2016-09-26 at 17 52 07

PR #214, opened 2016-10-05
Add buttons to set framework agreements as 'on hold' or 'countersigned'

Wynndow:
For [this story on Pivotal](https://www.pivotaltracker.com/story/show/128842457).
Also for [this bug on Pivotal](https://www.pivotaltracker.com/story/show/130099083).

This PR adds couple of POST routes to the admin frontend which hit the API and timestamp the relevant `framework_agreement` to show when they were countersigned or put on hold.

If countersigned some details are also stored about who legally countersigned as well as who actually hit the button.

STILL TO DO:
Currently, the post routes return the user to the framework agreement they were working on. This is not the expected behaviour. It should take the user to the next framework agreement that needs countersigning/putting on hold. The logic for getting this next agreement is a separate piece of work outside the scope of this story.

Will need [this PR on the toolkit to be merged before the tests will go green](https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/274)

### DIFF
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -79,7 +79,7 @@ def view_supplier_declaration(supplier_id, framework_slug):
 
 @main.route('/suppliers/<supplier_id>/agreements/<framework_slug>', methods=['GET'])
 @login_required
-@role_required('admin-ccs-sourcing')
+@role_required('admin', 'admin-ccs-sourcing')
 def view_signed_agreement(supplier_id, framework_slug):
     supplier = data_api_client.get_supplier(supplier_id)['suppliers']
     framework = data_api_client.get_framework(framework_slug)['frameworks']
@@ -116,6 +116,38 @@ def view_signed_agreement(supplier_id, framework_slug):
         lot_slugs_names=lot_slugs_names,
         agreement_url=url,
         agreement_ext=agreement_documents[-1]['ext']
+    )
+
+
+@main.route('/suppliers/agreements/<agreement_id>/on-hold', methods=['POST'])
+@login_required
+@role_required('admin-ccs-sourcing')
+def put_signed_agreement_on_hold(agreement_id):
+    data_api_client.put_signed_agreement_on_hold(agreement_id, current_user.email_address)
+
+    organisation = request.args['organisation']
+    flash('The agreement for {} was put on hold.'.format(organisation), 'message')
+
+    return redirect(url_for(
+        '.list_agreements',
+        framework_slug='g-cloud-8')
+    )
+
+
+@main.route('/suppliers/agreements/<agreement_id>/approve', methods=['POST'])
+@login_required
+@role_required('admin-ccs-sourcing')
+def approve_agreement_for_countersignature(agreement_id):
+    data_api_client.approve_agreement_for_countersignature(agreement_id,
+                                                           current_user.email_address,
+                                                           current_user.id)
+
+    organisation = request.args['organisation']
+    flash('A confirmation was sent to {}.'.format(organisation), 'message')
+
+    return redirect(url_for(
+        '.list_agreements',
+        framework_slug='g-cloud-8')
     )
 
 
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -125,7 +125,7 @@ def view_signed_agreement(supplier_id, framework_slug):
 def put_signed_agreement_on_hold(agreement_id):
     data_api_client.put_signed_agreement_on_hold(agreement_id, current_user.email_address)
 
-    organisation = request.args['organisation']
+    organisation = request.form['nameOfOrganisation']
     flash('The agreement for {} was put on hold.'.format(organisation), 'message')
 
     return redirect(url_for(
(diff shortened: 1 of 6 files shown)


app/main/views/suppliers.py
```diff
@@ -119,6 +119,40 @@ def view_signed_agreement(supplier_id, framework_slug):
...
+@main.route('/suppliers/<supplier_id>/agreements/<framework_slug>/on-hold', methods=['POST'])
+@login_required
+@role_required('admin-ccs-sourcing')
+def put_signed_agreement_on_hold(supplier_id, framework_slug):
+    supplier_framework = data_api_client.get_supplier_framework_info(supplier_id, framework_slug)['frameworkInterest']
+    data_api_client.put_signed_agreement_on_hold(supplier_framework['agreementId'], current_user.email_address)
+
+    flash(supplier_framework['declaration']['nameOfOrganisation'], 'put on hold')
+
+    return redirect(url_for(
```
Wynndow: Yeah, I'd been working with just returning the agreement that was just put onhold/approved with the view that finding the next one to be processed was another story. I'm gonna leave it just redirecting to the `admin/agreements/g-cloud-8` page

app/main/views/suppliers.py
```diff
@@ -119,6 +119,40 @@ def view_signed_agreement(supplier_id, framework_slug):
     )
 
 
+@main.route('/suppliers/<supplier_id>/agreements/<framework_slug>/on-hold', methods=['POST'])
```
Wynndow: Cool. I was missing the fact you could pull out `supplier_framework.agreementId` in the template. Makes things a lot tidier too.

Wynndow:
@TheDoubleK All done!

app/main/views/suppliers.py
```diff
@@ -119,6 +119,38 @@ def view_signed_agreement(supplier_id, framework_slug):
...
+@main.route('/suppliers/agreements/<agreement_id>/approve', methods=['POST'])
+@login_required
+@role_required('admin-ccs-sourcing')
+def approve_agreement_for_countersignature(agreement_id):
+    data_api_client.approve_agreement_for_countersignature(agreement_id,
+                                                           current_user.email_address,
+                                                           current_user.id)
+
+    organisation = request.args['organisation']
+    flash('A confirmation was sent to {}.'.format(organisation), 'message')
```
TheDoubleK: I don't think this message is right - no-one is sent anything at this point.  How about `'The agreement for {} was approved for countersigning.`?  Or maybe @laurenceberry has better words?

app/main/views/suppliers.py
```diff
@@ -119,6 +119,38 @@ def view_signed_agreement(supplier_id, framework_slug):
     )
 
 
+@main.route('/suppliers/agreements/<agreement_id>/on-hold', methods=['POST'])
+@login_required
+@role_required('admin-ccs-sourcing')
+def put_signed_agreement_on_hold(agreement_id):
+    data_api_client.put_signed_agreement_on_hold(agreement_id, current_user.email_address)
+
+    organisation = request.args['organisation']
```
TheDoubleK: Passing in the organisation as a parameter on the form target seems odd to me.  Two options that I think would be better would be:
1. Pass it in a hidden form field from the page (`input type="hidden"`)
2. Do two API fetches in the POST route here - first get the FrameworkAgreement and then the SupplierFramework

I think option 1 would be OK as the value is only used for the flash messaging, so we can probably do without the API fetches.

app/main/views/suppliers.py
```diff
@@ -119,6 +119,38 @@ def view_signed_agreement(supplier_id, framework_slug):
...
+    flash('The agreement for {} was put on hold.'.format(organisation), 'message')
```
TheDoubleK: The flashed messages don't appear on the list page, but should be fine to leave it for when the "next agreement" flow gets written.

app/main/views/suppliers.py
laurenceberry: Ran this past Andy too, can we go with "The agreement for {} was approved. They will be sent the countersigned version soon."

This makes it clearer for admins who might need to know what happens next for the Supplier.

TheDoubleK: "They will be sent..." doesn't sound like something we're actually planning to do though?  We will be making them available to download from their account page, not actively sending them.

laurenceberry: "The agreement for {} was approved. They will receive a countersigned version soon" is more vague, and should provide enough info I think.

app/main/views/suppliers.py
Wynndow: I thought passing it in would be good as it'd save an api call, but the hidden form field is much better!

app/main/views/suppliers.py
Wynndow: Yeah, I've put them in in anticipation of the route redirecting to another agreement page instead of the list view. It's only pointing at the list view atm cos the filtering isn't in place to get the next agreement.

Wynndow:
@TheDoubleK All those changes should be in!

TheDoubleK:
Good stuff.  Let's get this in once Travis is happy (needs the frontend toolkit to get merged first I think).
:+1:

PR #217, opened 2016-10-11
Create companies house link with correct target

Wynndow:
For [this story on Pivotal](https://www.pivotaltracker.com/story/show/128842457).

The link to companies house was being generated by the frontend toolkit
with the value for target set to `_blank` (missing quotes). This wasn't
valid but failed silently. This small update adds the quotes so it's
correctly applied by the toolkit.

### DIFF
--- a/app/templates/suppliers/view_signed_agreement.html
+++ b/app/templates/suppliers/view_signed_agreement.html
@@ -63,7 +63,7 @@ Countersign {{ framework.name }} agreement for {{ supplier_framework.declaration
           with
               text = supplier_framework.declaration.companyRegistrationNumber,
               link = "https://beta.companieshouse.gov.uk/company/%s" | format(supplier_framework.declaration.companyRegistrationNumber),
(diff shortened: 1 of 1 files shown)


TheDoubleK:
Would it be better to have the quotes added by the toolkit?

Wynndow:
Yep. That's what other things seems to be doing. Will update.

TheDoubleK:
Sorry!

Wynndow:
Having said that, the toolkit already was adding the quotes: `{% if target %}target="{{ target }}"{% endif %}` and leaving the quotes off in the admin-frontend caused it to not work.

Also, the existing link in the admin app has quotes around it, and they're also being added in the frontend toolkit.

I've got a feeling that having the quotes in the admin app is allowing jinja to recognise it as a string and process it correctly. Leaving the quotes off means that nothing at all gets passed through. What do you think?

TheDoubleK:
:+1:

PR #218, opened 2016-10-12
[#128843445]agreement countersigning: add a next_agreement view...

risicle:
…which will redirect to the "next" SupplierFramework with a signed agreement

(pivotal story https://www.pivotaltracker.com/story/show/128843445)

Not yet added anything that actually _uses_ this view yet, because I'm not entirely sure about the volatility of the areas that will reference this view.

### DIFF
--- a/tests/app/main/views/test_suppliers.py
+++ b/tests/app/main/views/test_suppliers.py
@@ -1189,44 +1189,68 @@ class TestViewingASupplierDeclaration(LoggedInApplicationTest):
 class TestPutSignedAgreementOnHold(LoggedInApplicationTest):
     user_role = 'admin-ccs-sourcing'
 
+    @property
+    def put_signed_agreement_on_hold_return_value(self):
+        # a property so we always get a clean *copy* of this to work with
+        return {
+            "agreement": {
+                "id": 123,
+                "supplierId": 4321,
+                "frameworkSlug": "g-cloud-99-flake",
+            },
+        }
+
     def test_it_fails_if_not_ccs_admin(self, data_api_client):
         self.user_role = 'admin'
-        res = self.client.post('/admin/suppliers/agreements/123/on-hold')
+        data_api_client.put_signed_agreement_on_hold.return_value = self.put_signed_agreement_on_hold_return_value
+        res = self.client.post('/admin/suppliers/agreements/123/on-hold', data={"nameOfOrganisation": "Test"})
 
+        assert not data_api_client.put_signed_agreement_on_hold.called
         assert res.status_code == 403
 
-    def test_it_correctly_calls_the_apiclient(self, data_api_client):
-        res = self.client.post('/admin/suppliers/agreements/123/on-hold')
+    def test_happy_path(self, data_api_client):
+        data_api_client.put_signed_agreement_on_hold.return_value = self.put_signed_agreement_on_hold_return_value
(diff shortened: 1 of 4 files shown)


risicle:
Ok the only remaining outstanding thing is whether we put a plain-ol' "Next" link/button on the countersigning page if the agreement is already on hold.

risicle:
Added "Next agreement" links to agreement page.
201601014_nextagreement

tests/app/main/views/test_suppliers.py
```diff
@@ -1268,53 +1300,130 @@ def test_none_shown_if_user_not_ccs_admin(self, s3, get_signed_url, data_api_cli
         res = self.client.get('/admin/suppliers/1234/agreements/g-cloud-8')
         assert res.status_code == 200
 
-        data = res.get_data(as_text=True)
+        document = html.fromstring(res.get_data(as_text=True))
 
-        assert 'Accept and continue' not in data
-        assert 'Put on hold and continue' not in data
+        assert len(document.xpath("//input[@type='submit'][@value='Accept and continue']")) == 0
```
TheDoubleK: I find these very specific tests for where something *isn't* to be less convincing than the previous "text is nowhere in the page" tests.  I agree that for checking that something *is* rendered on the page then these xpath specific checks are better, but for asserting things *not* being there I think it's better to cast a wider net. (Because designs and HTML structure can change and leave the specific test passing even though the unwanted text now appears on the page in a slightly different element).

tests/app/main/views/test_suppliers.py
```diff
@@ -1268,53 +1300,130 @@ def test_none_shown_if_user_not_ccs_admin(self, s3, get_signed_url, data_api_cli
...
         res = self.client.get('/admin/suppliers/1234/agreements/g-cloud-8')
         assert res.status_code == 200
 
-        data = res.get_data(as_text=True)
-
-        assert 'Accept and continue' in data
-        assert 'Put on hold and continue' in data
+        document = html.fromstring(res.get_data(as_text=True))
+
+        assert len(document.xpath(
```
TheDoubleK: Yep, these ones are much better.

tests/app/main/views/test_suppliers.py
```diff
@@ -1268,53 +1300,130 @@ def test_none_shown_if_user_not_ccs_admin(self, s3, get_signed_url, data_api_cli
...
-        data = res.get_data(as_text=True)
+        document = html.fromstring(res.get_data(as_text=True))
 
-        assert 'Accept and continue' in data
-        assert 'Put on hold and continue' not in data
+        assert len(document.xpath(
+            "//form[@action='/admin/suppliers/agreements/4321/approve']"
+            "//input[@type='submit'][@value='Accept and continue']"
+        )) == 1
+        assert len(document.xpath("//input[@type='submit'][@value='Put on hold and continue']")) == 0
```
TheDoubleK: But these ones catch a smaller number of failure cases.

tests/app/main/views/test_suppliers.py
risicle: Yes interesting - the main reason I replaced this is because doing a raw substring search is the latter doesn't perform any normalization of the text, and so it's a bad habit to get into.

tests/app/main/views/test_suppliers.py
risicle: Oh I see, because they're negative. Yeah. Hmm. I'm going to think about this. This one would fail to pick up if someone switched this to be a `<button>` and it started appearing.

PR #219, opened 2016-10-12
Change agreementStatus from 'on hold' to 'on-hold'

idavidmcdonald:
This is about to change in the API and therefore needs to be updated in the front end.


pcraig3:
👍

PR #220, opened 2016-10-13
Do not reverse order of agreements

TheDoubleK:
The API will soon ([see this pull request](https://github.com/alphagov/digitalmarketplace-api/pull/474)) return agreements in ascending order of date as that is the order we want.  So then there will be no need to reverse the agreements here.


idavidmcdonald:
Looks good. Probably merge it when the API has been merged? 👍

PR #222, opened 2016-10-18
[#128841821]list_agreements: add ability to filter list of suppliers by latest agreement status

risicle:
Story https://www.pivotaltracker.com/story/show/128841821

I'm intending to push the functionality enabling the "next" link between agreements to just traverse over the currently filtered set of agreements. However, to do that, this would need to sit on top of #218 which is not merged yet. So here's the straightforward filtering functionality.

### DIFF
--- a/app/main/views/agreements.py
+++ b/app/main/views/agreements.py
@@ -13,11 +13,11 @@ from ..auth import role_required
 from ... import data_api_client
 
 
-_status_labels = OrderedDict((
+status_labels = OrderedDict((
     ("signed", "Waiting for countersigning"),
     ("on-hold", "On hold"),
-    ("approved,countersigned", "Countersigned"),  # ugly, but i don't want to start inventing new status values - much
-                                                  # easier to just act as a filter
+    ("approved,countersigned", "Countersigned"),  # ugly key, but i don't want to start inventing new status values -
+                                                  # much easier to just act as a filter
 ))
 
 
@@ -39,7 +39,7 @@ def list_agreements(framework_slug):
     framework = data_api_client.get_framework(framework_slug)['frameworks']
 
     status = request.args.get("status")
-    if status and status not in _status_labels:
+    if status and status not in status_labels:
         abort(400)
 
     supplier_frameworks = _get_ordered_supplier_frameworks(framework_slug, status=status)
@@ -54,7 +54,7 @@ def list_agreements(framework_slug):
         supplier_frameworks=supplier_frameworks,
         degenerate_document_path_and_return_doc_name=lambda x: degenerate_document_path_and_return_doc_name(x),
         status=status,
-        status_labels=_status_labels,
+        status_labels=status_labels,
     )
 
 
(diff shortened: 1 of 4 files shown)


app/main/views/agreements.py
```diff
@@ -10,13 +12,28 @@
...
     framework = data_api_client.get_framework(framework_slug)['frameworks']
+
+    status = request.args.get("status")
+    if status and status not in status_labels:
+        abort(400)
+
     supplier_frameworks = data_api_client.find_framework_suppliers(
-        framework_slug, agreement_returned=True
+        framework_slug,
+        agreement_returned=True,
```
TheDoubleK: I'll fix the API to do the right thing, but it still wouldn't hurt to pass only one or the other in to the API client here.

tests/app/main/views/test_agreements.py
```diff
@@ -52,35 +47,98 @@ def test_happy_path_g8(self, data_api_client):
...
+            ({"status": [status_key]}, status_label)
+            for status_key, status_label in iteritems(status_labels)
+        )
+
+    def test_happy_path_notall_g8(self, data_api_client):
+        data_api_client.get_framework.return_value = self.load_example_listing('framework_response')
+        data_api_client.find_framework_suppliers.return_value = self.find_framework_suppliers_return_value_g8
+
+        # choose the second status (if there is one, otherwise first)
+        chosen_status_key = tuple(iterkeys(status_labels))[:2][-1]
```
TheDoubleK: I don't understand what this is doing, or how there may or may not be a "second status"? `status_labels` is a fixed ordered dict of three things, so there will always be a second status and it will always be the same thing? What have I missed? Is there some randomisation in the choosing that I'm missing?

tests/app/main/views/test_agreements.py
```diff
@@ -52,35 +47,98 @@ def test_happy_path_g8(self, data_api_client):
...
+        response = self.client.get('/admin/agreements/g-cloud-8?status={}'.format(chosen_status_key))
+        page = html.fromstring(response.get_data(as_text=True))
+
+        assert response.status_code == 200
+
+        call_args = data_api_client.find_framework_suppliers.call_args
+        assert call_args[0] == ("g-cloud-8",)
+        # slightly elaborate assertion here to allow for missing kwargs defaulting to None
+        assert all(call_args[1].get(key) == value for key, value in (
+            ("agreement_returned", True),
```
TheDoubleK: This should change when the API client isn't called with both `agreement_returned` and `statuses` keys.

tests/app/main/views/test_agreements.py
risicle: Yeah I'm just trying to make it so that someone can come along and drop another status option in there or remove one and not have to come along and fix the tests.

tests/app/main/views/test_agreements.py
TheDoubleK: Yes it is :)

TheDoubleK:
Let's get this in, apologies for the delay :+1:

risicle:
Bam!

Got some fun stuff queued up on top of this waiting...

PR #223, opened 2016-10-24
[#128841821] Filter "next agreement" behaviour by status

risicle:
A bolt-on to story https://www.pivotaltracker.com/story/show/128841821

This PR causes the "next agreement" behaviour of the views surrounding countersigning to obey any current status filter applied to the list view that the agreement was accessed from.

(This is exactly the sort of behaviour the functional tests would be ideal to test, but I don't know if that would be worth my time...)

This necessitated a redesign of the `next_agreement` view - as previously designed, retrieving an agreement (A)'s "next" agreement when filtered by a status other than A's current status would fail because the view would fail to find A in the status-filtered API listing it requests (which it needs to do to be able to determine which one is "next"). So instead we now fetch an unfiltered listing (annotated with statuses) and perform any "filtering" ourselves.


app/main/views/agreements.py
```diff
@@ -62,28 +62,43 @@ def list_agreements(framework_slug):
 @login_required
 @role_required('admin', 'admin-ccs-sourcing')
 def next_agreement(supplier_id, framework_slug):
+    status = request.args.get("status")
```
risicle: This is a good question. I have no idea as to the answer. I think alexey's reaction that there was a distinction between `approved` and `countersigned` in the backend at all was surprise, so perhaps it's an unnecessary distinction...?

TheDoubleK: It's a necessary distinction, but perhaps not for the admin app. If countersigned agreements were instantly created and uploaded to S3 when Admin hits the button then there would be no need, but as it is the countersigned document is not returned to the supplier until some time after Admin hits the button, and so the supplier frontend needs some way to distinguish if it should show a link to download a countersigned document or not.

TheDoubleK:
:+1: let's do this as-is and add the individual `approved` and `countersigned` options if a user need arises.

PR #225, opened 2016-10-28
Change redirect location following supplier name update

TheDoubleK:
Also adds a couple of basic tests for previously untested route.


allait:
👍

PR #227, opened 2016-11-04
CCS admin countersigning - FINAL pieces

TheDoubleK:
Fixes this bug: https://www.pivotaltracker.com/story/show/133327957
And completes this story: https://www.pivotaltracker.com/story/show/127983609

## Homepage
 - Adds a link to the list of G-Cloud 8 agreements to be countersigned
screen shot 2016-11-04 at 13 43 31

## Suppliers list (for CCS Sourcing users view only)
 - Adds a "View agreement" link to the G-Cloud 8 column
 - Reverts to old behaviour of showing "signed agreement" and "countersigned agreement" links for all suppliers and all frameworks regardless of whether or not the supplier is on the framework (this is part of the bug-fix #133327957 where the page load timed out because of multiple API calls).
screen shot 2016-11-04 at 13 46 20
 - Also as part of that bug-fix this adds a new "jump page" route for downloading signed agreement files - this is required as the path to the file in S3 is no longer predictable from just the supplier and framework IDs as it was previously - the path now instead needs to be read from the SupplierFramework record.  The new route - linked to from the suppliers list page - gets the correct path from the SupplierFramework and passes it through to the existing `download_agreement_file()` route, which is a dumb proxy to S3.

## Agreements list (for Admin and CCS Sourcing users)
 - Show the search summary even when there are no agreements in the list
screen shot 2016-11-04 at 13 50 51

 - Show an "Easter egg" animated cat gif when all returned agreements have been countersigned
screen shot 2016-11-04 at 13 50 34


app/main/views/suppliers.py
```diff
@@ -183,6 +163,16 @@ def approve_agreement_for_countersignature(agreement_id):
     ))
 
 
+@main.route('/suppliers/<supplier_id>/agreement/<framework_slug>', methods=['GET'])
+@login_required
+@role_required('admin', 'admin-ccs-sourcing')
+def download_signed_agreement_file(supplier_id, framework_slug):
+    # This route is used for pre-G-Cloud-8 agreement document downloads
```
idavidmcdonald: Could be good for this comment to indicate the relationship of this route with the `download_agreement_file` route, i.e. why does this route exist and what is the difference between the two in terms of usage

app/main/views/agreements.py
```diff
@@ -28,7 +28,9 @@ def _get_ordered_supplier_frameworks(framework_slug, status=None):
         **({"statuses": status} if status else {})
     )['supplierFrameworks']
 
-    # API now returns SupplierFrameworks by agreementReturnedAt ascending (oldest first)
+    # Timestamps in our API responses are formatted using dmutils DATETIME_FORMAT which sorts in lexicographical order
+    # i.e. "%Y-%m-%dT%H:%M:%S.%fZ" - we want the newest things at the bottom of the list so this ordering is correct
+    supplier_frameworks = sorted(supplier_frameworks, key=lambda k: k.get("agreementReturnedAt", "2000-01-01"))
```
idavidmcdonald: Maybe a short comment to say that "2000-01-01" is an arbitrary date to enable ordering for supplier frameworks that are not returned. If we hadn't already discussed this before I don't think I would have got this particularly quickly.

tests/app/main/views/test_suppliers.py
```diff
@@ -810,6 +810,21 @@ def test_should_set_declaration(self, data_api_client):
...
+@mock.patch('app.main.views.suppliers.download_agreement_file')
+class TestDownloadSignedAgreementFile(LoggedInApplicationTest):
+    user_role = 'admin-ccs-sourcing'
+
+    def test_download_agreement_is_called_with_the_right_parameters(self, download_agreement_file, data_api_client):
+        data_api_client.get_supplier_framework_info.return_value = {
+            'frameworkInterest': {'agreementPath': '/path/to/file/in/s3/1234-signed-agreement-file.pdf'}
+        }
+        # Mock out a response from download_agreement_file() - we don't care what it is
+        download_agreement_file.side_effect = HTTPError(Response(404))
```
idavidmcdonald: Maybe I haven't quite understood what's going on but if we don't care what the response is from `download_agreement_file`, do we even need to set a side effect? Putting an HTTPError here makes me think that the error is important/relevant to the test and we are testing error conditions when we aren't...

idavidmcdonald:
👍 Looks good. I've left a few minor comments about improving the readability, feel free to implement before merging if you agree.

tests/app/main/views/test_suppliers.py
TheDoubleK: The test doesn't work without a response from `download_agreement_file` and I couldn't figure out how to mock a 200 so settled for a 404 with a comment.

TheDoubleK:
Damn, I totally forgot to push my updates up before I hit the merge button.

TheDoubleK:
I think it should be OK as is.

PR #229, opened 2016-11-10
Fix npm bootstrap

galund:
Notes and fixes for the initial set-up process (specifically under OSX).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,9 +11,11 @@ virtualenv:
 
 requirements: virtualenv requirements.txt
 	${VIRTUALENV_ROOT}/bin/pip install -r requirements.txt
+	npm install
 
 requirements_for_test: virtualenv requirements_for_test.txt
 	${VIRTUALENV_ROOT}/bin/pip install -r requirements_for_test.txt
+	npm install
 
 frontend_build:
 	npm run --silent frontend-build:production
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 SHELL := /bin/bash
 VIRTUALENV_ROOT := $(shell [ -z $$VIRTUAL_ENV ] && echo $$(pwd)/venv || echo $$VIRTUAL_ENV)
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -11,9 +11,11 @@ virtualenv:
 
 requirements: virtualenv requirements.txt
 	${VIRTUALENV_ROOT}/bin/pip install -r requirements.txt
+	npm install
```
allait: This should probably be a separate make step that `frontend_build` could depend on

README.md
```diff
@@ -10,6 +10,10 @@ Frontend administration application for the digital marketplace.
 
 ## Quickstart
 
+System requirements include a working node.js and npm.
+
+- note that on macOS, a `brew install` of node/npm results in `Error: Cannot find module 'internal/fs'` - use nodejs.org instead
```
allait: This is only true for the current latest version of node and our npm packages, I don't think it's worth adding this without specifying that

README.md
```diff
@@ -10,6 +10,10 @@ Frontend administration application for the digital marketplace.
 
 ## Quickstart
 
+System requirements include a working node.js and npm.
+
+- note that on macOS, a `brew install` of the latest node/npm results in `Error: Cannot find module 'internal/fs'` - use nodejs.org instead
```
galund: I'm not sure that entirely makes sense either: once you're using nvm, version 7 is fine too. I think mentioning specific versions here can only get out-of-date, but it's useful to mention the error and some possible fixes, in case it affects other people in future.

I'll try and re-word once more.

galund: (Oh, I'm wrong, >=7.0 doesn't seem to work with npm either. I think? I'm confused anyway, and therefore clearly not the right person to write this comment. I still don't think it makes sense to write anything too specific in here, as it will probably change / get fixed at some point.)

galund:
Shall I close this PR - is less comprehensive than #231. cc @danielknell

PR #230, opened 2016-11-11
XSS bug 132916889

galund:
These XSS bugs are fixed by applying patches / library upgrades that were already applied to the buyer and supplier apps.


allait:
I'm assuming this was fixed by the removal of `| safe` filters from the toolkit? Which toolkit templates were causing the issue? Would it make sense to write a test for this?

galund:
The relevant commit in the frontend-toolkit is abcd6b616ac40a570f7ef0ac9b99b1735358e698 (the change to the notification banner specifically I believe, @pcraig3 can correct me if I got that wrong).

Which component would you add a test to here?

allait:
@galund I think we should write regression tests for the scenarios we've had highlighted by ITHC - visit a URL/submit a form and check that the page has the escaped version of the input

Otherwise, there's very little evidence of the problem existing in the admin frontend at all, which means it can slip in again at one point.

galund:
@allait I think these two tests cover all three of the reported issues (i.e. cos 1 & 3 both redirect to the same place), is this what you had in mind?

galund:
@allait I'll probably need some help with understanding the approach you want us to take re python 2 and 3 compatibility :)

tests/app/main/views/test_xss.py
```diff
@@ -0,0 +1,35 @@
...
+
+import mock
+from ...helpers import LoggedInApplicationTest
+from dmapiclient import HTTPError
+import urllib
+
+
+class TestXSS(LoggedInApplicationTest):
+    @mock.patch('app.main.views.services.data_api_client')
+    def test_service_not_found_flash_message_injection(self, data_api_client):
```
allait: I think these tests should go to the corresponding view test files/classes. That's what we've done before in similar cases too, so it should be easier to figure out if we've accidentally duplicated a test this way.

tests/app/main/views/test_xss.py
```diff
@@ -0,0 +1,35 @@
...
+        """
+        # impl copied from test_services.TestServiceView#test_redirect_with_flash_for_api_client_404
+        api_response = mock.Mock()
+        api_response.status_code = 404
+        data_api_client.get_service.side_effect = HTTPError(api_response)
+
+        evil_service_id = "1<img src=a onerror=alert(1)>"
+        response1 = self.client.get('/admin/services/' + urllib.quote(evil_service_id))
+        response2 = self.client.get(response1.location)
+        self.assertNotIn(evil_service_id, response2.get_data(as_text=True))
```
allait: In addition to this we should be checking that the escaped version IS in the response text. Otherwise this test succeeds even if the request fails with a 404, 500 etc.

tests/app/main/views/test_xss.py
```diff
@@ -0,0 +1,35 @@
...
+        """
+        Asserts that raw HTML in a bad service ID cannot be injected into a flash message.
+        """
+        # impl copied from test_services.TestServiceView#test_redirect_with_flash_for_api_client_404
+        api_response = mock.Mock()
+        api_response.status_code = 404
+        data_api_client.get_service.side_effect = HTTPError(api_response)
+
+        evil_service_id = "1<img src=a onerror=alert(1)>"
+        response1 = self.client.get('/admin/services/' + urllib.quote(evil_service_id))
```
allait: I'd put the literal escaped text here instead of using `urllib.quote`. This solves the Py3 failures, and I'd argue that it makes the test clearer in a way, since you can see exactly what's being sent.

tests/app/main/views/test_xss.py
```diff
@@ -0,0 +1,35 @@
...
+    @mock.patch('app.main.views.services.data_api_client')
+    def test_brief_not_found_flash_message_injection(self, data_api_client):
+        """
+        Asserts that raw HTML in a bad brief ID cannot be injected into a flash message.
+        """
+        # impl copied from test_buyers.TestBuyersView#test_should_display_a_useful_message_if_no_brief_found
+        data_api_client.get_brief.return_value = None
+        evil_brief_id = "1<img src=a onerror=alert(1)>"
+        response = self.client.get('admin/buyers?brief_id=' + urllib.quote(evil_brief_id))
+        self.assertNotIn(evil_brief_id, response.get_data(as_text=True))
```
allait: Same point as above about checking that the escaped text is present

tests/app/main/views/test_xss.py
galund: I think if we want to check for status codes, that should be a new test - all I care about here is that there's no XSS? In the case of a 500, the test is likely to pass, but that really isn't a problem I think. In any case, `test_redirect_with_flash_for_api_client_404` will fail in that case, so adding a duplicate failing test that mentions XSS would just make what's broken less clear.

tests/app/main/views/test_xss.py
allait:
```python
try:
    from urlparse import urlparse, parse_qs
except ImportError:
    from urllib.parse import urlparse, parse_qs
```

is what we do almost everywhere, so I'd stick with that if you insist on using `urllib`. But I really don't think repetition is a concern here. We tend to be explicit in test assertions, repeating the literal value rather then referencing a variable since it removes a whole class of potential mistakes where the data could've been changed in the process.

tests/app/main/views/test_xss.py
galund: Meh, was going to put in a status code test here too, but turns out that we're *expecting* a 404 anyway. Not to worry...

allait:
Looks good to go 👍

PR #231, opened 2016-11-14
Fix initial project setup

danielknell:
- added build_frontend to run_app.sh inline with other projects
- added a bootstrap.sh for consistency with other projects
- added npm_install to run_all directive in Makefile
- added npm_install to the .PHONY directive in Makefile
- added package.json as dependency of npm_install in Makefile (inline with requirements / requirements_for_test)
- made run_app use the virtualenv python
- used DM_ENVIRONMENT to infer a GULP_ENVIRONMENT in order to build development assets on development environments and production assets in all other environments


galund:
cheeky 👍

PR #233, opened 2016-11-23
[#134679127]list_agreements: no longer re-sort supplierFrameworks in-view

risicle:
we should now be able to trust api will return them to us in desired order. tweak
some tests to assert this behaviour.


TheDoubleK:
:+1:

PR #234, opened 2016-11-29
encrypt tokens rather than just signing them

leohemsted:
so that personal emails aren't leaked to logs, analytics etc.

bump dmutils to 23.0.0

- [ ] https://github.com/alphagov/digitalmarketplace-utils/pull/288


pcraig3:
👍  wow, this looks awesome.

pcraig3:
I bet it definitely won't take down preview.

PR #235, opened 2016-12-01
Add Private ACL and download filename to admin-uploaded agreements

TheDoubleK:
Fixes this bug: https://www.pivotaltracker.com/story/show/134844463

Our bulk upload script sets ACL=Private and a human-friendly filename, but our admin upload page wasn't.

I've moved the generate download filename method to utils so it can be used by both scripts and this frontend route.

Depends on: 
 - [x] https://github.com/alphagov/digitalmarketplace-utils/pull/289

### DIFF
--- a/tests/app/main/views/test_suppliers.py
+++ b/tests/app/main/views/test_suppliers.py
@@ -971,7 +971,8 @@ class TestUploadCountersignedAgreementFile(LoggedInApplicationTest):
                 "onFramework": True,
                 "agreementStatus": "signed",
                 "agreementId": 1212,
-                "countersignedPath": None
+                "countersignedPath": None,
+                "declaration": {"nameOfOrganisation": "Supplier Mc Supply Face"},
             }}
         response = self.client.post('/admin/suppliers/1234/countersigned-agreements/g-cloud-7',
                                     data=dict(
@@ -986,13 +987,16 @@ class TestUploadCountersignedAgreementFile(LoggedInApplicationTest):
         )
 
         s3.S3.return_value.save.assert_called_once_with(
-            "g-cloud-7/agreements/1234/1234-countersigned-framework-agreement-2016-12-25-063001.pdf",
-            mock.ANY
+            "g-cloud-7/agreements/1234/1234-agreement-countersignature-2016-12-25-063001.pdf",
+            mock.ANY,
+            acl='private',
+            move_prefix=None,
+            download_filename='Supplier_Mc_Supply_Face-1234-agreement-countersignature.pdf'
         )
 
         data_api_client.update_framework_agreement.assert_called_once_with(
             1212,
-            {"countersignedAgreementPath": "g-cloud-7/agreements/1234/1234-countersigned-framework-agreement-2016-12-25-063001.pdf"},  # noqa
+            {"countersignedAgreementPath": "g-cloud-7/agreements/1234/1234-agreement-countersignature-2016-12-25-063001.pdf"},  # noqa
             'test@example.com'
         )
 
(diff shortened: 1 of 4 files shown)


tests/app/main/views/test_suppliers.py
```diff
@@ -1031,13 +1039,16 @@ def test_can_upload_countersigned_agreement_for_already_countersigned_agreement(
         data_api_client.approve_agreement_for_countersignature.assert_not_called()
 
         s3.S3.return_value.save.assert_called_once_with(
```
risicle: Offtopic but I think we should move away from using `assert_called_once_with` and family - they're too easy to get wrong and be interpreted as a do-nothing mocked call. Making sure an actual assert statement is evaluated by your own code seems to be the only way to make sure.

tests/app/main/views/test_suppliers.py
```diff
@@ -1046,13 +1057,77 @@ def test_can_upload_countersigned_agreement_for_already_countersigned_agreement(
...
+    response = self.client.post('/admin/suppliers/1234/countersigned-agreements/g-cloud-7',
+                                data=dict(
+                                    countersigned_agreement=(BytesIO(b"this is a test"),
+                                                             'countersigned_agreement.pdf'),
+                                ))
+
+    data_api_client.approve_agreement_for_countersignature.assert_called_once_with(
+        1212,
+        'test@example.com',
+        '1234'
```
risicle: Weird - why does this come through as a string?

tests/app/main/views/test_suppliers.py
```diff
@@ -1046,13 +1057,77 @@ def test_can_upload_countersigned_agreement_for_already_countersigned_agreement(
...
+    )
+
+    s3.S3.return_value.save.assert_called_once_with(
+        "g-cloud-7/agreements/1234/1234-agreement-countersignature-2016-12-25-063001.pdf",
+        mock.ANY,
```
risicle: Hmm I guess `assert_called_once_with` allows you to do this `mock.ANY` thing...

tests/app/main/views/test_suppliers.py
```diff
@@ -1046,13 +1057,77 @@ def test_can_upload_countersigned_agreement_for_already_countersigned_agreement(
...
+        user='test@example.com',
+        object_type='suppliers',
+        object_id=u'1234',
+        data={
+            'upload_countersigned_agreement':
+                'g-cloud-7/agreements/1234/1234-agreement-countersignature-2016-12-25-063001.pdf'
+        }
+    )
+
+    assert response.status_code == 302
```
risicle: I suspect that asserting the status code as the first check would produce more obvious errors quicker.

tests/app/main/views/test_suppliers.py
TheDoubleK: This is a weird thing, I remember being confused by it a while back.  It seems like this is (maybe) a Flask thing - `current_user.id` gives us a string, even though the user JSON serialization of Users from the API is clearly an integer.

TheDoubleK: This is where our logged-in user object gets created: https://github.com/alphagov/digitalmarketplace-utils/blob/master/dmutils/user.py

And it still looks to me like ID should be a string ¯\_(ツ)_/¯

TheDoubleK: OK< no it doesn't - out `get_id()` method here casts user id to a string:
https://github.com/alphagov/digitalmarketplace-utils/blob/master/dmutils/user.py#L38

I have no idea why :/

risicle:
I'm happy with this is you are :+1:

PR #236, opened 2016-12-02
Add new watcher for frameworks to make framework dev nicer

galund:
- Add new watcher for frameworks to make framework dev nicer
 - bower now copies only the frameworks YAML directory, not the entire frameworks
   repo, for consistency with the admin frontend


tombye:
👍

PR #239, opened 2016-12-23
[#134368805]allow all g7, g8 service fields to be edited by both `admin` and `admin-ccs-category`

risicle:
Story https://www.pivotaltracker.com/story/show/134368805

Being careful not to allow `admin-ccs-category` edit the `status` field. Also ended up fleshing out the tests covering this a bit.

### DIFF
--- a/tests/app/main/views/test_services.py
+++ b/tests/app/main/views/test_services.py
@@ -14,10 +14,11 @@ from ...helpers import LoggedInApplicationTest
 
 class TestServiceView(LoggedInApplicationTest):
     @mock.patch('app.main.views.services.data_api_client')
-    def test_service_view_with_no_features_or_benefits(self, data_api_client):
+    def test_service_view_no_features_or_benefits_status_disabled(self, data_api_client):
         data_api_client.get_service.return_value = {'services': {
             'frameworkSlug': 'g-cloud-8',
             'id': "314159265",
+            "status": "disabled",
         }}
         response = self.client.get('/admin/services/314159265')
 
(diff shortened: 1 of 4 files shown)


risicle:
Hmm travis is now passing for this but github doesn't appear to have picked that up yet.

TheDoubleK:
This is missing the "Edit" links on the service page for admin-ccs-category users:
## Admin view:
screen shot 2017-01-05 at 12 23 42

## Admin-ccs-category view:
screen shot 2017-01-05 at 12 23 28

TheDoubleK:
Should just need an amended condition here I think: https://github.com/alphagov/digitalmarketplace-admin-frontend/blob/master/app/templates/view_service.html#L69

tests/app/main/views/test_services.py
```diff
@@ -6,32 +6,86 @@
...
-            'frameworkSlug': 'g-cloud-8'
+            'frameworkSlug': 'g-cloud-8',
+            'id': "314159265",
+            "status": "disabled",
         }}
-        response = self.client.get('/admin/services/1')
+        response = self.client.get('/admin/services/314159265')
 
-        data_api_client.get_service.assert_called_with('1')
+        assert data_api_client.get_service.call_args_list == [
```
TheDoubleK: This feels less nice than `assert_called_with`, especially with the not-very-obvious empty dict in there. What's the motivation behind changing how we test the call here?

risicle: (the results of the little audit are in https://github.com/alphagov/digitalmarketplace-admin-frontend/pull/240)

Another thing that will help with this sort of problem is autospeccing of mocked `data_api_client`s which is something I'm trying to make some advance on in the great PR of doom that I dare not mention.

risicle: (those partial autospeccing-related changes are visible here for now https://github.com/alphagov/digitalmarketplace-admin-frontend/compare/ris-old-test-fw-purge...ris-app-apiclient-refactor)

TheDoubleK: Yep true - I've lost time to those non-existent methods before now.

TheDoubleK:
Reviewed - will be good to go once those Edit links are in.

TheDoubleK:
Cool, seems good to me now :+1:

Would be worth having a quick chat with Cath/Nicole before merging/releasing this - to find out if CCS Category need some warning/instruction on how to use it before it ships.

TheDoubleK:
I don't quite see how the policy annotation thing would work for cases like this where the view is accessible by various roles and the distinction is at the template level for what to show (e.g. links/sections)?  Wouldn't you just end up doing exactly the same logic in the template, only checking a proxy of the current_user.role rather than the current_user.role itself?

risicle:
> find out if CCS Category need some warning/instruction on how to use it before it ships.

What would the fun in that be?

PR #240, opened 2016-12-29
Purge use of old test frameworks

risicle:
It's happened. I've removed all uses of nose and unittest's `self.assert*` so that we will be able to in the future use py.test features such as parametrization.

~~Unfortunately this sits on top of #239 so will have to wait until it's been merged (and as such will also not pass its tests until https://github.com/alphagov/digitalmarketplace-frameworks/pull/349 is merged). But that's not really avoidable without me having to do a bunch of ugly rebasing.~~ merged now.

I've also done a bit of an audit of uses of the mock `*called*` properties/methods switching them to use safer ones. More information in the comment for that commit:

```
assert_called() does not exist (except in the *absolute latest* mock
packages) and doesn't do anything. assert_not_called() does not exist
(same caveat) and does not do anything. the only way to make absolutely
sure an assertion is being performed is to include the assert keyword. for
similar reasons identity comparisons of .called against True, False are
preferred over just asserting against the property's truthiness/falsiness.
```

### DIFF
--- a/tests/app/helpers.py
+++ b/tests/app/helpers.py
@@ -7,11 +7,10 @@ from flask import json
 
 from app import create_app
 from app import login_manager
-from unittest import TestCase
 
 
-class BaseApplicationTest(TestCase):
-    def setUp(self):
+class BaseApplicationTest(object):
+    def setup_method(self, method):
         self.app = create_app('test')
         self.client = self.app.test_client()
 
@@ -27,7 +26,7 @@ class BaseApplicationTest(TestCase):
         )
         self._default_suffix_patch.start()
 
-    def tearDown(self):
+    def teardown_method(self, method):
         self._s3_patch.stop()
         self._default_suffix_patch.stop()
 
@@ -59,8 +58,8 @@ class Response:
 class LoggedInApplicationTest(BaseApplicationTest):
     user_role = 'admin'
 
-    def setUp(self):
-        super(LoggedInApplicationTest, self).setUp()
+    def setup_method(self, method):
+        super(LoggedInApplicationTest, self).setup_method(method)
 
         patch_config = {
             'authenticate_user.return_value': {
@@ -94,9 +93,10 @@ class LoggedInApplicationTest(BaseApplicationTest):
                 user_id, 'test@example.com', None, None, False, True, 'tester', self.user_role
             )
 
-    def tearDown(self):
+    def teardown_method(self, method):
         self._data_api_client.stop()
         login_manager.user_loader(self._user_callback)
+        super(LoggedInApplicationTest, self).teardown_method(method)
 
     def _replace_whitespace(self, string, replacement_substring=""):
             # Replace all runs of whitespace with replacement_substring
--- a/tests/app/main/helpers/test_sum_counts.py
+++ b/tests/app/main/helpers/test_sum_counts.py
@@ -1,135 +1,102 @@
 # coding=utf-8
 from unittest import TestCase
-from nose.tools import assert_equal, assert_true, assert_false
 
 from app.main.helpers.sum_counts import _label_and_count, _sum_counts, _find
 
 
 class TestLabelAndCount(TestCase):
     def test_formatting(self):
-        assert_equal(
-            _label_and_count({}, {
-                'one': {
-                    'filter_one': None
-                },
-                'two': {
-                    'filter_one': False,
-                    'filter_two': True,
-                },
-                'three': {
-                    'filter_two': True
-                }
-            }, created_at='today'),
-            {'created_at': 'today', 'one': 0, 'two': 0, 'three': 0}
-        )
+        assert _label_and_count({}, {
+            'one': {
+                'filter_one': None
+            },
+            'two': {
+                'filter_one': False,
+                'filter_two': True,
+            },
+            'three': {
+                'filter_two': True
+            }
+        }, created_at='today') == {'created_at': 'today', 'one': 0, 'two': 0, 'three': 0}
 
 
 class TestSumCounts(TestCase):
     def test_summing_without_filtering(self):
-        assert_equal(
-            _sum_counts([
-                {'count': 1},
-                {'count': 2},
-                {'count': 3},
-                {'count': 4}
-            ]),
-            10
-        )
+        assert _sum_counts([
+            {'count': 1},
+            {'count': 2},
+            {'count': 3},
+            {'count': 4}
+        ]) == 10
 
     def test_summing_filtering_on_one_attribute(self):
-        assert_equal(
-            _sum_counts([
+        assert _sum_counts(
+            [
                 {'count': 1, 'include': False},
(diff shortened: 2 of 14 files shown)


risicle:
Hold on wait a second I've screwed up my rebase

risicle:
Ok all better.

tests/app/main/helpers/test_sum_counts.py
```diff
@@ -1,135 +1,102 @@
...
 
 class TestFind(TestCase):
     def test_find_string(self):
-        assert_true(
-            _find('word', 'word'),
-        )
-        assert_false(
-            _find('drow', 'word')
-        )
+        assert _find('word', 'word')
```
TheDoubleK: Might it be worth asserting `is True` and `is False` for all these `assert _find()` tests?  Probably fine as is though.

tests/app/main/views/test_service_updates.py
```diff
@@ -24,105 +24,80 @@ def test_should_render_activity_page_with_date(self, data_api_client):
...
-        self.assertEquals(200, response.status_code)
-        self.assertIn(
-            '<input class="filter-field-text" id="audit_date" name="audit_date" placeholder="eg, 2015-07-23" type="text" value="2006-01-01">',  # noqa
-            response.get_data(as_text=True)
-        )
+        assert response.status_code == 200
+        assert '<input class="filter-field-text" id="audit_date" name="audit_date" placeholder="eg, 2015-07-23" type='\
+            '"text" value="2006-01-01">' in response.get_data(as_text=True)
+
+        assert self._replace_whitespace(
```
TheDoubleK: The call to `_replace_whitespace` seems unnecessary here (and some other places later on) as there's clearly no whitespace in the string.  Keeps the comparison nice and symmetrical though I guess.

tests/app/main/views/test_service_updates.py
```diff
@@ -211,21 +180,12 @@ def test_should_redirect_to_update_page(self, data_api_client):
...
-        data_api_client.acknowledge_audit_event.assert_called(
-            audit_event_id=123,
-            user='test@example.com'
-        )
+        assert response.status_code == 302
+        assert 'http://localhost/admin/service-updates' in response.location
+        assert 'acknowledged=false' in response.location
+        assert 'audit_date=2010-01-05' in response.location
+
+        data_api_client.acknowledge_audit_event.assert_called_with("123", 'test@example.com')
```
TheDoubleK: The assortment of single and double quotes here bugs me for no real reason.

By your own earlier argument, would it be better to look at `call_args_list` here rather than relying on `assert_called_with`?  Swapping out all of those might be more than we need for this PR though.

tests/app/main/views/test_service_updates.py
```diff
@@ -302,33 +258,24 @@ def test_should_show_updates_if_valid_search(self, data_api_client):
...
-            self._replace_whitespace(
-               '<input name="audit_date" type="hidden" value="2010-01-01">'),  # noqa
-            self._replace_whitespace(response.get_data(as_text=True))
-        )
+        assert response.status_code == 200
+
+        assert self._replace_whitespace(
+            '<td class="summary-item-field-first"><span>Clouded Networks</span></td>'
+        ) in self._replace_whitespace(response.get_data(as_text=True))
+        assert self._replace_whitespace(
```
TheDoubleK: I'd be tempted to add blank newlines between each of these multi-line asserts to make it a bit easier to scan (the brackets previously broke it up a bit).

tests/app/main/views/test_suppliers.py
```diff
@@ -111,55 +103,21 @@ def test_should_show_user_details_on_page(self, data_api_client):
...
-            '<input type="submit" class="button-destructive"  value="Deactivate" />',
-            response.get_data(as_text=True)
-        )
-
-        self.assertIn(
-            '<form action="/admin/suppliers/users/999/deactivate" method="post">',
-            response.get_data(as_text=True)
-        )
+        assert response.status_code == 200
+        assert "Test User" in response.get_data(as_text=True)
```
TheDoubleK: It might be worth setting a variable something like `page_as_text = response.get_data(as_text=True)` to avoid calling `get_data()` over and over.  For some very minimal value of "worth".

tests/app/main/views/test_suppliers.py
```diff
@@ -313,39 +253,15 @@ def test_should_show_service_details_on_page(self, data_api_client):
...
-        self.assertIn(
-            '<a href="/admin/services/5687123785023488">',
-            response.get_data(as_text=True)
-        )
-        self.assertIn(
-            "Edit",
-            response.get_data(as_text=True)
-        )
+        assert response.status_code == 200
+        assert "Contract Management" in response.get_data(as_text=True)
```
TheDoubleK: Same here with all the calls to `get_data()`.

TheDoubleK:
:applause: :applause: :applause: :applause: :applause: 
So much better now.
:+1:

risicle:
Many thanks @TheDoubleK basically agreed on all points but I want to hit the big button before I'm made to rebase this again.

tests/app/helpers.py
```diff
@@ -7,11 +7,10 @@
 
 from app import create_app
 from app import login_manager
-from unittest import TestCase
 
 
-class BaseApplicationTest(TestCase):
-    def setUp(self):
+class BaseApplicationTest(object):
+    def setup_method(self, method):
```
galund: I don't really understand why these wouldn't just go into an `__init__` method?

risicle: I _think_ it's so you can distinguish between `setup_method` and `setup_class` - whether the state introduced can be shared between individual tests essentially.

Also, the analogous equivalent of `teardown_*` would be `__del__` which people are dissuaded from using much these days??

In short I'm not totally sure why test frameworks have historically preferred `setUp` style methods over "native object" interfaces. And as far as I'm concerned, `py.test` breaks a million and one standard python behaviours all over the place and little would surprise me these days. `setup_method` is just the blessed way in the docs I'm following.

galund: Ah, okay, I thought that these things could be called just once for all of the tests in a class, but I guess it's cleaner to reset for each one. (Which kinda begs the question why one uses a class at all, if no state is allowed to be shared between the method invocations, but eh.)

risicle: (it also strikes me that a lot of the complex pytest fixture behaviours could be achieved through multiple inheritance but whatever)

risicle:
> I thought that these things could be called just once for all of the tests in a class

I agree some of these things look like they could be. But this was meant as a mostly straight port of the existing `setUp` method, whose behaviour from the looks of it in the `unittest.TestCase` docs was akin to `setup_method`.

Using a class is still useful to group together a whole bunch of tests that are going to need the same things set up for them, even though they want to be isolated from each others collateral damage.

And then sometimes people just use the class as a structural/organizational thing.

PR #241, opened 2017-01-04
FIX-bug-introduced-by-csv-change

benvand:
Fix bug caused by CSV header change in 135223377-part-2-csv-header-change

CSV header was changed but method and tests here were not updated.
* Update method
* Update tests


pcraig3:
👍  once tests pass

PR #242, opened 2017-01-12
Admin edit dos2 declarations

TheDoubleK:
This adds the ability for users with the role `admin-ccs-sourcing` to view and edit DOS2 declarations.

It also removes the G-Cloud 7 column from their suppliers view, because work on agreements for that framework should be all finished now (although the pages will still be accessible via direct URL if they do need to do any G7 work).

**NOTE** ~~This will require changes to the functional tests, coming soon!~
Corresponding PR for functional tests is here:
 - [ ] https://github.com/alphagov/digitalmarketplace-functional-tests/pull/238

## BEFORE
screen shot 2017-01-12 at 06 20 08

## AFTER
screen shot 2017-01-12 at 06 13 42


pcraig3:
👍

PR #246, opened 2017-01-23
Wrap invitation link in anchor tag

TheDoubleK:
Some users have had issues with clicking the link in their invitation emails.
This *may* help alleviate the problem (and I have no other ideas how to address the issue).


allait:
👍

PR #247, opened 2017-01-24
[#137953083]: add DOS2 link to admin front page to approve agreements for countersigning

samuelhwilliams:
https://www.pivotaltracker.com/story/show/137953083

Removes hard-coded list of frameworks/links from the template; pulls from the API and shows link to list of agreements for all frameworks in standstill or live.

screen shot 2017-01-24 at 11 05 10

### DIFF
--- a/app/main/views/services.py
+++ b/app/main/views/services.py
@@ -21,7 +21,12 @@ from ..helpers.diff_tools import get_diffs_from_service_data, get_revision_dates
 @login_required
 @role_required('admin', 'admin-ccs-category', 'admin-ccs-sourcing')
 def index():
-    return render_template("index.html")
+    frameworks = data_api_client.find_frameworks()
+    frameworks = [fw for fw in frameworks['frameworks'] if fw['status'] in ('standstill', 'live')]
+    frameworks = sorted(frameworks, key=lambda x: x['id'], reverse=True)
+
+    return render_template("index.html",
+                           frameworks_for_countersigning=frameworks)
 
 
 @main.route('/services', methods=['GET'])
--- a/app/templates/index.html
+++ b/app/templates/index.html
@@ -110,26 +110,20 @@
     %}
       {% include "toolkit/page-heading.html" %}
     {% endwith %}
-    {% with items = [
+    {% for framework in frameworks_for_countersigning %}
+      {% with items = [
         {
-            "body": "Download G-Cloud 7 agreements",
-            "link": "/admin/agreements/g-cloud-7",
-            "title": "G-Cloud 7 agreements"
-        },
-        {
-            "body": "Download Digital Outcomes and Specialists agreements",
-            "link": "/admin/agreements/digital-outcomes-and-specialists",
-            "title": "Digital Outcomes and Specialists agreements"
-        },
-        {
-            "body": "Approve G-Cloud 8 agreements for countersigning",
-            "link": "/admin/agreements/g-cloud-8?status=signed",
-            "title": "G-Cloud 8 agreements"
-        },
+          "body": "Approve " + framework.name + " agreements for countersigning"
+                  if framework['frameworkAgreementVersion'] else
+                  "Download " + framework.name + " agreements",
+          "title": framework.name + " agreements",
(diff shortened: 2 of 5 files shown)


risicle:
@TheDoubleK I've suggested that this might be nicer to do more of the content-y bits in jinja rather than in python (essentially skipping the `body`/`link`/`title` dict generation entirely) - what do you think?

app/main/views/services.py
```diff
@@ -15,13 +15,26 @@
...
+        flash({'api_no_frameworks': True}, 'error')
+    else:
+        frameworks = [fw for fw in frameworks['frameworks'] if fw['status'] in ('standstill', 'live')]
+        frameworks = sorted(frameworks, key=lambda x: x['id'], reverse=True)
+        frameworks = [{'body': approve_or_download_sentence_for_framework(fw),
+                       'link': url_for('.list_agreements', framework_slug=fw['slug'], status='signed'),
+                       'title': '{name} agreements'.format(name=fw['name'])} for fw in frameworks]
+
+    return render_template("index.html",
+                           frameworks=frameworks)
```
TheDoubleK: This could be a bit more of a descriptive name - we generally have `frameworks` coming out of our ears, and these are not the usual `framework` objects that you'd generally expect.  So something like `frameworks_for_countersigning`.

Actually, I'd probably be tempted to pass the (sorted) frameworks list in to the template and do the `body`, `link`, `title` stuff (including the `if framework['frameworkAgreementVersion']` logic) in the index.html template itself.

app/main/helpers/service.py
```diff
@@ -6,3 +6,12 @@ def parse_document_upload_time(data):
     match = re.search("(\d{4}-\d{2}-\d{2}-\d{2}\d{2})\..{2,3}$", data)
     if match:
         return datetime.strptime(match.group(1), "%Y-%m-%d-%H%M")
+
+
+def approve_or_download_sentence_for_framework(framework):
```
TheDoubleK: I don't think this needs to be a helper method - we try to keep page content in the template files, and the conditional `if framework['frameworkAgreementVersion']` can be used in the template to choose between the two sentences.

app/templates/index.html
```diff
@@ -12,6 +12,8 @@
           {% set displayed_message = "Could not find a service with ID: {}".format(message['no_service']) %}
         {% elif message['api_error'] %}
           {% set displayed_message = "Error trying to retrieve service with ID: {}".format(message['api_error']) %}
+        {% elif message['api_no_frameworks'] %}
+          {% set displayed_message = "Could not retrieve frameworks." %}
```
TheDoubleK: I don't think this is worth flashing a message about - if we can't retrieve any frameworks then either the API is down and the page should probably just break, or there are no "standstill" or "live" frameworks and we all lost our jobs several months previously.

app/main/views/services.py
```diff
@@ -15,13 +15,26 @@
...
+from ..helpers.service import approve_or_download_sentence_for_framework
 
 
 @main.route('', methods=['GET'])
 @login_required
 @role_required('admin', 'admin-ccs-category', 'admin-ccs-sourcing')
 def index():
-    return render_template("index.html")
+    try:
+        frameworks = data_api_client.find_frameworks()
```
TheDoubleK: If there are no frameworks then something is very wrong - we probably shouldn't recover from that.

samuelhwilliams:
Made changes as requested.

samuelhwilliams:
Reverting slightly; putting the content back in services.py as jinja templating doesn't support list comprehensions and the toolkit browse-list.html is therefore creating a new <ul> for each item of the list, which is undesirable. Need to speak to Tom about possibly refactoring.

TheDoubleK:
As agreed in real life, we'll stick with this how it is :+1:

An aside: for future reference we generally prefer `rebase` over `merge` to bring PRs up to date - it avoids superfluous merge commits and guards against any bad guesses from git, and although not everyone is 100% convinced by these arguments it's the way we generally do things.

samuelhwilliams:
Noted, I will rebase in the future.

PR #249, opened 2017-01-27
Check draft services for lots if framework is not yet live

TheDoubleK:
https://www.pivotaltracker.com/story/show/138489291

Bugfix for admin app - lots need to be determined based on 'submitted' draft services, as we do not have actual services for DOS2.


samuelhwilliams:
👍

PR #250, opened 2017-01-31
[#134368805]service editing: handle validation error page redisplaying properly

risicle:
Story https://www.pivotaltracker.com/story/show/134368805

This should resolve @andrewchong 's remaining issues.

Some minor view changes to make sure we pass a full service object to the template and make sure the data is unformatted when we're redisplaying for validation errors.

Most of the work though expands the tests to cover this functionality which was previously untested.

### DIFF
--- a/tests/app/main/views/test_services.py
+++ b/tests/app/main/views/test_services.py
@@ -431,16 +431,16 @@ class TestServiceUpdate(LoggedInApplicationTest):
     def test_service_update_documents_with_validation_errors(
             self, data_api_client):
         data_api_client.get_service.return_value = {'services': {
-            'id': 1,
+            'id': 7654,
             'supplierId': 2,
             'frameworkSlug': 'g-cloud-7',
             'lot': 'SCS',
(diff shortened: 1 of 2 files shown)


tests/app/main/views/test_services.py
```diff
@@ -431,29 +431,111 @@ def test_service_update_documents(self, data_api_client):
...
+        assert document.xpath(
+            "//*[contains(@class,'validation-message')][contains(normalize-space(string()), $t)]",
+            t="Your document is not in an open format",
+        )
+        assert document.xpath(
+            "//a[normalize-space(string())=$t]/@href",
+            t="Return without saving",
+        ) == ["/admin/services/7654"]
+
+    # TODO test_service_update_non_assurance_questions_when_API_returns_error?
```
TheDoubleK: I think it's OK without this extra test.  But up to you - don't leave the `TODO` here though.

risicle:
Gracias Sr.

PR #251, opened 2017-01-31
Don't use apostrophes

TheDoubleK:
This commit exists only to prompt a merge to master so the release job will run properly.


idavidmcdonald:
👍

PR #255, opened 2017-02-22
Bring in new toolkit with new input fields

pcraig3:
This pull request

- updates the govuk_frontend toolkit to the latest we can get
- updates the digitalmarketplace frontend toolkit to the latest
- brings in the new radios/checkboxes
- treats links with a role="button" as buttons (ie, "Apply for this opportunity" link)
- removes some unused classnames
- deletes our `forms.html` file (these are super old macros we used to use before we had the frontend toolkit)
- brings the forgotten password link closer to the password input field on the login screen

💥  Ka-pow 💥


tombye:
👍

PR #259, opened 2017-03-09
[#141299531]signed agreements: "Cancel approval" button

risicle:
Story https://www.pivotaltracker.com/story/show/141299531

Feature in action:
20170309_cancelapproval

Things that I wonder:

- We use the text "**Accepted** by" and then "Cancel **approval**" which is a bit of a confusing mismatch - switch button to "Cancel acceptance"? "Accept" is the general term used in these pages.
- I decided to ignore the "next" behaviour entirely, including the `next_status` argument, though I now realize I might still want to at least pass through the `next_status` parameter so that the flow doesn't just totally forget what it was cycling through upon hitting "Cancel".
- Should a "Cancel" button perhaps *not* be green?

### DIFF
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -178,6 +178,9 @@ def approve_agreement_for_countersignature(agreement_id):
 @login_required
 @role_required('admin-ccs-sourcing')
 def unapprove_agreement_for_countersignature(agreement_id):
+    # not properly validating this - all we do is pass it through
+    next_status = request.args.get("next_status")
+
     agreement = data_api_client.unapprove_agreement_for_countersignature(
         agreement_id,
         current_user.email_address,
@@ -191,6 +194,7 @@ def unapprove_agreement_for_countersignature(agreement_id):
         '.view_signed_agreement',
         framework_slug=agreement["frameworkSlug"],
         supplier_id=agreement["supplierId"],
+        next_status=next_status,
     ))
 
 
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -174,6 +174,26 @@ def approve_agreement_for_countersignature(agreement_id):
     ))
 
 
+@main.route('/suppliers/agreements/<agreement_id>/unapprove', methods=['POST'])
+@login_required
+@role_required('admin-ccs-sourcing')
+def unapprove_agreement_for_countersignature(agreement_id):
+    agreement = data_api_client.unapprove_agreement_for_countersignature(
+        agreement_id,
+        current_user.email_address,
+        current_user.id,
+    )["agreement"]
+    organisation = request.form['nameOfOrganisation']
+    flash(u'The agreement for {} had its approval cancelled. You can approve it again at any time.'
+          .format(organisation), 'message')
+
+    return redirect(url_for(
+        '.view_signed_agreement',
+        framework_slug=agreement["frameworkSlug"],
+        supplier_id=agreement["supplierId"],
+    ))
+
+
 @main.route('/suppliers/<supplier_id>/agreement/<framework_slug>', methods=['GET'])
 @login_required
 @role_required('admin', 'admin-ccs-sourcing')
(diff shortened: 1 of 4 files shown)


TheDoubleK:
* I think harmonising the language would be good.  Whether we go with "Accept" or "Approve" as the thing, I'm not sure which is better, maybe @andymcaleer can help?  But should definitely be the same either way.

* If we can easily remember the "next" thing in the background then that might be nice.  But don't sweat it - it might teach people not to mess it up if they have to go back to the list and find their place :)
 
* I'd check with @laurenceberry or @ralph-hawkins but we have a red button for exactly this sort of thing: http://alphagov.github.io/digitalmarketplace-frontend-toolkit/button.html
Or even a grey button could work.  It definitely shouldn't be a green one, it shouldn't be a "call to action", rather a thing of last resort.

andymcaleer:
If we're using 'accept' as the verb in this bit of UI, and it looks like we are, then 'Cancel acceptance' is definitely what we should use. Nothing I've seen in the guidance contradicts this. Well spotted, Rob.

risicle:
Ok now looks like this:
20170310_cancelacceptance

The only remaining concern I have is that the "listing" view filters don't have a distinction between `approved` & `countersigned`, meaning you can't list them separately. This was fine when there was no functional difference between the two to the user, but now you can do different things to either of them. (what if someone needed to cancel all acceptences that were made today?).

Also in the listing view, one of the filters is named "Waiting for countersigning", which is a little confusing as I'd use that language to describe an agreement that is _approved_ but not yet _countersigned_. However it's actually used to mean the initial `signed` state. Should this filter instead be named "Waiting for acceptance"?

TheDoubleK:
The banner message seems pretty tight in to the PDF pane.  Also, I wonder if the banner should be "dstructive" (red) rather than green?
screen shot 2017-03-10 at 17 23 32

app/main/views/suppliers.py
```diff
@@ -174,6 +174,30 @@ def approve_agreement_for_countersignature(agreement_id):
...
+    # not properly validating this - all we do is pass it through
+    next_status = request.args.get("next_status")
+
+    agreement = data_api_client.unapprove_agreement_for_countersignature(
+        agreement_id,
+        current_user.email_address,
+        current_user.id,
+    )["agreement"]
+    organisation = request.form['nameOfOrganisation']
+    flash(u'The agreement for {} had its approval cancelled. You can approve it again at any time.'
```
TheDoubleK: I'd like to try to keep the message content in the templates where possible... I know there's a tech debt card to consolidate flash message handling, but it feels like in the templates is the best place to keep them for now still.

risicle: indeed however this is how all of its neighbour views do it so I thought I'd just not diverge here.

risicle:
As for using a "destructive" message, is there a danger that would make it look like the user's done something wrong? Maybe the in-between, marmalade-coloured message?

app/main/views/suppliers.py
galund: fwiw I personally think that having flash messages in the templates is a bit of an anti-pattern; it begs the question "which template?" Because you want the flash messages to be able to appear everywhere, that basically means having some special "lookup table" template included from all other templates.

risicle: I don't know where I stand on that one - but don't think here is the place to break formation.

TheDoubleK: There is a thing in the tech debt Trello about this - I think we should consolidate all flash messages to e.g. the base page template, keep the message content there, and make the way we pass flash messages consistent.  But no, this PR isn't the place to do it.

TheDoubleK: And yep, if the neighbour views do this then let's stick with it.

risicle:
20170313_cancelacceptance_bannerspacing

risicle:
The only thing I can see here remaining is my concern over the `approved`/`countersigned` split. Will have to ask @cathrooney I think.

risicle:
@TheDoubleK says the filtering is ok as it is. fine by me.

TheDoubleK:
:+1: good stuff.

PR #263, opened 2017-03-17
Update dev config to use a single bucket for all upload types

allait:
Using a single bucket allows us to use S3 bucket domain as our
assets URL without a running a separate nginx proxy.

The bucket is empty at the moment and might get replaced in the
future when we start managing it with Terraform.

Don't set bucket names for the test config since S3 should always
be mocked.


galund:
I think dmutils will need to be updated? (And then there are a few further 'upload_document' calls that will need updating I think.)

Wynndow:
👍

PR #267, opened 2017-04-19
Fix summary tables

carolinegreen:
Remove rowspan option where admin is not sourcing as these tables are different.

Before:
screen shot 2017-04-19 at 17 58 55

After:
screen shot 2017-04-19 at 17 56 55

### DIFF
--- a/app/templates/view_suppliers.html
+++ b/app/templates/view_suppliers.html
@@ -55,7 +55,11 @@
     %}
 
       {% call summary.row() %}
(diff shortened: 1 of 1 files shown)


TheDoubleK:
Feels a bit hacky, but yep let's try it :+1:

carolinegreen:
No it's not that hacky, we added the rowspan in.  That's how it works.  It's a bit messy with all the if elses but that's kind of a macro problem.

PR #269, opened 2017-05-02
G-Cloud supplier editing - to be acknowledged page

MuriloDalRi:
[Story](https://www.pivotaltracker.com/story/show/138029873)
As a CCS Category admin user, I need to see a list of services which have unacknowledged changes so that I can review them.

## Before
before

## After
after

[Story](https://www.pivotaltracker.com/story/show/138031127)
As a CCS Category admin user, I need to see a list of services that have already been acknowledged so that they can debug questions/issues.

new

### DIFF
--- a/app/main/views/service_updates.py
+++ b/app/main/views/service_updates.py
@@ -68,11 +68,11 @@ def service_update_audits():
     )
 
 
-@main.route('/services/updates', methods=['GET'])
+@main.route('/services/updates/acknowledged', methods=['GET'])
 @login_required
 @role_required('admin', 'admin-ccs-category')
 def acknowledged_services():
-    audit_events = data_api_client.find_audit_events(
+    audit_events_response = data_api_client.find_audit_events(
         audit_type=AuditTypes.update_service,
         acknowledged='true',
         latest_first='false',
@@ -81,7 +81,7 @@ def acknowledged_services():
 
     return render_template(
         "acknowledged_services.html",
-        audit_events=audit_events['auditEvents'],
+        audit_events=audit_events_response['auditEvents'],
     )
 
 
--- a/app/main/views/service_updates.py
+++ b/app/main/views/service_updates.py
@@ -97,7 +97,8 @@ def submit_service_update_acknowledgment(service_id, audit_id):
     if audit_event['acknowledged']:
         abort(410)
 
-    data_api_client.acknowledge_audit_event_including_previous(
+    data_api_client.acknowledge_service_update_including_previous(
+        service_id,
         audit_event["id"],
         current_user.email_address
     )
--- a/app/main/views/service_updates.py
+++ b/app/main/views/service_updates.py
@@ -52,33 +52,21 @@ def service_status_update_audits(day=None, page=1):
     )
 
 
-_SERVICE_UPDATE_AUDITS_EVENT_LIMIT = 1000
-
-
 @main.route('/services/updates/unacknowledged', methods=['GET'])
 @login_required
 @role_required('admin', 'admin-ccs-category')
 def service_update_audits():
-    audit_events = list(islice(data_api_client.find_audit_events_iter(
+    audit_events_response = data_api_client.find_audit_events(
         audit_type=AuditTypes.update_service,
         acknowledged='false',
         latest_first='false',
-    ), 0, _SERVICE_UPDATE_AUDITS_EVENT_LIMIT))
-    audit_events.sort(key=lambda audit_event: (audit_event["data"]["serviceId"], audit_event["createdAt"]))
-
-    # Build a tuple of the first of the earliest unacknowledged audit events for each service id
-    earliest_unacknowledged_audit_events = sorted(
-        (next(group_iter) for key, group_iter in groupby(
-            audit_events,
-            lambda audit_event: audit_event["data"]["serviceId"]
-        )),
-        key=lambda audit_event: audit_event["createdAt"],
+        earliest_for_each_object='true',
     )
 
     return render_template(
         "service_update_audits.html",
-        audit_events=earliest_unacknowledged_audit_events,
-        event_limit_reached=(len(audit_events) >= _SERVICE_UPDATE_AUDITS_EVENT_LIMIT),
+        audit_events=audit_events_response['auditEvents'],
+        has_next=bool(audit_events_response['links'].get('next')),
     )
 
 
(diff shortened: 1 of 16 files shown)


TheDoubleK:
Talking to @risicle - not originally in the story but this page should collapse multiple edits of the same service into a single row.

I'll add a note to the story now.

risicle:
@MuriloDalRi you've pushed the `tmp` commit which you should probably squash.

app/main/views/service_updates.py
```diff
@@ -54,33 +56,24 @@ def service_status_update_audits(day=None, page=1):
...
-        return render_template(
-            "service_update_audits.html",
-            today=datetime.utcnow().strftime(DATETIME_FORMAT),
-            acknowledged=form.default_acknowledged(),
-            audit_events=[],
-            form=form
-        ), 400
-
-    audit_events = data_api_client.find_audit_events(
+    audit_events = list(data_api_client.find_audit_events_iter(
```
allait: I'm very worried about this approach. This will paginate though the API, downloading all unacknowledged audit events, potentially making hundreds or thousands of requests based on how many edits we get and how often admins check them.

It feels like this could easily end up in a situation were this page just times out and there's no way for us to recover without marking events as acknowledged in the API.

As far as I remember we've discussed this with @TheDoubleK and decided to use archived services table to drive this page.

risicle:
Sorry I forgot on these ones we shouldn't be using `.next()` anymore - that's the py2 way of doing it. The py3 way of doing it is `next(<the iterable>)` and that way will work on py2 too if you do a `from six import next` at the top of the file.

app/main/views/service_updates.py
```diff
@@ -50,37 +52,60 @@ def service_status_update_audits(day=None, page=1):
     )
 
 
-@main.route('/service-updates', methods=['GET'])
+_service_update_audits_event_limit = 1000
```
samuelhwilliams: Would be nice if this was uppercase, as a global constant.

risicle: I guess the choice is either make it uppercase, remove the underscore and imply it's a dickable-with ... thing ... or imply that it's not.

samuelhwilliams: Eh, I don't think it implies it's dickable-with if it doesn't have a leading underscore. To me, global mutables are lowercase, global constants are uppercase. I don't mind if you have a leading underscore either way.

Google Python style guide (which I generally think is pretty good: http://google.github.io/styleguide/pyguide.html#Global_variables) agrees, although I accept that's not the source of truth for our repositories.

risicle: um cool yeah, gotta choose one either way

app/main/views/service_updates.py
risicle: So.. this isn't really about table sizes. It's about the inappropriateness of the api endpoint that's being used here.

I agree that as written it has the potential for performance issues, or actually rather just general crapness. In the worst case scenario the oldest 1000 edits were made to the same service, and we spend, like, 10 pages retrieving event descriptions that we don't care about (we only care about the earliest for each service) just to end up listing a single service that has changes. And maybe being able to say "oh, and an unspecified number of other edited services...".

This is mostly because we're doing the grouping in the front-end and can't tell the api not to waste its time sending back events we don't care about. Being able to do the "grouping"  (actually it would behave more like `DISTINCT ON` but could be implemented using either `GROUP BY` or `DISTINCT ON`) on the server-side might initially be a little difficult to fit into the mental model of how the call works (I would propose a flag such as `earliest_for_each_object=true`), but would allow the superfluous events to be omitted. (whatever table the information ended up on, an endpoint behaviour like this would be required somewhere to avoid this problem)

When it comes to performance, table size concerns can be hugely mitigated by creating a partial index on `service_update` event types and that partial index can also serve the purpose of providing a sorted input for either `GROUP BY` or `DISTINCT ON` to operate on. If we're being really clever it might even be possible to perform the grouping in a subquery that uses an index-only-scan, meaning the heap tuples only ever need to be accessed for the earliest event of each object. Possibly only for the <=100 that are going to returned in this request. So these requests would by-and-large only have to read a small-table-sized index to perform the bulk of their work. And as many writes as the `AuditEvent` table gets, they're almost all `INSERT` and likely all end up grouped together in the same pages at the end of the table, leaving the visibility map comparatively clean, hopefully not hindering the ability to use an index only scan.

We actually already have an regular multi-column index on `AuditEvent` which would probably go some way towards making this query fast without any extra work. Even without an index only scan, an index can easily identify the `service_update` events.

If we want to look at reasons we get an api slowdown, table-size will be dwarfed by e.g. 
- use of `LIMIT .. OFFSET` as a paging mechanism, which only discards rows *after* it has pulled all the data for the possibly-hundreds-of-pages-long response out of the heap
- overzealous eager-joining, which is rife because we been defaulting to (I'm guilty of this too) eager joins on new relationships.
- the amount of time & memory we spend jsonifying & de-jsonifying data (json blobs in the tables need to be deserialized as they come out too). I'm not joking about this - json (de-)serialization is not particularly cheap and we would probably save significant amounts of both resources by just plugging in something such as https://pypi.python.org/pypi/ujson.

Of course, we're unable to talk about this truly scientifically and this text is full of "probably, hopefully..."s, because we don't have any proper profiling on a database with a live-like dataset on the same hardware. In an ideal world we'd be able to see exactly how much time a request spent performing each of these stages.

As to why *not* put the data on the `ArchivedService`?
- firstly it's more unnecessary denormalization of data we have elsewhere.
- it breaks a nice pattern we could set up for reviewing different types of events. It doesn't take the wildest mind to imagine a desire to monitor updates made to briefs, brief responses, supplier contact details... possibly even if we were to extend collaborative abilities to let suppliers see edits made to `DraftService`s during the submission phase, god knows...
- the interfaces are there. There currently isn't even a *basic* "list archived services" endpoint. That would be a lot of extra needless work to implement. And it's work that would need to be repeated any time we wanted to provide a similar change-monitoring ability for a new model. And inevitably it would be a different person implementing each one and we'd end up with something where none of those interfaces were *quite* alike for no good reason.

allait: I think we're largely on the same page on this. The API endpoint was never meant to support this use case, so building this directly on top of it doesn't feel like a good approach (for all the reasons you've mentioned). Same goes for https://github.com/alphagov/digitalmarketplace-admin-frontend/pull/281/files, which is trying to collect events for an individual service.

I think as an alternative, having a separate collection of API endpoints (eg `/service-updates` would be a simple option, but we can think of what it would take to build a more general one to cover other types of updates as well) with a clear interface would go a long way towards addressing this. Or we can look into extending the interface of the existing endpoint if it fits better.

I'm not particularly committed to the idea of using `ArchivedServices` or denormalizing this data either. I'd be more than happy if we can get it to work with the existing data model. But as you know we don't have a lot of examples of similar queries against large tables or a robust set of performance metrics we can compare against, so there are a lot of assumptions going into this both in terms of performance of the admin page / API endpoint / DB queries themselves and of their effect on the rest of the apps. This doesn't mean that we can't do this, but if we're going to then we do need to explicitly highlight the risks and our expectations, make a plan to test, measure and validate them before we release and ensure that we'll be able (including having time in the future sprints) to fix any issues that arise if our assumptions don't hold.

I'm going to schedule a chat on Monday so we can all talk about this.

risicle: So, yes, this does indeed start to feel like using an api that wasn't designed for this, but then again specializing down into a set of endpoints for "service updates" starts to go in the direction of "backend has special purpose endpoint for each frontend endpoint" which is very... non-api-y. I *think* simply adding a "first for each object" option on the listing endpoint would achieve all we need for this view and might have uses elsewhere.

When it comes to the whole "people making edits faster than we can keep up with them", I assume we're going to have to put up with the idea of some edits being live for a couple of days anyway what with CCS having weekends. And a couple of days is enough for us to fix anything (right?).

I was thinking about being able to pre-test the setup and reckoned the cheapest/easiest way to do it for now for lack of a similar testing setup would be to roll out some conservative api changes to production and test e.g. hitting it ten times in a row. We *do have* some service update events in the log at the moment to give us an example dataset - I don't think any database writes would be needed.

risicle: p.s. interestingly, the only reason #281 has to fetch all the unacknowledged events at all is just to be able to give a summary message around e.g. how many different users edited it.

TheDoubleK:
There's a new branch with all the changes in for this feature so closing this.

PR #272, opened 2017-05-09
Remove plain text logs in live

Wynndow:
Trello card: https://trello.com/c/o3Jkp3rK

25.2.0 of utils will create json logs by default, unless the
DM_PLAIN_TEXT_LOGS config variable is set to True in the config.

This config variable is set in dev and test so that logs are more
legible/useful.

The other thing 25.2.0 does is log to stdout instead of file by default,
unless the DM_LOG_PATH config variable is set. In the PaaS
environment, we'll need to log to stdout so will overwrite this
variable. Logging to stdout is also more useful in dev and test.

### DIFF
--- a/config.py
+++ b/config.py
@@ -37,8 +37,9 @@ class Config(object):
 
     # Logging
     DM_LOG_LEVEL = 'DEBUG'
-    DM_APP_NAME = 'admin-frontend'
+    DM_PLAIN_TEXT_LOGS = False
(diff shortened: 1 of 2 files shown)


Wynndow:
[Remove plaintext logger in live environments](https://trello.com/c/o3Jkp3rK/63-remove-plaintext-logger-in-live-environments)

PR #279, opened 2017-05-25
Update staging config

allait:
### Fix missing DM_HTTP_PROTO in staging config

Staging was sending user invite emails with `http://` because staging
config didn't set the correct DM_HTTP_PROTO.

The config was copied from Live in #87 to disable CSRF checks in
staging. We couldn't find a reason why we'd still want this, so
I'm reverting the setting to match the Live config.

### Remove DM_LOG_PATH config variable

After the move to PaaS, we're no longer logging to files in any
environments (DM_LOG_PATH is overwritten with an empty string in
all PaaS spaces)

### Replace DM_DOCUMENTS_URL with DM_ASSETS_URL

DM_DOCUMENTS_URL was used before we introduced DM_ASSETS_URL.
Both variables should contain the same value, but DM_ASSETS_URL is
set through environment variables to the correct value for each
stage.

### DIFF
--- a/config.py
+++ b/config.py
@@ -38,7 +38,6 @@ class Config(object):
     # Logging
     DM_LOG_LEVEL = 'DEBUG'
     DM_PLAIN_TEXT_LOGS = False
-    DM_LOG_PATH = None
     DM_APP_NAME = 'admin-frontend'
     DM_DOWNSTREAM_REQUEST_ID_HEADER = 'X-Amz-Cf-Id'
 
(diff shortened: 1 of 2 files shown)


config.py
```diff
@@ -86,7 +84,6 @@ class Development(Config):
     DM_AGREEMENTS_BUCKET = 'digitalmarketplace-dev-uploads'
     DM_COMMUNICATIONS_BUCKET = 'digitalmarketplace-dev-uploads'
     DM_S3_DOCUMENT_BUCKET = "digitalmarketplace-dev-uploads"
-    DM_DOCUMENTS_URL = "https://{}.s3-eu-west-1.amazonaws.com".format(DM_S3_DOCUMENT_BUCKET)
```
TheDoubleK: Do we not need to set this S3 URL for local dev still?  Where does the dev assets url come from otherwise?

allait: Ah, good catch! Renamed this to `DM_ASSETS_URL` instead of removing. Thanks Kev!

PR #281, opened 2017-06-01
[#138031851]service updates page

risicle:
Story https://www.pivotaltracker.com/story/show/138031851

In action:
20170601_service_updates
20170601_service_updates_reversed

This should probably be rebased after https://github.com/alphagov/digitalmarketplace-admin-frontend/pull/269 is merged and modified to accommodate its changes. (The tests will fail until we've done this patchup because of tests of old views trying to link to the old comparison view)

### DIFF
--- a/app/assets/scss/_diff.scss
+++ b/app/assets/scss/_diff.scss
@@ -1,29 +1,10 @@
 .diff {
   margin-bottom: $gutter;
 
-  .diff-title {
-    @include heading-27;
-    font-weight: bold;
-    margin-bottom: $gutter-half;
-  }
-
-  .revision-dates {
-    @extend %contain-floats;
-    margin-bottom: $gutter-half;
-
-    p {
-      float: none;
-      width: 100%;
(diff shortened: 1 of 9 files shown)


samuelhwilliams:
This looks cool. Just wondering if it's possible to make the diff lib generate slightly more coarse diffs; it would probably be easier to understand what's changed at a glance if it said that `bums on` had been removed and `position in` had been added, rather than some letters within each word.

risicle:
I agree that would be nice, but I don't think there's a totally straightforward way of doing it. Probably possible using some more complex post-processing or with some monkeypatching.

risicle:
(damn, I *thought* I had thought of a way of doing it yesterday but realized some gotchas today)

risicle:
This is going to have a couple of implementation changes made to it which I haven't done yet...

risicle:
I've made those changes but I'm going to close this PR because I think this is all going to get balled together in a bigger PR which can get all the URLs wired up properly.

PR #283, opened 2017-07-04
Flake8

benvand:
Fix flake8 violations detailed in the attached:
[flake8_violations.txt](https://github.com/alphagov/digitalmarketplace-admin-frontend/files/1122468/flake8_violations.txt)

### DIFF
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -89,25 +89,16 @@ def view_signed_agreement(supplier_id, framework_slug):
     if not supplier_framework.get('agreementReturned'):
         abort(404)
 
-    # build an OrderedDict of applied-for lotSlug against lotName, ordered by lotSlug
     if framework["status"] in ("live", "expired"):
-        lot_slugs_names = OrderedDict(sorted(
-            (service["lotSlug"], service["lotName"],)
-            for service in data_api_client.find_services_iter(
-                supplier_id=supplier_id,
-                framework=framework_slug,
-                )
-            )
-        )
+        # If the framework is live or expired we don't need to filter drafts, we only care about successful services
+        service_iterator = data_api_client.find_services_iter(supplier_id=supplier_id, framework=framework_slug)
+        lot_slugs_names = [(service["lotSlug"], service["lotName"]) for service in service_iterator]
     else:
-        lot_slugs_names = OrderedDict(sorted(
-            (service["lotSlug"], service["lotName"],)
-            for service in data_api_client.find_draft_services_iter(
-                supplier_id=supplier_id,
-                framework=framework_slug,
-                ) if service["status"] == "submitted"
-            )
-        )
+        # If the framework has not yet become live we need to filter out unsuccessful services
+        service_iterator = data_api_client.find_draft_services_iter(supplier_id=supplier_id, framework=framework_slug)
+        lot_slugs_names = [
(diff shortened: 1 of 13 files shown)


app/main/views/suppliers.py
```diff
@@ -93,25 +93,16 @@ def view_signed_agreement(supplier_id, framework_slug):
     if not supplier_framework.get('agreementReturned'):
         abort(404)
 
-    # build an OrderedDict of applied-for lotSlug against lotName, ordered by lotSlug
     if framework["status"] in ("live", "expired"):
```
benvand: Don't do this as a one liner (assign iterator and _then_ construct list) and DRY (OrderedDict, sorted).

app/main/views/suppliers.py
```diff
@@ -381,7 +372,8 @@ def edit_supplier_declaration_section(supplier_id, framework_slug, section_id):
     '/suppliers/<string:supplier_id>/edit/declarations/<string:framework_slug>/<string:section_id>',
     methods=['POST'])
 def update_supplier_declaration_section(supplier_id, framework_slug, section_id):
```
benvand: unused assignment

app/main/views/suppliers.py
```diff
@@ -381,7 +372,8 @@ def edit_supplier_declaration_section(supplier_id, framework_slug, section_id):
...
-    supplier = data_api_client.get_supplier(supplier_id)['suppliers']
+    # Supplier must exist.
+    data_api_client.get_supplier(supplier_id)
```
samuelhwilliams: This feels really weird to me for a couple reasons.

1) It would be nice to catch this exception and 404 a little more explicitly, rather than relying on the API client to terminate the request through an invisibly-handled HTTPError. (Not really anything to do with your flake8 stuff, just a side thought)
2) Reading this line of code is confusing. We get something and don't store it or use it for anything; it would be easy to assume this line of code isn't even needed. It's also less consistent with surrounding code. Personally I'd rather we just keep the unused reference around, but I guess you'll disagree with that. In which case, a much better comment should be fine.

tests/app/main/views/test_services.py
```diff
@@ -786,11 +786,10 @@ def setup_method(self, method):
...
 
-        self._service_not_found = {
-            'error': 'The requested URL was not found on the server.  If you '
-                     'entered the URL manually please check your spelling and '
-                     'try again.'
-            }
+        self._service_not_found = {'error': (
+            'The requested URL was not found on the server.  '
+            'If you entered the URL manually please check your spelling and try again.'
+        )}
```
samuelhwilliams: Huh. What flake8 rule is this? It feels weird.

app/main/views/suppliers.py
samuelhwilliams: This feels really weird to me for a couple reasons.

1) It would be nice to catch this exception and 404 a little more explicitly, rather than relying on the API client to terminate the request through an invisibly-handled HTTPError. (Not really anything to do with your flake8 stuff, just a side thought)
2) Reading this line of code is confusing. We get something and don't store it or use it for anything; it would be easy to assume this line of code isn't even needed. It's also less consistent with surrounding code. Personally I'd rather we just keep the unused reference around, but I guess you'll disagree with that. In which case, a much better comment should be fine.

PR #284, opened 2017-07-05
Add default.nix

risicle:
See https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/719

This should be near-identical, probably best to have any discussion there.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -66,3 +66,6 @@ app/content
 
 # Front end generated assets
 app/static
+
+# local environment customizations
+local.nix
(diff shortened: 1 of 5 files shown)


risicle:
See https://github.com/alphagov/digitalmarketplace-manual/pull/6 for proposed corresponding manual page.

risicle:
I will rebase this PR after https://github.com/alphagov/digitalmarketplace-admin-frontend/pull/288 has gone in.

TheDoubleK:
#288 is in, good luck with that.

:+1:

PR #287, opened 2017-07-14
Pin requirements

allait:
**NOTE** this is removing a few old and seemingly unused packages (Flask-Bootstrap, pbkdf2) and bumping utils to 27.1.2, which includes the new boto3-based S3 class. Tests are passing locally, but it's worth checking the file uploads are working correctly once this goes to preview.

requirements.txt is generated using a Python 3 virtualenv to keep the py2 backports of Python 3 stdlib packages (like `enum34` and `ipaddress`) out. This leaves those packages unpinned, but shouldn't cause any issues otherwise.

### Reorganize Makefile and requirements files

Renames requirements.txt to requirements-app.txt and requirements_for_test.txt to requirements-dev.txt.

Removes any unused dependencies or dependencies provided by other packages since those will be listed in the full requirements.txt.

Updates Makefile to add freeze-requirements and rename targets to be more consistent with other apps.

(see alphagov/digitalmarketplace-api#615 for details)


Makefile
```diff
@@ -8,49 +8,79 @@ else
...
-test_pep8: virtualenv
+.PHONY: test-pep8
+test-pep8: virtualenv
 	${VIRTUALENV_ROOT}/bin/pep8 .
 
-test_python: virtualenv
-	${VIRTUALENV_ROOT}/bin/py.test ${PYTEST_ARGS}
+.PHONY: test-python
+test-python: virtualenv
+	${VIRTUALENV_ROOT}/bin/py.test ${PYtest-ARGS}
```
Wynndow: Is the lower case here intentional? All uppercase is used in the `.travis.yml`

TheDoubleK:
Tests will pass locally anyway I think because the S3 stuff ends up being mocked out in all the tests. Would be worth a grep for methods affected by the breaking changes in the code too.

allait:
@TheDoubleK I've checked document and communication uploads and agreements countersigning. Communication uploads had to be updated to fix the "Last updated" timestamp, but they seem to show the oldest file date (instead of the newest one), so the date doesn't reflect the actual last upload time. I'm not sure if that's a bug or intended behaviour, but I'm keeping it as is.

TheDoubleK:
:+1:

PR #282, opened 2017-06-09
[#146370407] Admin frontend - route to acknowledge changes to a service

MuriloDalRi:
https://www.pivotaltracker.com/story/show/146370407

This assumes the interface to acknowledge services in the data_api_client.

PR #288, opened 2017-07-18
Admin view and approval of edits to services

TheDoubleK:
This supersedes the following old PRs, with (we believe) fixes to address all the comments previously made:
https://github.com/alphagov/digitalmarketplace-admin-frontend/pull/269
https://github.com/alphagov/digitalmarketplace-admin-frontend/pull/281
https://github.com/alphagov/digitalmarketplace-admin-frontend/pull/282

The PR messages and discussion in those places is probably worth looking at for context.

The relevant Pivotal tickets are: 
https://www.pivotaltracker.com/story/show/138029873 (List of edited services)
https://www.pivotaltracker.com/story/show/138031851 (Detailed view of services)
https://www.pivotaltracker.com/story/show/146370407 (POST to approve edits)

Reviewers - ask in real life for demos and more detail if you need it.


samuelhwilliams:
There doesn't seem to be a flash message indicating that the action (acknowledging edits) was successful; feels like we should probably add one.

app/main/helpers/diff_tools.py
```diff
@@ -1,209 +1,132 @@
...
 
-    for word in words:
-        type = get_word_type(word)
-        if type in types_to_look_for:
-            return type
+def _question_iter(question):
+    # type testing is weird - this should eventually belong in the question subclass itself
+    if isinstance(question, Multiquestion):
+        for sub_question in question.questions:
+            yield sub_question
```
samuelhwilliams: You will need to recurse here as we have some multiquestions that contain multiquestions.

app/main/views/service_updates.py
```diff
@@ -50,59 +48,40 @@ def service_status_update_audits(day=None, page=1):
...
-            form=form
-        ), 400
-
-    audit_events = data_api_client.find_audit_events(
+    audit_events_response = data_api_client.find_audit_events(
         audit_type=AuditTypes.update_service,
-        acknowledged=form.default_acknowledged(),
-        audit_date=form.format_date(),
-        page=form.page.data
+        acknowledged='false',
```
samuelhwilliams: Weird stringy bool.

app/main/views/services.py
```diff
@@ -119,72 +123,96 @@ def edit(service_id, section_id):
...
+        oldest_update_events_response = data_api_client.find_audit_events(
+            latest_first="false",
+            **common_request_kwargs
+        )
+        oldest_update_events = oldest_update_events_response["auditEvents"][::-1]
+
+        # now we check to see if we're able to splice together the full list using these two responses. we'll know this
+        # if we are able to find the last of latest_update_events in oldest_update_events. if we're not able to, we'll
+        # have to leave all_update_events as None to signal that we don't have all the events
+        if latest_update_events[-1]["id"] in (ae["id"] for ae in oldest_update_events):
```
samuelhwilliams: Can we just call this `audit_event` rather than `ae`?

app/templates/compare_revisions.html
```diff
@@ -41,89 +45,96 @@
...
-              <label class="selection-button">
-                  <input type="radio" name="service_status" id="service_status_private" value="private" {% if service_data['status'] == 'enabled' %}checked="checked"{% endif %} />
-                  Private
-              </label>
-              {% if service_data.status != 'disabled' %}
-              <label class="selection-button">
-                  <input type="radio" name="service_status" id="service_status_published" value="public" {% if service_data['status'] == 'published' %}checked="checked"{% endif %} />
-                  Public
-              </label>
+          unapproved {{ pluralize((all_update_events or (oldest_update_events+latest_update_events))|length, "edit", "edits") }}
```
samuelhwilliams: This is a weird function.

TheDoubleK: Looking at the commit message in utils it apparently "works in a similar way to Rails's pluralize function", which might explain the perceived weirdness :smile:

app/main/views/services.py
```diff
@@ -119,72 +123,96 @@ def edit(service_id, section_id):
...
     )
+    latest_update_events = latest_update_events_response["auditEvents"]
+
+    if latest_update_events_response["links"].get("next"):
+        # we haven't got all update events for this object. fetch the oldest
+        oldest_update_events_response = data_api_client.find_audit_events(
+            latest_first="false",
+            **common_request_kwargs
+        )
+        oldest_update_events = oldest_update_events_response["auditEvents"][::-1]
```
samuelhwilliams: I don't know if it's sensible to reverse the order of these events. I would expect `latest_update_events` to have the most recent event at `0`, which it does, and I would expect `oldest_update_events` to have the oldest event at `0`, which it no longer does. It has the most recent event at `0` after this mutation. Unless there's a good reason for doing this, it's probably better to leave it in the order it's received from the API or it's going to be easy to be confused about the order.

TheDoubleK: The reverse makes sense to me - it ensures that both lists of audit events are in the same chronological order (i.e. from newest to oldest), which allows `all_update_events` to be made by splicing the two lists together at the crossover point (if it exists).

To my mind marshalling two lists of audit events that are ordered chronologically in opposite directions is likely to be more error-prone than remembering which list goes in which direction.

The reason for doing things this way at all (rather than just getting **all** the relevant audit events for the service together in the same order) is to limit the number of API calls required to get the audit event data to at most two.  So this is exactly mimicking what `data_api_client.find_audit_events_iter(latest_first=True)` would give us, without making a potentially-unbounded number of API calls.

samuelhwilliams:
I still disagree re: audit events, but let's go ahead with this as it's two v one. :+1:

app/main/views/service_updates.py
risicle: Don't blame me.

risicle:
If you're referring to the ordering of `oldest_update_events` it has to be that way around because the intention is to graft the two ends together in cases we can. So it's really just a question of what it's *called*.

samuelhwilliams:
The order doesn't prevent you grafting two ends together because you can just change the order at the point you do the grafting. The nomenclature might be a factor, but it's a different issue. Really, to me it's down to the fact we're mutating data received from the API into a non-standard format (in this instance, reversing the chronology) and whether doing that is worth it. :man_shrugging: It's not worth fighting over, it just feels a bit off to me.

risicle:
> it just feels a bit off to me.

I won't say it

PR #292, opened 2017-08-17
Ensure existing behaviour for new 'awarded' brief status

idavidmcdonald:
As part of this trello story - https://trello.com/c/kGVfzXuH/749-3-collect-award-status-on-buyer-dashboard - we are adding a new 'awarded' status for a brief.

This PR ensures the admin app retains all existing functionality when this new status lands.

### DIFF
--- a/app/main/views/users.py
+++ b/app/main/views/users.py
@@ -14,6 +14,8 @@ from .. import main
 from ... import data_api_client
 from ..auth import role_required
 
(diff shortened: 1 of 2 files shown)


katstevens:
As per https://github.com/alphagov/digitalmarketplace-brief-responses-frontend/pull/2, I've put the status list into a constant so we can easily add a new status for the 'no' path.

PR #293, opened 2017-08-25
Backwards compatibility for new Brief statuses

katstevens:
First of 5 backwards compatibility PRs to handle the two new Brief statuses `cancelled` and `unsuccessful`.

1) Admin FE
2) Buyer FE https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/588
3) Scripts https://github.com/alphagov/digitalmarketplace-scripts/pull/144
4) Brief Responses FE https://github.com/alphagov/digitalmarketplace-brief-responses-frontend/pull/5
5) Briefs FE https://github.com/alphagov/digitalmarketplace-briefs-frontend/pull/17

See https://github.com/alphagov/digitalmarketplace-admin-frontend/pull/292/ for the equivalent changes for the `awarded` status.

As I was unfamiliar with this repo I thought I'd have a look through the tests and include some extra  coverage (mostly straightforward error path checks). I left alone anything even slightly complex 😸 

Finally there's some PEP8 imports cleanup, because why not.

### DIFF
--- a/tests/app/main/views/test_agreements.py
+++ b/tests/app/main/views/test_agreements.py
@@ -203,6 +203,12 @@ class TestListAgreements(LoggedInApplicationTest):
 
         assert response.status_code == 403
 
+    def test_invalid_status_raises_400(self, data_api_client):
+
+        response = self.client.get('/admin/agreements/g-cloud-7?status=bad')
+
+        assert response.status_code == 400
+
 
 @mock.patch('app.main.views.agreements.data_api_client')
 class TestNextAgreementRedirect(LoggedInApplicationTest):
@@ -337,3 +343,7 @@ class TestNextAgreementRedirect(LoggedInApplicationTest):
             (ca_kwargs.get("with_declarations") is False and not ca_kwargs.get("status"))
             for ca_args, ca_kwargs in data_api_client.find_framework_suppliers.call_args_list
         )
+
+    def test_invalid_status_raises_400(self, data_api_client):
+        response = self.client.get('/admin/suppliers/151/agreements/g-cloud-8/next?status=bad')
+        assert response.status_code == 400
--- a/tests/app/main/views/test_stats.py
+++ b/tests/app/main/views/test_stats.py
@@ -23,6 +23,18 @@ class TestStats(LoggedInApplicationTest):
 
         assert response.status_code == 200
 
+    def test_get_stats_page_for_open_framework_includes_framework_stats(self, data_api_client):
+        data_api_client.find_audit_events.return_value = {
+            'auditEvents': []
+        }
+        data_api_client.get_framework.return_value = {
+            'frameworks': {'status': 'open', 'lots': []}
+        }
+        response = self.client.get('/admin/statistics/g-cloud-7')
+
+        assert data_api_client.get_framework_stats.call_args == mock.call('g-cloud-7')
+        assert response.status_code == 200
+
     def test_supplier_counts_on_stats_page(self, data_api_client):
         data_api_client.find_audit_events.return_value = {
             "auditEvents": [
--- a/tests/app/main/views/test_stats.py
+++ b/tests/app/main/views/test_stats.py
@@ -14,7 +14,7 @@ class TestStats(LoggedInApplicationTest):
         }
         response = self.client.get('/admin/statistics/g-cloud-7')
 
-        data_api_client.find_audit_events.assert_called_with(
+        data_api_client.find_audit_events.assert_called_once_with(
             audit_type=AuditTypes.snapshot_framework_stats,
             object_type='frameworks',
             object_id='g-cloud-7',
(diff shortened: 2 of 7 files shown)


katstevens:
Strange, that fix seems to introduce a failure for me locally (and on Travis). It passes without the fix. Is your local `venv` up to date?

CrystalPea:
Ok, I rolled it back, I  must have old version of 'forms' component or something.

tests/app/main/views/test_agreements.py
```diff
@@ -203,6 +203,12 @@ def test_unauthorised_roles_are_rejected_access(self, data_api_client):
 
         assert response.status_code == 403
 
+    def test_invalid_status_raises_400(self, data_api_client):
```
benvand: All spaces here

tests/app/main/views/test_stats.py
```diff
@@ -23,6 +23,18 @@ def test_get_stats_page(self, data_api_client):
...
+    def test_get_stats_page_for_open_framework_includes_framework_stats(self, data_api_client):
+        data_api_client.find_audit_events.return_value = {
+            'auditEvents': []
+        }
+        data_api_client.get_framework.return_value = {
+            'frameworks': {'status': 'open', 'lots': []}
+        }
+        response = self.client.get('/admin/statistics/g-cloud-7')
+
+        assert data_api_client.get_framework_stats.call_args == mock.call('g-cloud-7')
```
benvand: This can use `called_once_with`

tests/app/main/views/test_suppliers.py
```diff
@@ -174,6 +174,21 @@ def test_should_call_api_to_activate_user(self, data_api_client):
...
+    def test_should_call_api_to_activate_user_and_redirect_to_source_if_present(self, data_api_client):
+        data_api_client.get_supplier.return_value = self.load_example_listing("supplier_response")
+        data_api_client.update_user.return_value = self.load_example_listing("user_response")
+
+        response = self.client.post(
+            '/admin/suppliers/users/999/activate',
+            data={'source': "http://example.com"}
+        )
+
+        data_api_client.update_user.assert_called_with(999, active=True, updater="test@example.com")
```
benvand: This can also use `called_once_with`

tests/app/main/views/test_suppliers.py
```diff
@@ -1102,35 +1139,45 @@ def test_should_404_if_agreement_not_returned(self, s3, data_api_client):
         data_api_client.get_framework.assert_called_with('g-cloud-8')
         data_api_client.get_supplier_framework_info.assert_called_with('1234', 'g-cloud-8')
 
+    def test_should_404_if_agreement_has_no_version(self, s3, data_api_client):
+        data_api_client.get_supplier.return_value = self.load_example_listing('supplier_response')
+        data_api_client.get_framework.return_value = {'frameworks': {}}
+        response = self.client.get('/admin/suppliers/1234/agreements/g-cloud-8')
+
+        assert response.status_code == 404
+        data_api_client.get_supplier.assert_called_with('1234')
```
benvand: These 2 can be called once with

tests/app/main/views/test_suppliers.py
```diff
@@ -1102,35 +1139,45 @@ def test_should_404_if_agreement_not_returned(self, s3, data_api_client):
...
+        data_api_client.get_framework.assert_called_with('g-cloud-8')
+
+    def _find_services_iter_side_effect(self, *args, **kwargs):
```
benvand: I'd be tempted to reqrite this as
```
data_api_client.find_services_iter.return_value = iter((
    {"id": 1111, "lotSlug": "dried-fruit", "lotName": "Raisins & dates"},
    {"id": 2222, "lotSlug": "salad", "lotName": "Lettuce & cucumber"},
    {"id": 3333, "lotSlug": "dried-fruit", "lotName": "Raisins & dates"}
))
```
In the actual test method. Those assertions don't really belong there or add any value and if we're not going to have them then we can avoid having a function or method all together.

benvand:
As a reviewer I'd much prefer to have the unrelated tests and import changes in another branch for a few reasons.
1. The first, completely selfish reason is that the context switching is a pain for the reviewer
2. If we start finding bugs in the new tests or in the import changes then fixing them will hold up the more urgent stuff.
3. It's really hard to define the scope of this sort of thing. For example; in that `test_suppliers.py` there are some lines that are `#noqa` d just because they have long strings. I don't think that's a good enough reason to `#noqa` a whole line. It would be hard to argue that that is out of scope of this PR since we're doing refactoring work in it now.

Got no problems with the actual tests and removing unused imports, that's great, but in another PR would be cool next time and we could tag it as lower priority or something. 👍  Few small changes about consistent spacing and using `called_once_with` instead of `called_with` or `call_args ==`

tests/app/main/views/test_stats.py
katstevens: I'm usually reluctant to use `assert_called_once_with` for 1) fear of a stray typo on a mock method always returning `True` 2) the clarity of having an `assert` at the beginning of the statement.   If we want to retain the ability ensure the mock is only called once we can use `call_args_list == [foo]` (probably overkill here). However I agree that whatever we use it should be consistent, which it clearly isn't below 😸 What do you think?

tests/app/main/views/test_suppliers.py
katstevens: Good call. I admit I didn't look at what this test was actually doing too closely, just moved it around. Will fix.

PR #299, opened 2017-10-17
Add permissions for CCS Category users

TheDoubleK:
Ticket: https://trello.com/c/DkZWVNRe/81-transfer-existing-admin-permissions-to-a-different-admin-category

This gives admin-ccs-category access to the existing functionality for:
 -  Editing all fields for G-Cloud services (NO CODE CHANGE HERE - as already had this)
 - Edit company names for suppliers
 - Publish communications files or questions/answers 

The ticket says "during framework applications" for the final point, but that isn't what admin users currently have - we have a hard-coded list of frameworks that communications can be uploaded for (this was a requirement at some point in the past - to be able to add communications after framework was live) - so I've given CCS sourcing the same ability.


tests/app/main/views/test_suppliers.py
```diff
@@ -516,25 +516,26 @@ def test_should_be_a_503_if_email_fails(self, data_api_client, DMNotifyClient):
...
+            data_api_client.update_supplier.assert_called_with(1234, {'name': "Something New"}, "test@example.com")
+
+    def test_ccs_sourcing_role_can_not_update_supplier_name(self, data_api_client):
+        self.user_role = 'admin-ccs-sourcing'
+        response = self.client.post(
+            '/admin/suppliers/1234/edit/name',
+            data={'new_supplier_name': 'Something New'}
+        )
+        assert response.status_code == 403
+        assert data_api_client.update_supplier.called is False
```
katstevens: I'd prefer `assert_not_called` (well, really I'd prefer `assert foo.call_args_list == []` everywhere but no-one seems keen on that style apart from me!) however we do use this style elsewhere in this module. So up to you if you want to change.

app/templates/index.html
```diff
@@ -148,36 +148,44 @@
...
+    {% set items = [
+        {
+          "body": "Manage G-Cloud 8 communications",
+          "link": url_for(".manage_communications", framework_slug="g-cloud-8"),
+          "title": "G-Cloud 8 communications"
+        },
+        {
           "body": "Manage Digital Outcomes and Specialists 2 communications",
-          "link": "/admin/communications/digital-outcomes-and-specialists-2",
+          "link": url_for(".manage_communications", framework_slug="digital-outcomes-and-specialists-2"),
```
katstevens: Good stuff 👍

tests/app/main/views/test_communications.py
```diff
@@ -11,19 +11,56 @@
...
+            self.s3.return_value.save.call_count == 2
+            flash_messages = self._get_flash_messages()
+            # arguably a misunderstanding here about what an appropriate 'message category' looks like
+            assert set(flash_messages.getlist('upload_communication')) == set(('clarification', 'communication',))
+
+            assert response.status_code == 302
+
+    def test_ccs_sourcing_can_not_post_documents_for_framework(self, data_api_client):
+        # check we are mocking s3 as we must not actually post there!
+        assert isinstance(s3.S3, mock.Mock)
```
katstevens: Out of interest, do we use the `moto` library at all for mocking `boto`/AWS actions? I've found it quite useful in the past as you can just add a `@mock_s3` decorator to your test class, and never worry about accidentally putting something in a real bucket ever again 😸

tests/app/main/views/test_suppliers.py
```diff
@@ -309,7 +309,6 @@ def test_should_show_correct_fields_for_disabled_service(self, data_api_client):
 
         assert response.status_code == 200
         assert "Removed" in response.get_data(as_text=True)
-        assert "Details" in response.get_data(as_text=True)
```
katstevens: Should we be asserting that `Edit` is there instead?

app/templates/view_suppliers.html
```diff
@@ -41,6 +34,7 @@
     {% else %}
```
katstevens: Is it a good idea having the case where more permission-requiring stuff is visible being the default? The answer may well be 'yes' in this case - I don't know enough about the context really.

tests/app/main/views/test_communications.py
TheDoubleK: I know Rob looked at it at some point but can't remember why he decided not to use it.  This is a straight copy-and-paste job from `test_post_documents_for_framework` above TBH so I've given it zero thought here.

app/templates/view_suppliers.html
TheDoubleK: Im not sure what you mean by more and less here?  Basically this page is completely different for admin-ccs-sourcing than for other admin users - sourcing get a lit of frameworks and links to framework agreement approval/download.  Other admin users get the "Change name, Users, Services" links - it's not more- or less-allowed rather than a totally different set of stuff.  They should probably be entirely different views.  I might even do that now because this is horrible.

tests/app/main/views/test_suppliers.py
TheDoubleK: Maybe. But if we're going to do it properly then we should be rendering the document and using XPath lookup in the right place.  The test felt so frail as it was I didn't think losing this would make a lot of difference.

There, I've said it and now feel ashamed.

tests/app/main/views/test_suppliers.py
TheDoubleK: Rob also likes the `call_args_list`.  I'm ambivalent.

This is only green in the diff because of an indentation change where I removed the loop... not really new code.

TheDoubleK: But I can fix it while I'm here now.

app/templates/view_suppliers.html
katstevens: Yep, I think that would be easier to understand (especially for admin n00bs like me). If the templates show two completely different things, I am happy to have two different templates.

tests/app/main/views/test_suppliers.py
katstevens: 😸 I guess revamping the tests to be stricter is one for the tech debt pile.

app/templates/view_suppliers.html
```diff
@@ -23,100 +23,9 @@
...
-      ] %}
+      {% include "_view_suppliers_agreements.html" %}
     {% else %}
-      {% set field_headings = [
-        "Name",
-        summary.hidden_field_heading("Change name"),
-        summary.hidden_field_heading("Users"),
-        summary.hidden_field_heading("Services"),
-      ] %}
+      {% include "_view_suppliers_edit_list.html" %}
```
katstevens: Much easier to follow 👍

tests/app/main/views/test_suppliers.py
TheDoubleK: I'm going to at least put the check for "Edit" in, as the following test for an enabled service has it.

tests/app/status/test_views.py
```diff
@@ -10,7 +10,7 @@ class TestStatus(BaseApplicationTest):
     def test_should_return_200_from_elb_status_check(self, data_api_client):
         status_response = self.client.get('/admin/_status?ignore-dependencies')
         assert status_response.status_code == 200
-        assert data_api_client.called is False
+        assert data_api_client.call_args_list == []
```
katstevens: 😻 Beautiful!

PR #300, opened 2017-10-20
Admin page to add buyer domains

TheDoubleK:
Trello card: https://trello.com/c/TRYLtlWB/100-allow-admin-to-add-buyer-domain-to-the-digital-marketplace-whitelist

A new page, a link to it from the homepage, with some bits and pieces of tidying up admin page templates thrown in for good measure.

## Homepage link
screen shot 2017-10-20 at 16 46 45

## New page
screen shot 2017-10-20 at 16 46 53

## Added a domain
screen shot 2017-10-20 at 16 47 07

## Errors
screen shot 2017-10-20 at 16 47 17

screen shot 2017-10-23 at 11 42 00

### DIFF
--- a/app/main/views/buyers.py
+++ b/app/main/views/buyers.py
@@ -1,5 +1,8 @@
-from flask import render_template, request, flash
+from dmapiclient import HTTPError
+from flask import flash, redirect, render_template, request, url_for
+from flask_login import current_user
 
+from ..forms import EmailDomainForm
 from .. import main
 from ... import data_api_client
 from ..auth import role_required
@@ -29,3 +32,26 @@ def find_buyer_by_brief_id():
         title=title,
         brief_id=brief_id
     )
+
+
+@main.route('/buyers/add-buyer-domains', methods=['GET', 'POST'])
+@role_required('admin')
+def add_buyer_domains():
+    email_domain_form = EmailDomainForm()
+
+    if email_domain_form.validate_on_submit():
+        try:
+            new_domain = request.form.get('new_buyer_domain')
+            data_api_client.create_buyer_email_domain(new_domain, current_user.email_address)
+            flash('You’ve added {}.'.format(new_domain), 'message')
+            return redirect(url_for('.add_buyer_domains'))
+        except HTTPError as e:
+            if "has already been approved" in e.message:
+                flash('You cannot add this domain because it already exists.', 'error')
+            elif "was not a valid format" in e.message:
+                flash('The domain {} is not a valid format'.format(new_domain), 'error')
+
+    return render_template(
+        "add_buyer_email_domain.html",
+        email_domain_form=email_domain_form
+    ), 400 if email_domain_form.new_buyer_domain.errors else 200
--- a/app/main/views/buyers.py
+++ b/app/main/views/buyers.py
@@ -38,7 +38,7 @@ def find_buyer_by_brief_id():
 @role_required('admin')
 def add_buyer_domains():
     email_domain_form = EmailDomainForm()
-
+    status_code = 200
     if email_domain_form.validate_on_submit():
         try:
             new_domain = request.form.get('new_buyer_domain')
@@ -46,12 +46,17 @@ def add_buyer_domains():
             flash('You’ve added {}.'.format(new_domain), 'message')
             return redirect(url_for('.add_buyer_domains'))
         except HTTPError as e:
+            status_code = 400
             if "has already been approved" in e.message:
                 flash('You cannot add this domain because it already exists.', 'error')
             elif "was not a valid format" in e.message:
                 flash('The domain {} is not a valid format'.format(new_domain), 'error')
+            else:
+                raise e
+    elif email_domain_form.new_buyer_domain.errors:
+        status_code = 400
 
     return render_template(
         "add_buyer_email_domain.html",
         email_domain_form=email_domain_form
-    ), 400 if email_domain_form.new_buyer_domain.errors else 200
+    ), status_code
--- a/app/main/views/buyers.py
+++ b/app/main/views/buyers.py
@@ -48,9 +48,13 @@ def add_buyer_domains():
         except HTTPError as e:
             status_code = 400
             if "has already been approved" in e.message:
-                flash('You cannot add this domain because it already exists.', 'error')
+                email_domain_form.new_buyer_domain.errors.append(
+                    'You cannot add this domain because it already exists.'
+                )
             elif "was not a valid format" in e.message:
-                flash('The domain {} is not a valid format'.format(new_domain), 'error')
+                email_domain_form.new_buyer_domain.errors.append(
+                    "‘{}’ is not a valid format.".format(new_domain)
+                )
             else:
                 raise e
     elif email_domain_form.new_buyer_domain.errors:
--- a/app/templates/add_buyer_email_domain.html
+++ b/app/templates/add_buyer_email_domain.html
@@ -1,0 +1,72 @@
+{% extends "_base_page.html" %}
+{% block page_title %}
+  Add buyer email domains – Digital Marketplace admin
+{% endblock %}
+
+{% block breadcrumb %}
+  {%
+    with items = [
+      {
+        "link": url_for('.index'),
+        "label": "Admin home"
+      },
+      {
+        "label": "Add buyer email domains"
+      }
+    ]
+  %}
+    {% include "toolkit/breadcrumb.html" %}
+  {% endwith %}
+{% endblock %}
+
+{% block main_content %}
+  {% with messages = get_flashed_messages(with_categories=true) %}
+  {% if messages %}
+      {% for category, message in messages %}
+          {%
+          with
+              message = message,
+              type = "destructive" if category == 'error' else 'success'
+          %}
+          {% include "toolkit/notification-banner.html" %}
+          {% endwith %}
+      {% endfor %}
+  {% endif %}
+  {% endwith %}
+
+  {% with heading = "Add buyer email domains" %}
+    {% include "toolkit/page-heading.html" %}
+  {% endwith %}
+
+  <form method="post">
+    <div class="grid-row">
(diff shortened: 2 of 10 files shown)


TheDoubleK:
Laurence and Constance have seen this running locally and are happy with it.

tests/app/main/views/test_buyers.py
```diff
@@ -61,3 +64,83 @@ def test_should_show_buyers_contact_details(self, data_api_client):
...
+        assert actual_code == expected_code, "Unexpected response {} for role {}".format(response.status_code, role)
+
+    def test_admin_user_can_add_a_new_buyer_domain(self, data_api_client):
+        response = self.client.post('/admin/buyers/add-buyer-domains',
+                                    data={'new_buyer_domain': 'kev.uk'},
+                                    follow_redirects=True
+                                    )
+        assert response.status_code == 200
+        assert data_api_client.create_buyer_email_domain.call_args_list == [mock.call("kev.uk", "test@example.com")]
+        assert "You’ve added kev.uk" in response.get_data(as_text=True)
```
katstevens: In the Briefs FE we use `assert_flashes()` (a helper method on `BaseApplicationTest`) to check the right message key and category has been sent.  It's not on the Admin FE `BaseApplicationTest` but we could easily copy it over? Or ... put it in the poor neglected `test_utils` repo (https://github.com/alphagov/digitalmarketplace-test-utils) for all apps to use?

tests/app/main/views/test_buyers.py
```diff
@@ -61,3 +64,83 @@ def test_should_show_buyers_contact_details(self, data_api_client):
...
+
+    def test_post_empty_form_error(self, data_api_client):
+        response = self.client.post('/admin/buyers/add-buyer-domains',
+                                    data={'new_buyer_domain': ''}
+                                    )
+        assert response.status_code == 400
+        assert "Domain can not be empty" in response.get_data(as_text=True)
+        assert data_api_client.create_buyer_email_domain.call_args_list == []
```
katstevens: 😻 for `call_args_list == []`

tests/app/main/views/test_buyers.py
```diff
@@ -61,3 +64,84 @@ def test_should_show_buyers_contact_details(self, data_api_client):
...
+
+    def test_post_bad_domain_error(self, data_api_client):
+        mock_api_error = mock.Mock(status_code=400)
+        mock_api_error.json.return_value = {"error": "JSON was not a valid format: 'inv@lid.co' does not match..."}
+        data_api_client.create_buyer_email_domain.side_effect = HTTPError(mock_api_error)
+        response = self.client.post('/admin/buyers/add-buyer-domains',
+                                    data={'new_buyer_domain': 'inv@lid.co'}
+                                    )
+        assert response.status_code == 400
+        print("PAGE: {}".format(response.get_data(as_text=True)))
```
katstevens: Leftover `print()`

app/templates/index.html
```diff
@@ -89,6 +89,19 @@
 <div class="grid-row">
     <div class="column-two-thirds">
 
+  {% if current_user.has_role('admin') %}
```
katstevens: Do we have any tests to assert what each user role can view on the Admin home page?

app/templates/view_supplier_users.html
```diff
@@ -156,24 +151,17 @@
...
-                      {% if invite_form.email_address.errors %}
-                  </div>
-                  {% endif %}
+                  {%
+                    with
+                    question = invite_form.email_address.label,
+                    hint = "Enter the email address of the person you wish to invite",
+                    name = invite_form.email_address.name,
+                    value = invite_form.email_address.data,
+                    error = invite_form.email_address.errors[0]
```
katstevens: This might be a daft question but will `.errors[0]` break if there are no errors here? Or does the toolkit template deal with it for us?

app/templates/view_supplier_users.html
```diff
@@ -185,24 +173,17 @@
...
-                        {% if move_user_form.user_to_move_email_address.errors %}
-                    </div>
-                    {% endif %}
+                    {%
+                      with
+                      question = move_user_form.user_to_move_email_address.label,
+                      hint = "Enter the email address of the existing user you wish to move to this supplier",
+                      name = move_user_form.user_to_move_email_address.name,
+                      value = move_user_form.user_to_move_email_address.data,
+                      error = move_user_form.user_to_move_email_address.errors[0]
```
katstevens: Same here - what happens to `errors[0]` if the list is empty?

app/main/views/buyers.py
```diff
@@ -29,3 +32,35 @@ def find_buyer_by_brief_id():
...
+                email_domain_form.new_buyer_domain.errors.append(
+                    'You cannot add this domain because it already exists.'
+                )
+            elif "was not a valid format" in e.message:
+                email_domain_form.new_buyer_domain.errors.append(
+                    "‘{}’ is not a valid format.".format(new_domain)
+                )
+            else:
+                raise e
+    elif email_domain_form.new_buyer_domain.errors:
```
katstevens: Why do we need to check if there are `email_domain_form.new_buyer_domain.errors` here? If `email_domain_form.validate_on_submit()` fails, don't we want to return a 400 status whether or not there's an error on that particular form field (i.e. just do an `else:` on this line)?

app/templates/add_buyer_email_domain.html
```diff
@@ -0,0 +1,82 @@
...
+          </a> before you add a domain.
+        </p>
+
+        {%
+          with
+          question = "",
+          hint = "For example police.uk",
+          name = email_domain_form.new_buyer_domain.name,
+          value = email_domain_form.new_buyer_domain.data,
+          error = ""
```
katstevens: I think `error` is optional for the textbox template so you might be able to leave it out (I think `question` is required though).

app/main/views/buyers.py
TheDoubleK: It's for the `GET` route - we want to return a 200 for the empty form page on a `GET`.

app/templates/index.html
TheDoubleK: Ooh probably yes - I should add one for this.  I'd forgotten about the homepage link and only added it at the end of Friday after doing all the other tests.

app/templates/add_buyer_email_domain.html
TheDoubleK: Yes, I was leaving it in to explicitly make it clear that we're not passing the error to the form though (because I think it's weird that we're not, and that the banners should actually be form field validation errors instead, but that's an argument I lost).

Can drop it if you think it's fine implicitly not passing the error through though.

app/templates/view_supplier_users.html
TheDoubleK: It's definitely fine - tried and tested - Jinja just carries on if it gets a None for things like this.

app/templates/view_supplier_users.html
TheDoubleK: It works OK with no error message shown.

tests/app/main/views/test_buyers.py
TheDoubleK: I looked at using this, and yes probably should for this test.  It was breaking where I was doing the weird thing of using flashes for the error messages to render in the template (rather than for a redirect).  I'll patch it in here.

tests/app/main/views/test_buyers.py
TheDoubleK: I learnt it all from you x

tests/app/main/views/test_buyers.py
TheDoubleK: Sorry :facepalm:

app/templates/index.html
TheDoubleK: Hmm, turns out we don't have a lot around this. We do have functional tests that check this stuff, but feels like it should be unit tested.  I'll add some for my new bits at least.

app/templates/view_supplier_users.html
```diff
@@ -23,37 +23,32 @@
...
-      User not moved to this supplier - please check you entered the address of an existing supplier user
-  {% endif %}
-        </p>
-    </div>
-  {% endfor %}
+      {% for category, message in messages %}
+          {%
+          with
+              message = flash_messages_content[message],
+              type = "destructive" if category == 'error' else 'success'
```
CrystalPea: Do we use destructive-type messages when there is an error? Shouldn't there be a separate error-type message, since we are not really deleting anything?

app/templates/view_supplier_users.html
katstevens: Phew - I do get overly paranoid about `[0]` sometimes...

app/templates/add_buyer_email_domain.html
katstevens: That makes sense - let's leave it as it is.

app/main/views/buyers.py
katstevens: Ahaaa of course, this view has both methods doesn't it? 😸  I'd probably still go for the more generic `elif email_domain_form.errors`, but this form is unlikely to get more complicated in the future so it's probably fine as is.

app/templates/view_supplier_users.html
TheDoubleK: Yes it's an established pattern to use these banners for error messages (e.g. validation errors submitting declaration or services during applications).  This is definitely a legit use of the macro.

CrystalPea: Ok, thanks a lot for clarifying!

app/templates/view_supplier_users.html
TheDoubleK: Me too - hence I know from recent experience it's fine.
I was looking for a neater way to do it but couldn't come up with anything 😞  - open to suggestions if anything springs to mind, but I think this is fine if not.

TheDoubleK:
@katstevens discussed at stand up and we think this is OK to merge and QA on Preview now.

PR #301, opened 2017-10-27
Add acl to communications, deextrapolate get_path

benvand:
https://trello.com/c/pZNuiwDv/67-communication-files-in-s3-files-are-public

2nd line fix ticket to save communications and clarifications with S3 private ACL
Remove overly extrapolated code
Add tests

### DIFF
--- a/app/main/views/communications.py
+++ b/app/main/views/communications.py
@@ -9,17 +9,6 @@ from dmutils import s3  # this style of import so we only have to mock once
 from dmutils.documents import file_is_pdf, file_is_csv, file_is_open_document_format
 
 
-def _get_path(framework_slug, path):
-    return '{}/communications/{}'.format(framework_slug, path)
-
-
-def _save_file(bucket, framework_slug, path, the_file, flash_message):
-    path = "{}/{}".format(_get_path(framework_slug, path), the_file.filename)
-    # setting download_filename ensures Content-Disposition: attachment
-    bucket.save(path, the_file, download_filename=the_file.filename)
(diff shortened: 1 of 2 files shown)


app/main/views/communications.py
```diff
@@ -53,15 +42,19 @@ def upload_communication(framework_slug):
...
 
     if request.files.get('clarification'):
         the_file = request.files['clarification']
         if not file_is_pdf(the_file):
             errors['clarification'] = 'not_pdf'
 
         if 'clarification' not in errors.keys():
-            _save_file(communications_bucket, framework_slug, 'updates/clarifications', the_file, 'clarification')
+            path = "{}/communications/updates/clarifications/{}".format(framework_slug, the_file.filename)
+            communications_bucket.save(path, the_file, download_filename=the_file.filename)
```
TheDoubleK: I think both should have private ACL maybe? I don't see any reason why we'd have one public and one private.

benvand: Cool, just looked at the ticket and `agreements` is implied too. Will update when the tests are merged https://github.com/alphagov/digitalmarketplace-admin-frontend/pull/302

TheDoubleK: I meant both communications and clarifications. Agreements are already private (see https://github.com/alphagov/digitalmarketplace-admin-frontend/blob/master/app/main/views/suppliers.py#L277)

benvand: Derp, I also this

benvand:
@idavidmcdonald @TheDoubleK 
These are updated now

PR #302, opened 2017-10-27
Tests for communications and clarifications files upload

benvand:


TheDoubleK:
These tests look fine, but would it be worth adding in the ACL fix and tests that ACL is set properly here? It should only be a small piece of work I think.

TheDoubleK:
Sorry! Just seen your other PR. Ignore me.

tests/app/main/views/test_communications.py
```diff
@@ -59,35 +57,62 @@ def test_ccs_sourcing_can_not_post_documents_for_framework(self, data_api_client
...
+        response = self.client.post(
+            "/admin/communications/{}".format(self.framework_slug),
+            data={
+                'communication': (self.dummy_file, 'test-comm.pdf'),
+            }
+        )
+
+        assert response.status_code == 302
+        self.s3.return_value.save.assert_called_once_with(
+            'g-cloud-8/communications/updates/communications/test-comm.pdf',
```
idavidmcdonald: Should the framework slugs in here be taken from `self.framework_slug` rather than hardcoded?

tests/app/main/views/test_communications.py
```diff
@@ -59,35 +57,62 @@ def test_ccs_sourcing_can_not_post_documents_for_framework(self, data_api_client
...
+        response = self.client.post(
+            "/admin/communications/{}".format(self.framework_slug),
+            data={
+                'clarification': (self.dummy_file, 'test-comm.pdf'),
+            }
+        )
+
+        assert response.status_code == 302
+        self.s3.return_value.save.assert_called_once_with(
+            'g-cloud-8/communications/updates/clarifications/test-comm.pdf',
```
idavidmcdonald: Same here on the framework slug hardcoding

tests/app/main/views/test_communications.py
```diff
@@ -59,35 +57,62 @@ def test_ccs_sourcing_can_not_post_documents_for_framework(self, data_api_client
...
+            "/admin/communications/{}".format(self.framework_slug),
+            data={
+                'communication': (self.dummy_file, 'test-comm.pdf'),
+            }
+        )
+
+        assert response.status_code == 302
+        self.s3.return_value.save.assert_called_once_with(
+            'g-cloud-8/communications/updates/communications/test-comm.pdf',
+            mock.ANY,
```
idavidmcdonald: An inline comment here saying what this would make this really readable
i.e.

```mock.ANY,  # this is...```

File: tests/app/main/views/test_communications.py [hunk 4]
```diff
@@ -59,35 +57,62 @@ def test_ccs_sourcing_can_not_post_documents_for_framework(self, data_api_client
...
+            download_filename='test-comm.pdf'
+        )
+
+    def test_clarification_file_saves_with_correct_path(self, data_api_client):
+        assert isinstance(s3.S3, mock.Mock)
```
idavidmcdonald: Is it `s3.S3` that we want to check is the mock or is it `self.s3`. Maybe I'm just a little confused - might be one to explain to me in person

idavidmcdonald: Ah I see how it's being done in the base application test by `dmutils.s3` is being patched.

idavidmcdonald: Small thought that I'm not 100% on, but I wonder if it's worth moving the instance check either into the setup for every test or the set up for the class?

tests/app/main/views/test_communications.py
benvand: We could hard code the string in the url to make it more obvious that it doesn't check the validity of the `framework_slug`. 
Either way, changed to your suggestion for now, let me know what you think

tests/app/main/views/test_communications.py
benvand: Done

tests/app/main/views/test_communications.py
benvand: Actually I'm going to rip out those checks. I think they'd be useful while writing and deploying tests if a dev was extra nervous but like you say, S3 _is_ mocked on every test anyway. It would be like creating an object in the setup method then exerting its existence in every test.

benvand:
@TheDoubleK 
I did that thing that people always say is a good idea but never actually do! 
I wanted to change some code but I couldn't confidently do it because it didn't have test coverage. So I wrote tests for the _existing_ functionality. 
The other pull request changes the code. So when these tests are merged the other PR will go red. Then I'll update these tests in that PR to make them green (by adding the acl).

TheDoubleK:
mind-blown

TheDoubleK:
Maybe the fact that this PR came after the other one put me off the scent :trollface:

benvand:
Haha caught me 😂

PR #303, opened 2017-11-01
Bugfix validation messages for admin service editing

TheDoubleK:
The toolkit version of the validation masthead understands how to show errors for multiquestions, so I swapped the template out to use that.
And the bit that shows individual questions also needed to pass errors through to multiquestions to get the bars next to the individual fields.

The xpath stuff in the new test could probably be neater, but it has taken me hours (considerably longer than the fix itself) to get this test working properly, so only suggest changes if you have the **exact code** that would be better - no vague Xpath hints please :p

Also update API client because there were `print` statements in error handling in the currently imported version 😕 

# BEFORE
screen shot 2017-11-01 at 18 44 29

# AFTER
screen shot 2017-11-01 at 18 42 25

### DIFF
--- a/app/templates/edit_section.html
+++ b/app/templates/edit_section.html
@@ -31,28 +31,19 @@
   %}
     {% include "toolkit/page-heading.html" %}
   {% endwith %}
+
   {% if errors %}
-    <div class="validation-masthead">
-      <h2 class="validation-masthead-heading">
-        There were problems with your response to the following questions:
-      </h2>
-      <ul>
-        {% for question in section.questions %}
-          {% if question.id in errors %}
-            <li>
-              <a class="validation-masthead-link" href="#{{ question.id }}">{{ question.question }}</a>
-            </li>
-          {% endif %}
-        {% endfor %}
-      </ul>
-    </div>
+    {% with errors = errors.values() %}
+      {% include 'toolkit/forms/validation.html' %}
+    {% endwith %}
   {% endif %}
+
   <form method="post" enctype="multipart/form-data">
     <div class="grid-row">
       <div class="column-two-thirds">
         <div style="display:none;"><input id="csrf_token" name="csrf_token" type="hidden" value="{{ csrf_token() }}"></div>
         {% for question in section.questions %}
-          {% if errors and errors[question.id] %}
+          {% if errors and errors[question.id] or question.type == 'multiquestion' %}
             {{ forms[question.type](question, service_data, errors) }}
           {% else %}
             {{ forms[question.type](question, service_data, {}) }}
--- a/tests/app/main/views/test_services.py
+++ b/tests/app/main/views/test_services.py
@@ -574,6 +574,56 @@ class TestServiceUpdate(LoggedInApplicationTest):
         assert response.status_code == 302
 
     @mock.patch('app.main.views.services.data_api_client')
+    def test_service_update_with_multiquestion_validation_error(self, data_api_client):
+        data_api_client.get_service.return_value = {'services': {
+            'id': 1,
+            'supplierId': 2,
+            'frameworkSlug': 'g-cloud-9',
+            'lot': 'cloud-hosting',
+            'serviceFeatures': [
+                "bar",
+            ],
+            'serviceBenefits': [
+                "foo",
+            ],
+        }}
+        mock_api_error = mock.Mock(status_code=400)
+        mock_api_error.json.return_value = {
+            "error": {"serviceBenefits": "under_10_words", "serviceFeatures": "under_10_words"}
+        }
+        data_api_client.update_service.side_effect = HTTPError(mock_api_error)
+
+        response = self.client.post(
+            '/admin/services/1/edit/service-features-and-benefits',
+            data={
+                'serviceFeatures': 'one 2 three 4 five 6 seven 8 nine 10 eleven',
+                'serviceBenefits': '11 10 9 8 7 6 5 4 3 2 1',
+            },
+            follow_redirects=True
+        )
+        assert data_api_client.update_service.call_args_list == [(
+            (
+                '1',
+                {
+                    'serviceFeatures': ['one 2 three 4 five 6 seven 8 nine 10 eleven'],
+                    'serviceBenefits': ['11 10 9 8 7 6 5 4 3 2 1'],
+                },
+                'test@example.com',
+            ),
+            {},
+        )]
+
+        document = html.fromstring(response.get_data(as_text=True))
+
+        validation_banner = document.xpath("//div[@class='validation-masthead']//text()[normalize-space(.) != '']")
+        validation_banner_text = ' '.join([elem.strip() for elem in validation_banner])
+        validation_errors = [error.strip() for error in document.xpath("//span[@class='validation-message']//text()")]
+        assert response.status_code == 400
+        assert validation_banner_text == "There was a problem with your answer to: Service benefits Service features"
+        assert "You can’t write more than 10 words for each feature." in validation_errors
+        assert "You can’t write more than 10 words for each benefit." in validation_errors
+
+    @mock.patch('app.main.views.services.data_api_client')
     def test_service_update_with_assurance_questions(self, data_api_client):
         data_api_client.get_service.return_value = {'services': {
             'id': 567,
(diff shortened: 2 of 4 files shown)


app/templates/edit_section.html
```diff
@@ -31,28 +31,19 @@
   %}
     {% include "toolkit/page-heading.html" %}
   {% endwith %}
+
   {% if errors %}
-    <div class="validation-masthead">
-      <h2 class="validation-masthead-heading">
-        There were problems with your response to the following questions:
```
katstevens: Do we care about pluralisation here? The toolkit template uses the singular by default, so if we answer two questions wrong on the same page, the message is 

```
There was a problem with your answer to: 
Thing 1 
Thing 2
```

If we prefer the toolkit's singular version (I think it's probably better than pluralisation by default!), do we have any other pluralised error messages lurking around that need changing?

app/templates/edit_section.html
```diff
@@ -31,28 +31,19 @@
   %}
     {% include "toolkit/page-heading.html" %}
   {% endwith %}
+
   {% if errors %}
-    <div class="validation-masthead">
-      <h2 class="validation-masthead-heading">
```
katstevens: The toolkit template replaces this `<h2>` heading with an `<h1>`. We end up with two h1s on the page. This seems... bad? And also... happening in a lot of places?

tests/app/main/views/test_services.py
```diff
@@ -574,6 +574,56 @@ def test_service_update_with_one_service_feature(self, data_api_client):
...
+
+        document = html.fromstring(response.get_data(as_text=True))
+
+        validation_banner = document.xpath("//div[@class='validation-masthead']//text()[normalize-space(.) != '']")
+        validation_banner_text = ' '.join([elem.strip() for elem in validation_banner])
+        validation_errors = [error.strip() for error in document.xpath("//span[@class='validation-message']//text()")]
+        assert response.status_code == 400
+        assert validation_banner_text == "There was a problem with your answer to: Service benefits Service features"
+        assert "You can’t write more than 10 words for each feature." in validation_errors
+        assert "You can’t write more than 10 words for each benefit." in validation_errors
```
katstevens: I think we can be more exact in asserting both the masthead (with its bookmark links) and the errors. Something like:

```
validation_banner_h1 = document.xpath("//h1[@class='validation-masthead-heading']//text()")[0].strip()
assert validation_banner_h1 == "There was a problem with your answer to:"

validation_banner_links = [
    (anchor.text_content(), anchor.get('href')) for anchor in document.xpath("//a[@class='validation-masthead-link']")
]
assert validation_banner_links == [
    ("Service benefits", "#serviceBenefits"),
    ("Service features", "#serviceFeatures")
]

validation_errors = [error.strip() for error in document.xpath("//span[@class='validation-message']//text()")]
assert validation_errors == [
    "You can’t write more than 10 words for each feature.",
    "You can’t write more than 10 words for each benefit."
]
```

Going to the effort of doing this revealed to me that a) there are two h1 elements in the page, b) the list of bookmark links in the masthead is in a different order to the questions! I think the latter's a known bug, but I had forgotten about it until just now...

app/templates/edit_section.html
TheDoubleK: Ooh, good spot.  Not sure of the best way out here...

app/templates/edit_section.html
TheDoubleK: I think the toolkit version is better (it's certainly newer).  As to whether there are other messages about... quite possibly.  I can have a look in this app for now.

tests/app/main/views/test_services.py
TheDoubleK: That rings a bell.  It might be a problem that was fixed in supplier app, but I'm not sure, will check it out.

app/templates/edit_section.html
TheDoubleK: **TL;DR** I think this is legit as it is.

Actually, I remember there was a long and involved discussion about this with Tom Byers when he was here.  We ended up agreeing that two H1's on the page are the best thing in this case (yep, honestly...).  I *think* the reasoning went something like this: for accessibility purposes it is generally good practice to have only one H1 on a page.  But in the case where there's an error on the page that you need to fix before you can continue then that is the most important thing you need to know at that point in time, and so is deserving of an H1.

I think this template pattern was only built for G-Cloud 9 submissions, and @samuelhwilliams and @galund might remember the discussion? (Or I might have just dreamed it up).

tests/app/main/views/test_services.py
TheDoubleK: Many thanks for not making vague suggestions BTW 😸

app/templates/edit_section.html
TheDoubleK: There are no obvious others anywhere in this repo.

app/templates/edit_section.html
samuelhwilliams: Not sure I recognise this template, so might have been George.

app/templates/edit_section.html
samuelhwilliams: Would be nice if the toolkit version pluralised itself if appropriate.

app/templates/edit_section.html
TheDoubleK: Looking at the commit history, the swap to H1 was in line with guidance from GOV.UK elements: https://github.com/alphagov/digitalmarketplace-frontend-toolkit/commit/07f13f8fb6a963f7e87357f71b94e990f40d944c#diff-4388d17aabba821f10d7ca1a64c1e797

app/templates/edit_section.html
katstevens: As long as it's consistent I think the toolkit version is fine for now (perhaps pluralisation is a nice 20% project for someone?).

app/templates/edit_section.html
TheDoubleK: Although the GOV.UK elements page seems to have downgraded it to an H2 again these days.  I don't think fixing this is part of this bugfix though.

app/templates/edit_section.html
TheDoubleK: I think the singular version works just fine with a list of things under it?

app/templates/edit_section.html
katstevens: Drawing attention to the validation makes sense I suppose. I guess as long as it doesn't confuse the styling or any JS then it's not a blocker (I certainly don't want to make an old discussion any longer / more involved than it has been!)

TheDoubleK: We try to keep in step with GOV.UK thinking on most things like this I think, but it's a moving target and without anyone dedicated to the frontend it's hard for us to keep up. (Changes are the sort of thing one might pick up in the weekly frontend meeting if we had anyone who attended it.)

There probably is a piece of work to be done in bringing our templates into line with most recent GOV.UK styles. (Which would drop this back to an H2), but I think right now by me isn't the time.

tests/app/main/views/test_services.py
```diff
@@ -615,13 +615,23 @@ def test_service_update_with_multiquestion_validation_error(self, data_api_clien
...
+            (anchor.text_content(), anchor.get('href')) for anchor in
+            document.xpath("//a[@class='validation-masthead-link']")
+        ]
+        assert validation_banner_links == [
+            ("Service benefits", "#serviceBenefits"),
+            ("Service features", "#serviceFeatures")
+        ]
+
         validation_errors = [error.strip() for error in document.xpath("//span[@class='validation-message']//text()")]
-        assert response.status_code == 400
```
katstevens: Ooops - did I delete the 400 status assertion?

TheDoubleK: All my fault I'm sure.  Will put it back.

app/templates/edit_section.html
katstevens: Fair enough!

tests/app/main/views/test_services.py
TheDoubleK: I looked at the ordering thing but supplier app doesn't deal with it that I can see.  I think we leave it as-is for now.

tests/app/main/views/test_services.py
TheDoubleK: It's back!

katstevens: \o/

PR #305, opened 2017-11-07
Nest supplier agreements table to apply custom scss only there

benvand:
Looks like the main scss `summary-item-field-*` classes were overridden in order to style a single table.
I've nested this table and applied the scss styles only to that nestings child elements. This seems like a better way for now and won't affect every single summary table in the admin.

Before:

screen shot 2017-11-07 at 16 24 45

After:

screen shot 2017-11-07 at 16 25 35

### DIFF
--- a/app/assets/scss/application.scss
+++ b/app/assets/scss/application.scss
@@ -95,15 +95,12 @@ $path: "/admin/static/images/";
   margin-bottom: 40px;
 }
 
-.summary-item-field-first {
-  width: 15%;
-}
-
-.summary-item-field {
-  width: 15%;
+.summary-item-field-heading, .summary-item-field {
   padding-left: 10px;
 }
 
-.summary-item-field-heading {
-  padding-left: 10px;
+.six-columns-fixed-width-table {
+  .summary-item-field-first, .summary-item-field {
+    width: 16.666%;
+  }
 }
--- a/app/templates/_view_suppliers_agreements.html
+++ b/app/templates/_view_suppliers_agreements.html
@@ -1,3 +1,4 @@
+<div class="six-columns-fixed-width-table">
 {% set field_headings = [
   "Name",
   "G-Cloud 7",
@@ -75,3 +76,4 @@
     {% endcall %}
   {% endcall %}
 {% endcall %}
+</div>


app/assets/scss/application.scss
```diff
@@ -95,15 +95,17 @@ $path: "/admin/static/images/";
...
+    padding-left: 10px;
+  }
 
-.summary-item-field-heading {
-  padding-left: 10px;
-}
+  .summary-item-field-heading {
+    padding-left: 10px;
+  }
+}
```
TheDoubleK: Missing line terminator here.

app/assets/scss/application.scss
```diff
@@ -95,15 +95,17 @@ $path: "/admin/static/images/";
...
+    width: 15%;
+  }
 
-.summary-item-field {
-  width: 15%;
-  padding-left: 10px;
-}
+  .summary-item-field {
+    width: 15%;
+    padding-left: 10px;
```
TheDoubleK: I think we probably want to keep the padding-left for all summary table fields, not just the `fixed-column-width-table`.  I reckon the only thing inside the `fixed-column-width-table` block should be the 15% width rules.

app/assets/scss/application.scss
```diff
@@ -95,15 +95,12 @@ $path: "/admin/static/images/";
   margin-bottom: 40px;
 }
 
-.summary-item-field-first {
-  width: 15%;
-}
-
-.summary-item-field {
-  width: 15%;
-  padding-left: 10px;
```
TheDoubleK: This loses the `padding-left` altogether - is that what we want?

benvand:
Interesting thought.
So it looks like it gives the `summary-item-field` elements a bit more room. _However_, the `summary-item-field-first` is playing by its own rules because it has its own `summary-row` and that's what the 16.666 is a percentage of. 
The parent `summary-row`of the `summary-item-field-first` is inexplicably wider than the parent `summary-row`of the 5 `summary-item-field`s

benvand:
I changed it though because I think the names and numbers better reflect what we're trying to do.

PR #306, opened 2017-11-07
Mark service updates as by_admin

TheDoubleK:
Tests will fail until API client https://github.com/alphagov/digitalmarketplace-apiclient/pull/110 is merged.

Following on from:
 - [ ] https://github.com/alphagov/digitalmarketplace-apiclient/pull/110
 - [ ] https://github.com/alphagov/digitalmarketplace-api/pull/687

This is the final part for ticket: https://trello.com/c/DbYlqSxv/121-exclude-changes-made-by-admin-in-the-approve-edits-section

Updates made by admin users pass the `by_admin=True` flag in so the API knows to use the `update_service_admin` audit event for the update.

### DIFF
--- a/tests/app/main/views/test_services.py
+++ b/tests/app/main/views/test_services.py
@@ -427,7 +427,7 @@ class TestServiceUpdate(LoggedInApplicationTest):
         data_api_client.update_service.assert_called_with('1', {
             'pricingDocumentURL': 'https://assets.test.digitalmarketplace.service.gov.uk/g-cloud-7/documents/2/1-pricing-document-2015-01-01-1200.pdf',  # noqa
             'sfiaRateDocumentURL': 'https://assets.test.digitalmarketplace.service.gov.uk/g-cloud-7/documents/2/1-sfia-rate-card-2015-01-01-1200.pdf',  # noqa
-        }, 'test@example.com')
+        }, 'test@example.com', by_admin=True)
 
         assert response.status_code == 302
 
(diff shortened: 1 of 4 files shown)


tests/app/main/views/test_services.py
```diff
@@ -603,17 +596,14 @@ def test_service_update_with_multiquestion_validation_error(self, data_api_clien
         )
 
         assert response.status_code == 400
-        assert data_api_client.update_service.call_args_list == [(
-            (
+        print("EXPECTED: {}".format(data_api_client.update_service.call_args_list))
```
katstevens: Leftover `print` here...

tests/app/main/views/test_services.py
```diff
@@ -560,17 +555,15 @@ def test_service_update_with_one_service_feature(self, data_api_client):
                 'serviceBenefits': 'foo',
             }
         )
-        assert data_api_client.update_service.call_args_list == [(
-            (
+        assert data_api_client.update_service.call_args_list == [
+            mock.call(
```
katstevens: Definitely more readable with `mock.call`!

tests/app/main/views/test_services.py
TheDoubleK: Sorry - fixed.

tests/app/main/views/test_services.py
TheDoubleK: ...and also I couldn't for the life of me figure out how to assert the new named parameter the other way 😛

PR #307, opened 2017-11-08
120 allow admins to view and edit dos services 

benvand:
New manifest:
https://github.com/alphagov/digitalmarketplace-frameworks/pull/479

https://trello.com/c/U2Yd4CGD/120-allow-admin-view-and-edit-dos-services-update-prices-for-digital-specialists

As an admin on a supplier services page, when I click on edit, I should be able to edit all the details of the service except service essentials.

* Pull in new manifest
* Allow for nested dos multiquestion as per supplier edit services (https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/master/app/main/views/services.py#L482-L482)
* Remove link to service for DOS as this is not public
* Use framework + lot in template if service has no name
* 'add' link if multi question has no existing data otherwise 'edit'
* Update/ add tests

screen shot 2017-11-08 at 15 51 44

screen shot 2017-11-08 at 15 53 39

### DIFF
--- a/app/templates/view_supplier_services.html
+++ b/app/templates/view_supplier_services.html
@@ -49,7 +49,13 @@
           field_headings_visible=True
         ) %}
           {% call summary.row() %}
-            {{ summary.service_link(item.serviceName, '/' + item.frameworkFramework + '/services/' + item.id|string) }}
+            {% if item.serviceName %}
+              {{ summary.service_link(item.serviceName, '/' + item.frameworkFramework + '/services/' + item.id|string) }}
+            {% else %}
+              {% call summary.field(first=True, wide=True) %}
+                {{ item.lotName }}
+              {% endcall %}
+            {% endif %}
             {{ summary.text(item.id) }}
             {{ summary.text(item.lotName) }}
             {% call summary.field() %}
--- a/tests/app/main/views/test_services.py
+++ b/tests/app/main/views/test_services.py
@@ -36,6 +36,7 @@ class TestServiceView(LoggedInApplicationTest):
     def test_service_view_no_features_or_benefits_status_disabled(self, data_api_client):
         data_api_client.get_service.return_value = {'services': {
             'frameworkSlug': 'g-cloud-8',
+            'serviceName': 'test',
             'lot': 'iaas',
             'id': "314159265",
             "status": "disabled",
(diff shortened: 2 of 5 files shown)


TheDoubleK:
The alignment is out in the list of services for a supplier:
screen shot 2017-11-10 at 11 48 47

app/templates/view_supplier_services.html
```diff
@@ -40,7 +40,11 @@
     field_headings_visible=True
   ) %}
     {% call summary.row() %}
-      {{ summary.service_link(item.serviceName, '/' + item.frameworkFramework + '/services/' + item.id|string) }}
+      {% if item.serviceName %}
+        {{ summary.service_link(item.serviceName, '/' + item.frameworkFramework + '/services/' + item.id|string) }}
+      {% else %}
+        {{ summary.text(item.frameworkName + ' - ' + item.lotName) }}
```
TheDoubleK: The framework name appears in the third column of the table, so maybe we don't need both here?  "Digital Outcomes and Specialists 2 - User research participants" is pretty long and both pieces of information are repeated later in the row.  Probably a question for @laurenceberry I guess.

(Also, this list page is going to be split out into per-framework tables in a story we just wrote today, so including the framework name here will be even less useful then.)

tests/app/main/views/test_services.py
```diff
@@ -253,7 +260,7 @@ def test_service_status_update_widgets_not_visible_when_not_permitted(self, data
 
 class TestServiceEdit(LoggedInApplicationTest):
     @mock.patch('app.main.views.services.data_api_client')
-    def test_edit_dos_service_title(self, data_api_client):
+    def test_edit_dos1_service_title(self, data_api_client):
```
TheDoubleK: `dos1` isn't a thing - only sequels get numbered!! 😛

app/templates/view_supplier_services.html
benvand: Done

tests/app/main/views/test_services.py
benvand: And done

TheDoubleK:
Did you fix the alignment in that table? We should do that before getting this in.

benvand:
@TheDoubleK To fix the alignment I've used a `summary.field_name` on the first cell in the table. This creates a `td class="summary-item-field-first"` element instead of a standard `summary-item-field` and solves the alignment problem.

PR #309, opened 2017-11-16
Admin can see details when more than one user edits an account

CrystalPea:
Also refactor code testing this content into a separate test.

Trello ticket: https://trello.com/c/ULWlebbv/137-admin-can-see-details-when-more-than-one-user-edits-an-account

### DIFF
--- a/app/main/views/services.py
+++ b/app/main/views/services.py
@@ -217,7 +217,9 @@ def service_updates(service_id):
         all_update_events=all_update_events,
         latest_update_events=latest_update_events,
         oldest_update_events=oldest_update_events,
-        number_of_users_who_made_edits=len(frozenset(audit_event["user"] for audit_event in chain(
+        # below number of users who made edits might not be reliable if over 2 default length pages
+        # (100 edits per page at the time of writing) of unapproved edits per service were made
+        min_number_of_users_who_made_edits=len(frozenset(audit_event["user"] for audit_event in chain(
             latest_update_events,
             oldest_update_events,
         ))),
--- a/app/main/views/services.py
+++ b/app/main/views/services.py
@@ -183,7 +183,6 @@ def service_updates(service_id):
     #   case it is set to None
     # - latest_update_events[0] and oldest_update_events[-1] should contain the latest and oldest update events
     #   respectively
-
     extra_context = {}
     if latest_update_events:
         archived_service_response = data_api_client.get_archived_service(
@@ -218,7 +217,7 @@ def service_updates(service_id):
         all_update_events=all_update_events,
         latest_update_events=latest_update_events,
         oldest_update_events=oldest_update_events,
-        n_editing_users_min=len(frozenset(audit_event["user"] for audit_event in chain(
+        number_of_users_who_made_edits=len(frozenset(audit_event["user"] for audit_event in chain(
             latest_update_events,
             oldest_update_events,
         ))),
--- a/app/templates/compare_revisions.html
+++ b/app/templates/compare_revisions.html
@@ -63,27 +63,19 @@
     <div class="grid-row">
       <div class="column-two-thirds page-section">
         <p>
-          {% if all_update_events is none %}
-            Multiple
+          {% if latest_update_events %}
+            {% if number_of_users_who_made_edits > 1 %}
+              The last user to edit this page was
+            {% endif %}
+              {{ latest_update_events.0.user }}
+            {% if number_of_users_who_made_edits == 1 %}
+              made {{ (all_update_events or (oldest_update_events+latest_update_events))|length }}
+              {{ pluralize((all_update_events or (oldest_update_events+latest_update_events))|length, "edit", "edits") }}
+            {% endif %}
+              on {{ latest_update_events[0].createdAt|dateformat -}}
           {% else %}
             {{ all_update_events|length }}
-          {% endif %}
-          unapproved {{ pluralize((all_update_events or (oldest_update_events+latest_update_events))|length, "edit", "edits") }}
-          {%- if latest_update_events %}
-            {% if all_update_events is not none %}
-              by
-              {% if n_editing_users_min == 1 %}
-                {{ latest_update_events.0.user }}
-              {% else %}
-                {{ n_editing_users_min }} users
-              {% endif %}
-            {% endif %}
-            {% if latest_update_events[0].createdAt|dateformat == oldest_update_events[-1].createdAt|dateformat %}
-              on {{ latest_update_events[0].createdAt|dateformat }}
-            {%- else %}
-              between {{ oldest_update_events[-1].createdAt|dateformat }}
-              and {{ latest_update_events[0].createdAt|dateformat }}
-            {%- endif -%}
+            unapproved {{ pluralize((all_update_events or (oldest_update_events+latest_update_events))|length, "edit", "edits") -}}
           {%- endif -%}
           .
         </p>
--- a/app/templates/compare_revisions.html
+++ b/app/templates/compare_revisions.html
@@ -65,17 +65,24 @@
         <p>
           {% if latest_update_events %}
             {% if min_number_of_users_who_made_edits > 1 %}
-              More than one user has edited this service. The last user to edit this page was
+              More than one user has edited this service.
+              The last user to edit this service was {{ latest_update_events.0.user }}
+              on {{ latest_update_events[0].createdAt|dateformat -}}
             {% endif %}
-            {{ latest_update_events.0.user }}
-            {% if min_number_of_users_who_made_edits == 1 %}
+
+            {%- if min_number_of_users_who_made_edits == 1 -%}
+              {{ latest_update_events.0.user }}
               made {{ (all_update_events or (oldest_update_events+latest_update_events))|length }}
               {{ pluralize((all_update_events or (oldest_update_events+latest_update_events))|length, "edit", "edits") }}
-            {% endif %}
-            on {{ latest_update_events[0].createdAt|dateformat -}}
-          {% else %}
-            {{ all_update_events|length }}
-            unapproved {{ pluralize((all_update_events or (oldest_update_events+latest_update_events))|length, "edit", "edits") -}}
+              {% if latest_update_events[0].createdAt|dateformat == oldest_update_events[-1].createdAt|dateformat %}
+                on {{ latest_update_events[0].createdAt|dateformat -}}
+              {%- else %}
+                between {{ oldest_update_events[-1].createdAt|dateformat }}
+                and {{ latest_update_events[0].createdAt|dateformat }}
+              {%- endif -%}
+            {%- endif -%}
+          {%- else -%}
+            This service has no unapproved edits
           {%- endif -%}
           .
         </p>
--- a/app/templates/compare_revisions.html
+++ b/app/templates/compare_revisions.html
@@ -64,15 +64,15 @@
       <div class="column-two-thirds page-section">
         <p>
           {% if latest_update_events %}
-            {% if number_of_users_who_made_edits > 1 %}
-              The last user to edit this page was
+            {% if min_number_of_users_who_made_edits > 1 %}
+              More than one user has edited this service. The last user to edit this page was
             {% endif %}
-              {{ latest_update_events.0.user }}
-            {% if number_of_users_who_made_edits == 1 %}
+            {{ latest_update_events.0.user }}
+            {% if min_number_of_users_who_made_edits == 1 %}
               made {{ (all_update_events or (oldest_update_events+latest_update_events))|length }}
               {{ pluralize((all_update_events or (oldest_update_events+latest_update_events))|length, "edit", "edits") }}
             {% endif %}
-              on {{ latest_update_events[0].createdAt|dateformat -}}
+            on {{ latest_update_events[0].createdAt|dateformat -}}
           {% else %}
             {{ all_update_events|length }}
             unapproved {{ pluralize((all_update_events or (oldest_update_events+latest_update_events))|length, "edit", "edits") -}}
--- a/tests/app/main/views/test_services.py
+++ b/tests/app/main/views/test_services.py
@@ -1092,8 +1092,8 @@ class TestServiceUpdates(LoggedInApplicationTest):
         }[service_id]}
 
     @staticmethod
-    def _mock_get_archived_service_side_effect(OLD_VERSION_OF_SERVICE, archived_service_id):
-        return {"services": OLD_VERSION_OF_SERVICE[archived_service_id]}
+    def _mock_get_archived_service_side_effect(old_versions_of_services, archived_service_id):
+        return {"services": old_versions_of_services[archived_service_id]}
 
     @staticmethod
     def _mock_get_supplier_side_effect(supplier_id):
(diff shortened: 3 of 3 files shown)


app/main/views/services.py
```diff
@@ -199,7 +198,7 @@ def service_updates(service_id):
         all_update_events=all_update_events,
         latest_update_events=latest_update_events,
         oldest_update_events=oldest_update_events,
-        n_editing_users_min=len(frozenset(audit_event["user"] for audit_event in chain(
+        number_of_users_who_made_edits=len(frozenset(audit_event["user"] for audit_event in chain(
```
risicle: This isn't the number of editing users - this is the *minimum* possible number of editing users, as we're only looking at the first & last pages of edits, hence the original name.

app/templates/compare_revisions.html
```diff
@@ -63,27 +63,19 @@
     <div class="grid-row">
       <div class="column-two-thirds page-section">
         <p>
-          {% if all_update_events is none %}
-            Multiple
+          {% if latest_update_events %}
+            {% if number_of_users_who_made_edits > 1 %}
+              The last user to edit this page was
```
risicle: Not sure about the term "this page". Should try and keep it to the idea they're editing a *service* not just a "page".

constancecerf: Happy to change 'page' to 'service'.

app/templates/compare_revisions.html
```diff
@@ -63,27 +63,19 @@
...
+            {% endif %}
+              {{ latest_update_events.0.user }}
+            {% if number_of_users_who_made_edits == 1 %}
+              made {{ (all_update_events or (oldest_update_events+latest_update_events))|length }}
```
risicle: Again, in the case where this falls back to `oldest_update_events+latest_update_events`, this is a value that should be indicated that it's an *at least* value.

risicle: And also you can't use this value you've called `number_of_users_who_made_edits` as the actual number of users who made edits. You could have an "edit sandwich" and be missing the pages that have edits by a different user.

tests/app/main/views/test_services.py
```diff
@@ -888,274 +888,230 @@ def _mock_find_audit_events_side_effect(fae_response_src, implicit_page_len, **k
...
+        ((5, EXPECTED_LATEST_EDIT_INFO_4), (3, EXPECTED_LATEST_EDIT_INFO_4), (2, EXPECTED_LATEST_EDIT_INFO_4))
+    )
+
+    ENABLED_SERVICE_WITH_NO_EDITS = (
+        (),
+        {},
+        "enabled",
+        None,
+        ((5, "0 unapproved edits.",),)
+    )
```
risicle: If you really think this is a better and more maintainable way of declaring these then I don't know what to say. It might *look* clearer right now as it is now (I don't think so personally, I have to do a bunch of variable chasing to see what references what), but the first time someone comes along to modify/attempt to reuse any of these scenarios it will start to go down a dark path.

app/templates/compare_revisions.html
```diff
@@ -63,27 +63,19 @@
...
-                {{ n_editing_users_min }} users
-              {% endif %}
-            {% endif %}
-            {% if latest_update_events[0].createdAt|dateformat == oldest_update_events[-1].createdAt|dateformat %}
-              on {{ latest_update_events[0].createdAt|dateformat }}
-            {%- else %}
-              between {{ oldest_update_events[-1].createdAt|dateformat }}
-              and {{ latest_update_events[0].createdAt|dateformat }}
-            {%- endif -%}
+            unapproved {{ pluralize((all_update_events or (oldest_update_events+latest_update_events))|length, "edit", "edits") -}}
```
risicle: I don't know if I really see the point in this change. I _did actually_ think this all this wording out quite hard based on the information we are actually able to _reliably_ show people, and it was only a few months ago. Is this churn really necessary?

app/main/views/services.py
CrystalPea: Good call, I changed the name of this variable and added a comment. Still, as per the comment, there would need to be a whole lot of unapproved edits for this code to stop showing the definite total number of users who made unapproved edits.

app/templates/compare_revisions.html
CrystalPea: Cool, I've changed that now :).

app/templates/compare_revisions.html
CrystalPea: This change has been made following user research done with people using admin page. They wanted to see who made last edit if more than one person made an edit, to be able to contact that person.  Earlier this information was not shown. After the need / pain point was recognised, it was run by our team, including product manager and content designer, and this ticket is the result of this process.
If you would like to question it, I think the best person might be Tobi.

app/main/views/services.py
risicle: Could easily happen (who knows when someone will decide to change the default page length for calls like this), and when it does happen it will quite happily show a number that is orders of magnitude wrong.

Or very happily make it look like all edits were done by a single (possibly admin?) user when they weren't.

app/templates/compare_revisions.html
risicle: It would have been really easy to prepend the most recent editor's email to the sentence in the case when a user wasn't shown and make minimal changes. There are a million things that *really need doing* on this page and instead of building the road forward we're deciding to tear up the bit that I've just tarmac'ed.

package.json
```diff
@@ -5,7 +5,7 @@
   "private": true,
   "engine": "node >= 4.2",
   "dependencies": {
-    "bower": "1.7.1",
+    "bower": "1.8.0",
     "del": "1.1.1",
```
benvand: This change shouldn't be propagated to production. We upped this version to force the admin-fe to work on your local environment rather than for some need based on the repo as a whole.

tests/app/main/views/test_services.py
CrystalPea: I worked on this refactor with @katstevens and both of us think it is far more readable in the new version. It also made it much easier for us to split asserting the message about recent edits into a separate test.
I do not understand why would things go down the dark path if someone tried to modify or attempt to reuse any of these scenarios in current setup (it went quite well actually when we did it), but not before. Would you mind explaining?
And what is declarative notation?

app/templates/compare_revisions.html
TheDoubleK: I agree with Rob - it would really be a shame to lose the nuance and additional information here.  I know the ticket says to use a blanket "More than one user has edited this service.' but that will only work until the day comes when admin need more information.

The ticket is about showing who the last user to edit the service was.  Like Rob says, the easy win here is just to add "The last user to edit this service was X" to the existing information where necessary.

tests/app/main/views/test_services.py
katstevens: In general, the learning curve to parse all these nested tuples and figure out what we were actually meant to be testing in each case was very steep. The changes in this PR so far feel like a step in the right direction, without losing any subtleties in the combination of parameters.

Saying that, we could probably make our test data a bit less ALL CAPS now that we've clarified what the scenarios are. I'd be happy with (carefully named!) helper functions that set up the elements of these scenarios, that would hopefully be more reusable.

Re: scenario setups - if a more complex scenario meant this test is no longer fit for purpose, it feels like that test should be separate anyway. I was tempted to split out the empty case anyway as the assertions are very different, but thought that might be change too far for this story 😸 .

app/templates/compare_revisions.html
TheDoubleK: @CrystalPea - to follow up on your question - the current version of the page gives admin users the most complete and accurate summary of the changes that have been made to the service that we can, given the data available to the view.  It seems to me that admin users should be interested in a complete and accurate picture of what has gone on.

So I'd rather see only the addition of the extra info here (i.e. who made the latest change, as the headline of the ticket says), rather than losing this detail.

tests/app/main/views/test_services.py
risicle: So previously the declaration of the data structure had a form of *referential transparency*. The structure of the file was a fairly strict tree-like hierarchy and every pair of brackets defined a structure or generator completely contained within it. In comparison, variables-referencing-each-other-all-over-the-place is spaghetti, exposing all kinds of internal details that you don't necessarily need to expose. This makes it less clear what the intent of the definitions are (what the intended *output* of the little variable dance is supposed to be), which are *supposed* to be used in tests, etcetera...

Gone is the encapsulation and ability to double click on a bracket and have your editor highlight the whole relevant section of declaration and know there are *no* side effects.

app/templates/compare_revisions.html
CrystalPea: Ok we can talk about this tomorrow on standup with Tobi and Constance

TheDoubleK:
I've been thinking about this a bit...

The original ticket that we discussed in sprint planning and scored was a quick win one-pointer to add the email address of the most recent user to edit a service.

Additional changes got added to the ticket after the story was written and planned, which is why we are where we are now I think.

I think the user need that we are catering to with this page is something like "as a framework owner and administrator I need a clear view of changes happening on my framework, so that I can have the greatest possible confidence in monitoring and assuring those changes".

Adding the email address to the page helps us meet that need, and so is a thing we should do. The other changes to the summary - making it less precise - I think are a step backwards from meeting that need, so we shouldn't do it.

The supplier-editing-services feature has been live for about a month now, which is no time at all to get a feel fort the range and combinations of changes that might be made by suppliers.

We did a check-in with CCS about how it was going, which is where this additional email address requirement came from in the first place. I don't believe that an aside that "we don't really look at the numbers much" equates to a user need to not have that information presented to them. At best it is a user "want", which is not what we cater to. The underlying need - which is what we cater for - is to minimize the risk of bad things happening in the framework, and that is best served by the existing summary plus a sentence about who last edited it in the case of multiple users' edits, just as the ticket originally stated.

TheDoubleK:
In terms of the testing, of the two possible versions here I prefer the original, sorry. All the CONSTANTS that than somehow get pulled together for a test are really confusing I think.

So overall I'd propose rolling back the changes here, leave the summary on the page as it is, and add just the new email address bit where necessary, as per the original ticket. The test changes required would then be minimal, you might even be able to just write a new test for the email address part and leave existing tasty as-is.

I appreciate that that test takes a while to get your head round - originally I balked at it -  but the audit history setup is subtle and tricky to test the possible variations - however it's done I think it's going to require some considerable thought to figure out the expected result. The way the current test setup is done  might not be appropriate for all pages, but in this instance I think it's an elegant and comprehensive solution. As developers we should be able to use the tools available to us. If that means our colleagues learn something along the way then that's not a bad thing.

app/templates/compare_revisions.html
```diff
@@ -63,27 +63,19 @@
...
+          {% if latest_update_events %}
+            {% if min_number_of_users_who_made_edits > 1 %}
+              More than one user has edited this service. The last user to edit this page was
+            {% endif %}
+            {{ latest_update_events.0.user }}
+            {% if min_number_of_users_who_made_edits == 1 %}
+              made {{ (all_update_events or (oldest_update_events+latest_update_events))|length }}
+              {{ pluralize((all_update_events or (oldest_update_events+latest_update_events))|length, "edit", "edits") }}
+            {% endif %}
+            on {{ latest_update_events[0].createdAt|dateformat -}}
```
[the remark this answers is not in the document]
TheDoubleK: The alternative would be to always say "The service was last edited on..."

app/templates/compare_revisions.html
```diff
@@ -63,27 +63,19 @@
     <div class="grid-row">
       <div class="column-two-thirds page-section">
         <p>
-          {% if all_update_events is none %}
-            Multiple
+          {% if latest_update_events %}
+            {% if min_number_of_users_who_made_edits > 1 %}
+              More than one user has edited this service. The last user to edit this page was
```
TheDoubleK: Should be `The last user to edit this service` - not page.

app/templates/compare_revisions.html
```diff
@@ -63,27 +63,19 @@
...
-                {{ n_editing_users_min }} users
-              {% endif %}
-            {% endif %}
-            {% if latest_update_events[0].createdAt|dateformat == oldest_update_events[-1].createdAt|dateformat %}
-              on {{ latest_update_events[0].createdAt|dateformat }}
-            {%- else %}
-              between {{ oldest_update_events[-1].createdAt|dateformat }}
-              and {{ latest_update_events[0].createdAt|dateformat }}
-            {%- endif -%}
+            unapproved {{ pluralize((all_update_events or (oldest_update_events+latest_update_events))|length, "edit", "edits") -}}
```
TheDoubleK: The way this code is restructured now it looks like this "else" part can only ever be reached if there are no edits to the service, so it could just say `0 unapproved edits` here rather than `{{ all_update_events|length }} unapproved {{ pluralize((all_update_events or (oldest_update_events+latest_update_events))|length, "edit", "edits") -}}`

Could even be a nicer sentence like `This service has no unapproved edits` or something.

Or have I missed some case where we might have no `latest_update_events` but still some edits to report?

app/templates/compare_revisions.html
```diff
@@ -63,27 +63,19 @@
     <div class="grid-row">
       <div class="column-two-thirds page-section">
         <p>
-          {% if all_update_events is none %}
-            Multiple
+          {% if latest_update_events %}
+            {% if min_number_of_users_who_made_edits > 1 %}
+              More than one user has edited this service. The last user to edit this page was
+            {% endif %}
+            {{ latest_update_events.0.user }}
```
TheDoubleK: I find the interleaving of messages here (to avoid repetition of `{{ latest_update_events.0.user }}` in the template) quite difficult to follow.  I think something like this would make it easier to see what's going on with the different possible cases:
```
<p>
  {% if latest_update_events %}

    {% if min_number_of_users_who_made_edits > 1 %}
      More than one user has edited this service.
      The last user to edit this service was {{ latest_update_events.0.user }}
      on {{ latest_update_events[0].createdAt|dateformat }}
    {%- endif -%}

    {% if min_number_of_users_who_made_edits == 1 %}
      {{ latest_update_events.0.user }} made {{ (all_update_events or (oldest_update_events+latest_update_events))|length }}
      {{ pluralize((all_update_events or (oldest_update_events+latest_update_events))|length, "edit", "edits") }}
      {% if latest_update_events[0].createdAt|dateformat == oldest_update_events[-1].createdAt|dateformat %}
        on {{ latest_update_events[0].createdAt|dateformat }}
      {%- else %}
        between {{ oldest_update_events[-1].createdAt|dateformat }}
        and {{ latest_update_events[0].createdAt|dateformat }}
      {%- endif -%}
    {%- endif -%}

  {% else %}
    0 unapproved edits
  {%- endif -%}
  .
</p>
```

app/templates/compare_revisions.html
CrystalPea: Absolutely, changing this now

app/templates/compare_revisions.html
CrystalPea: This looks very sensible :)

app/templates/compare_revisions.html
CrystalPea: Good point, I don't see how we could have no latest_update_edits but have some other ones.

tests/app/main/views/test_services.py
```diff
@@ -888,274 +888,247 @@ def _mock_find_audit_events_side_effect(fae_response_src, implicit_page_len, **k
...
+
     @pytest.mark.parametrize(
-        "fae_response_src,avail_arch_svcs,svc_status,exp_oldver_ctxt_text,fae_page_len,exp_summ_text",
+        "find_audit_events_api_response,old_versions_of_services,service_status,"
+        "expected_oldest_edit_info,find_audit_events_page_length",
         chain.from_iterable(
             (
-                (fae_r_s, avail_a_s, svc_s, exp_ov_c_t, fae_p_l, exp_s_t,)
+                (fae_r_s, avail_a_s, svc_s, exp_ov_c_t, fae_p_l,)
```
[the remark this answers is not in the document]
CrystalPea: Good point 👍

PR #310, opened 2017-11-17
Refactor change view supplier service url to be consistent with rest of app

benvand:


benvand:
From @ash
> Hi Ben. We currently don't have analytics instrumented for the admin section and there's no plans to do so. So you're free to do what you want. Cheers

PR #311, opened 2017-11-17
Order lists in tests to prevent intermittent failure

benvand:


tests/app/main/views/test_services.py
```diff
@@ -715,20 +715,22 @@ def test_service_update_with_multiquestion_validation_error(self, data_api_clien
         validation_banner_h1 = document.xpath("//h1[@class='validation-masthead-heading']//text()")[0].strip()
         assert validation_banner_h1 == "There was a problem with your answer to:"
 
-        validation_banner_links = [
+        validation_banner_links = sorted([
```
katstevens: Does this mean our errors are coming back unsorted in the view? I don't know if the content loader works differently in the admin, but shouldn't the fields be ordered according to the manifest?

benvand:
For posterity: Don't need sorted on the validation_banner_links list if it's put in the correct order

PR #312, opened 2017-11-17
Show admins who removed a service and when

katstevens:
Trello ticket: https://trello.com/c/ytFyrbql/144-show-admins-who-removed-a-service-and-when

When a service has been removed (either by an admin user, or the supplier themselves), a banner is displayed in the admin saying who removed it and when. The date in the banner should have non-breaking spaces.

For a visible (Public) service there is no change:
service-admin-info-banner-public

For removed/private services, we show the same banner whether it's been removed by the supplier or by an admin:
service-admin-info-banner-private
service-admin-info-banner-removed

Happily there is a separate audit event for status changes, so finding the right event is straightforward.

### DIFF
--- a/app/main/views/services.py
+++ b/app/main/views/services.py
@@ -51,6 +51,18 @@ def view(service_id):
         flash({'api_error': service_id}, 'error')
         return redirect(url_for('.index'))
 
+    most_recent_audit_events = data_api_client.find_audit_events(
+        latest_first="true",
+        object_id=service_id,
+        object_type="services",
+        audit_type=AuditTypes.update_service_status
+    )
+    if most_recent_audit_events.get('auditEvents'):
+        last_updated_by = most_recent_audit_events['auditEvents'][0]['user']
+        last_updated_date = most_recent_audit_events['auditEvents'][0]['createdAt']
+    else:
+        last_updated_by = last_updated_date = None
+
     service_data['priceString'] = format_service_price(service_data)
     content = content_loader.get_manifest(service_data['frameworkSlug'], 'edit_service_as_admin').filter(service_data)
 
@@ -58,7 +70,9 @@ def view(service_id):
         "view_service.html",
         sections=content.summary(service_data),
         service_data=service_data,
-        service_id=service_id
+        service_id=service_id,
+        last_updated_by=last_updated_by,
+        last_updated_date=last_updated_date,
     )
 
 
--- a/app/main/views/services.py
+++ b/app/main/views/services.py
@@ -51,7 +51,7 @@ def view(service_id):
         flash({'api_error': service_id}, 'error')
         return redirect(url_for('.index'))
 
-    last_updated_by = last_updated_date = None
+    removed_by = removed_at = None
     if service_data['status'] != 'published':
         most_recent_audit_events = data_api_client.find_audit_events(
             latest_first="true",
@@ -60,8 +60,8 @@ def view(service_id):
             audit_type=AuditTypes.update_service_status
         )
         if most_recent_audit_events.get('auditEvents'):
-            last_updated_by = most_recent_audit_events['auditEvents'][0]['user']
-            last_updated_date = most_recent_audit_events['auditEvents'][0]['createdAt']
+            removed_by = most_recent_audit_events['auditEvents'][0]['user']
+            removed_at = most_recent_audit_events['auditEvents'][0]['createdAt']
 
     service_data['priceString'] = format_service_price(service_data)
     content = content_loader.get_manifest(service_data['frameworkSlug'], 'edit_service_as_admin').filter(service_data)
@@ -71,8 +71,8 @@ def view(service_id):
         sections=content.summary(service_data),
         service_data=service_data,
         service_id=service_id,
-        last_updated_by=last_updated_by,
-        last_updated_date=last_updated_date,
+        removed_by=removed_by,
+        removed_at=removed_at,
     )
 
 
--- a/app/main/views/services.py
+++ b/app/main/views/services.py
@@ -51,17 +51,17 @@ def view(service_id):
         flash({'api_error': service_id}, 'error')
         return redirect(url_for('.index'))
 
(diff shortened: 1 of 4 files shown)


app/main/views/services.py
```diff
@@ -51,14 +51,28 @@ def view(service_id):
         flash({'api_error': service_id}, 'error')
         return redirect(url_for('.index'))
 
+    most_recent_audit_events = data_api_client.find_audit_events(
```
TheDoubleK: Can we check the service status here and not bother fetching any audit events at all if the service is 'published' (as 99% of services will be).  Searching the audit table feels like a reasonably heavy query on the API, and even though I think we have a partial index to support this one, it would be better to not make the call if we already know we're not going to need any audit information.

app/main/views/services.py
```diff
@@ -51,14 +51,28 @@ def view(service_id):
...
+        latest_first="true",
+        object_id=service_id,
+        object_type="services",
+        audit_type=AuditTypes.update_service_status
+    )
+    if most_recent_audit_events.get('auditEvents'):
+        last_updated_by = most_recent_audit_events['auditEvents'][0]['user']
```
TheDoubleK: I think `removed_by` and `removed_on` would be better variable names here - the service might have been updated since it was removed, but it's the removal we care about here.

tests/app/main/views/test_services.py
```diff
@@ -31,8 +31,20 @@ def test_service_find_returns_404_for_missing_service_id(self):
         assert response.status_code == 404
 
 
+@mock.patch('app.main.views.services.data_api_client', autospec=True)
 class TestServiceView(LoggedInApplicationTest):
-    @mock.patch('app.main.views.services.data_api_client')
+
+    find_audit_events_api_response = {'auditEvents': [
+        {
```
TheDoubleK: Could flesh these out a bit beyond minimal data to make it clear what we expect real responses to look like.  Doesn't need to have absolutely all fields, but something like this might be nice to give a bit more context:
```
{
      "createdAt": "2017-11-17T16:30:30.837267Z",
      "data": {
          "new_status": "disabled",
          "old_status": "published",
          "serviceId": "314159265",
      },
      "type": "update_service_status",
      "user": "anne.admin@example.com" 
}
```

tests/app/main/views/test_services.py
```diff
@@ -460,8 +518,9 @@ def test_service_edit_with_no_section_returns_404(self, data_api_client):
         assert response.status_code == 404
 
 
+@mock.patch('app.main.views.services.data_api_client', autospec=True)
```
TheDoubleK: Nice one pulling these patches out to class level 🎉

tests/app/main/views/test_services.py
```diff
@@ -849,6 +900,7 @@ def test_cannot_make_removed_service_public(self, data_api_client):
             'supplierId': 2,
             'status': 'disabled'
         }}
+        data_api_client.find_audit_events.return_value = {'auditEvents': []}
```
TheDoubleK: I'm sure there's a way to set this in the patch at the class level too, rather than having to add the same value  to each test - I have seen it done in our codebase somewhere. Probably by Rob. But I can't seem to remember/find the syntax right now 😞

app/main/views/services.py
katstevens: Yeah, good call. I might have to stick up a permanent post it note to myself saying 'QUERYING AUDIT EVENTS = EXPENSIVE' as I do forget sometimes...

tests/app/main/views/test_services.py
katstevens: I think you can yield the return value in a fancy way, yes. Will have a look....

katstevens: We can do something like:

`@mock.patch('app.main.views.services.data_api_client', autospec=True, **{'find_audit_events.return_value': {'auditEvents': []}})`

It's a bit messy :(

TheDoubleK: Yeah I wasn't saying we definitely need it.  That's not so pretty, so fine as is I reckon.

tests/app/main/views/test_services.py
katstevens: Sure - I'll include the AuditType as well so it's clear we're returning a `update_service_status` event rather than a general edit.

app/main/views/services.py
```diff
@@ -51,14 +51,28 @@ def view(service_id):
...
+    last_updated_by = last_updated_date = None
+    if service_data['status'] != 'published':
+        most_recent_audit_events = data_api_client.find_audit_events(
+            latest_first="true",
+            object_id=service_id,
+            object_type="services",
+            audit_type=AuditTypes.update_service_status
+        )
+        if most_recent_audit_events.get('auditEvents'):
+            last_updated_by = most_recent_audit_events['auditEvents'][0]['user']
```
TheDoubleK: I still like `removed_by` and `removed_on` for these variables...

TheDoubleK: Maybe `removed_at` if it's a full timestamp, actually.

app/templates/view_service.html
```diff
@@ -47,6 +47,20 @@
     {% endif %}
   {% endwith %}
 
+{% if service_data['status'] != 'published' and last_updated_by %}
+<div class="grid-row">
+  <div class="column-one-whole">
+    {%
+      with
+      type = "temporary-message",
+      heading = "Removed by {} on ".format(last_updated_by) + "{}.".format(last_updated_date|dateformat)|nbsp
```
TheDoubleK: Is there a reason for not doing this as one string `"Removed by {} on {}."`?

katstevens: Yeah - the non breaking space on the date. If there is a less messy way to do it I am happy to change.

app/main/views/services.py
katstevens: Whoops - I thought I'd changed that. Now fixed.

PR #316, opened 2017-11-23
155 allow admin manager to invite new admin users

benvand:
Trello:
https://trello.com/c/2Uc2WMl2/155-allow-super-admin-invite-new-admin-users
Prototype:
https://docs.google.com/document/d/1MjvYx9aQWgcbryiaGrRLOPt_lVm8gIbqOH3vmkKx_fo/edit#

* Refactor imports in `app/main/forms.py` to indicate function of imported classes
* Update `flask_wtf` import in app/main/forms.py to use newer `flask.ext.wtf`
* Up `dmapiclient` `11.4.0` > `12.1.0` to get new `email_is_valid_for_admin_user` method
* Add new validator in `app/main/forms.py` that applies `data_api_client.email_is_valid_for_admin_user` to a given field
* Add a new form `InviteAdminForm` with `email_address` and `role` fields
* Add new `GET`/`POST` route to invite a new admin user `app/main/views/admin_manager.py::invite_admin_user`
* Add new template `invite_admin_user.html` to display invite admin form
* Import new view into `app/main` so views are applied to app
* Add tests for new validator `app.main.forms.AdminEmailAddressValidator`
* Add tests for new view `app.main.views.admin_manager.invite_admin_user`

manageusers
errors
invite
done

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -1,5 +1,5 @@
 from flask.ext.wtf import Form
-from wtforms import validators
+from wtforms import RadioField, validators
 
 from dmutils.forms import StripWhitespaceStringField
 
@@ -34,3 +34,29 @@ class EmailDomainForm(Form):
     new_buyer_domain = StripWhitespaceStringField('Add a buyer email domain', validators=[
         validators.DataRequired(message="The domain field can not be empty.")
     ])
+
+
+class InviteAdminForm(Form):
+    role_choices = [
+        ('admin-ccs-category', 'Category'),
+        ('admin-ccs-sourcing', 'Sourcing'),
+        ('admin', 'Support'),
+    ]
+
+    email_address = StripWhitespaceStringField(
+        'Email address',
+        validators=[
+            validators.DataRequired(message='You must provide an email address'),
+            validators.Email(message='Please enter a valid email address'),
+            AdminEmailAddressValidator(message='The email address must belong to an approved domain')
+        ]
+    )
+    role = RadioField(
+        'Permissions',
+        validators=[validators.InputRequired(message='You must choose a permission')],
+        choices=role_choices
+    )
+
+    def __init__(self, *args, **kwargs):
+        super(InviteAdminForm, self).__init__(*args, **kwargs)
+        self.role.toolkit_macro_options = [{'value': i[0], 'label': i[1]} for i in self.role_choices]
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -1,23 +1,24 @@
 from flask_wtf import Form
-from wtforms.validators import DataRequired, Email
+from wtforms import validators
+
 from dmutils.forms import StripWhitespaceStringField
 
 
 class EmailAddressForm(Form):
     email_address = StripWhitespaceStringField('Email address', validators=[
-        DataRequired(message="Email can not be empty"),
-        Email(message="Please enter a valid email address")
+        validators.DataRequired(message="Email can not be empty"),
+        validators.Email(message="Please enter a valid email address")
     ])
 
 
 class MoveUserForm(Form):
     user_to_move_email_address = StripWhitespaceStringField('Move an existing user to this supplier', validators=[
-        DataRequired(message="Email can not be empty"),
-        Email(message="Please enter a valid email address")
+        validators.DataRequired(message="Email can not be empty"),
+        validators.Email(message="Please enter a valid email address")
     ])
 
 
 class EmailDomainForm(Form):
     new_buyer_domain = StripWhitespaceStringField('Add a buyer email domain', validators=[
-        DataRequired(message="The domain field can not be empty.")
+        validators.DataRequired(message="The domain field can not be empty.")
     ])
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -3,6 +3,18 @@ from wtforms import validators
 
 from dmutils.forms import StripWhitespaceStringField
 
+from .. import data_api_client
+
+
+class AdminEmailAddressValidator(object):
+
+    def __init__(self, message=None):
+        self.message = message
+
+    def __call__(self, form, field):
+        if not data_api_client.email_is_valid_for_admin_user(field.data):
+            raise validators.StopValidation(self.message)
+
 
 class EmailAddressForm(Form):
     email_address = StripWhitespaceStringField('Email address', validators=[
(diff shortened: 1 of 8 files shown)


app/main/views/admin_manager.py
```diff
@@ -28,3 +31,33 @@ def manage_admin_users():
...
+    form = InviteAdminForm(request.form)
+
+    if request.method == 'POST' and form.validate_on_submit():
+        email_address = form.data.get('email_address')
+        role = form.data.get('role')
+
+        send_user_account_email(
+            role,
+            email_address,
+            '08ab7791-6038-4ad2-9560-740bbcb675b7',
```
katstevens: Can we pull this out as something like `NOTIFY_TEMPLATE_ID = "..."` so it's clear what this mysterious string is? 😸

tests/app/main/views/test_admin_manager.py
```diff
@@ -108,3 +109,79 @@ def test_should_list_alphabetically_with_all_suspended_users_below_active_users(
...
+    @pytest.mark.parametrize('role', ('admin', 'admin-ccs-sourcing', 'admin-ccs-category'))
+    @mock.patch('app.main.views.admin_manager.send_user_account_email')
+    def test_successful_post_redirects(self, data_api_client, send_user_account_email, role):
+        data_api_client.email_is_valid_for_admin_user.return_value = True
+        res = self.client.post('/admin/admin-users/invite', data={'role': role, 'email_address': 'test@test.com'})
+        assert res.status_code == 302
+        assert res.location == "http://localhost/admin/admin-users"
+
+    @mock.patch('app.main.views.admin_manager.send_user_account_email')
+    def test_successful_post_sends_email(self, send_user_account_email, data_api_client):
```
katstevens: Should we bother parametrizing this test for each role as well?

app/main/forms.py
```diff
@@ -3,6 +3,18 @@
...
+
+
+class AdminEmailAddressValidator(object):
+
+    def __init__(self, message=None):
+        self.message = message
+
+    def __call__(self, form, field):
+        if not data_api_client.email_is_valid_for_admin_user(field.data):
+            raise validators.StopValidation(self.message)
```
katstevens: Probably something to do at the API level rather than here, but should we check whether the email address belongs to an existing user account? Obviously that user would get an error when they tried to sign up again so not a dealbreaker, but it would save sending an unnecessary email...

app/main/views/admin_manager.py
```diff
@@ -28,3 +31,33 @@ def manage_admin_users():
...
+            personalisation={'name': current_user.name}
+        )
+        flash('An invitation has been sent to {}.'.format(email_address), category='success')
+        return redirect(url_for('main.manage_admin_users'))
+
+    errors = {
+        key: {'question': form[key].label.text, 'input_name': key, 'message': form[key].errors[0]}
+        for key, value in form.errors.items()
+    }
```
katstevens: I noticed on the form error screengrab that the errors in the header are listed out of order. I only mention this as I just fixed the equivalent error on the content loader! https://github.com/alphagov/digitalmarketplace-content-loader/pull/52/ 

Completely up to you whether you think it's worth bothering to sort a very basic list of two errors 😸

tests/app/main/test_form_validators.py
```diff
@@ -0,0 +1,29 @@
...
+        self.form_mock.email_address.data = 'the_email_address'
+
+    def test_admin_email_address_validator_calls_api(self, data_api_client):
+        AdminEmailAddressValidator()(self.form_mock, self.form_mock.email_address)
+        assert data_api_client.email_is_valid_for_admin_user.called_once_with('the_email_address')
+
+    def test_admin_email_address_validator_raises_with_invalid_response(self, data_api_client):
+        data_api_client.email_is_valid_for_admin_user.return_value = False
+
+        with pytest.raises(StopValidation, message='The email address must belong to an approved domain'):
```
katstevens: This test class is for the validator - should it know about the `InviteAdminForm`'s message? Can we test what happens when the default message (`None`, I think) is used?

tests/app/main/views/test_form_validators.py
```diff
@@ -0,0 +1,29 @@
...
+    def test_admin_email_address_validator_raises_with_invalid_response(self, data_api_client):
+        data_api_client.email_is_valid_for_admin_user.return_value = False
+
+        with pytest.raises(StopValidation, message='The email address must belong to an approved domain'):
+            AdminEmailAddressValidator()(self.form_mock, self.form_mock.email_address)
+
+    def test_admin_email_address_validator_passes_with_valid_response(self, data_api_client):
+        data_api_client.email_is_valid_for_admin_user.return_value = True
+
+        assert AdminEmailAddressValidator()(self.form_mock, self.form_mock.email_address) is None
```
katstevens: Are these tests duplicates of `tests/app/main/test_form_validators.py`?

tests/app/main/test_form_validators.py
benvand: So the None case is just what happens with all validators (raises with no message) but I've changed these tests to agnostic of the invite form now

tests/app/main/views/test_admin_manager.py
benvand: Made this a separate test that _only_ tests that the view works for the valid roles. The test test_post_requires_valid_role  above tests for bad role.

app/main/views/admin_manager.py
benvand: Done 👍

app/main/forms.py
benvand: Hmm interesting, could be done on the apiclient method or api. Probably something to talk to product off. I'll bring it up with tobi

benvand:
@katstevens Thanks for the review. Dunno how that duplicate file crept in there 🙃

benvand:
Weirdly that bad error ordering isn't happening any more. Could have been I just mixed the fields round in development
screen shot 2017-11-28 at 08 38 21

benvand:
Yup, swapping them round on the form declaration changes the error message ordering but the template preserves the field ordering

app/main/forms.py
benvand: Gonna stick this screenshot here for now so I don't lose it

benvand: screen shot 2017-11-28 at 09 08 47

PR #320, opened 2017-11-29
Adjust Support and Category role permissions

TheDoubleK:
Changes required for these TWO tickets:
https://trello.com/c/GKq9PcC0/161-edit-permissions-for-dm-admins-support
https://trello.com/c/0iWPAKel/162-edit-permissions-for-category

These are rolled together because some permissions have moved from Support to Category and it was nice to be able to continue to test these functions rather than disable/comment out tests for actions that no user could perform.  For the same reason - to keep functionality testable - I have added the `admin-framework-manager` role to some routes, even though in theory we don't have this role defined yet (will be part of https://trello.com/c/iUS9D0ll/163-add-new-framework-manager-role).

I also added some extra test coverage where things didn't break when I expected them to.  (I haven't gone all-out on new tests for the index page yet though, as big changes so close on the horizon, we may as well sort out full coverage for that as the new design gets implemented.)

**Reviewer(s) note**: It might be easier to see how the view changes relate to test changes by looking through individual commits rather than the changeset as a whole.

### DIFF
--- a/app/templates/index.html
+++ b/app/templates/index.html
@@ -73,7 +73,7 @@
           <input type="submit" value="Search" class="button-save">
         </form>
 
-        {% if current_user.has_any_role('admin') %}
+        {% if current_user.has_any_role('admin', 'admin-ccs-category') %}
 
         <form action="{{ url_for('.find_user_by_email_address') }}" method="get" class="question">
           <label class="question-heading" for="email_address">Find users by email address</label>
--- a/app/templates/index.html
+++ b/app/templates/index.html
@@ -173,7 +173,7 @@
     {% endfor %}
   {% endif %}
 
-  {% if current_user.has_any_role('admin', 'admin-ccs-category') %}
+  {% if current_user.has_any_role('admin-ccs-category', 'admin-framework-manager') %}
     {% set items = [
       {
         "body": "Manage G-Cloud 8 communications",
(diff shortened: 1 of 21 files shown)


tests/app/main/views/test_services.py
```diff
@@ -1109,7 +1093,23 @@ def test_services_with_missing_id(self, data_api_client):
...
+        ("admin-manager", 403),
+    ])
+    def test_view_service_updates_is_only_accessible_to_specific_user_roles(
+            self, data_api_client, html_diff_tables_from_sections_iter, role, expected_code
+    ):
+        self.user_role = role
+        data_api_client.get_service.return_value = self._mock_get_service_side_effect("published", "151")
+        response = self.client.get('/admin/services/31415/updates')
+        actual_code = response.status_code
+        assert actual_code == expected_code, "Unexpected response {} for role {}".format(actual_code, role)
```
katstevens: Should we have an equivalent test for allowed roles on the service update (and status update) POST routes?

app/templates/index.html
```diff
@@ -173,7 +173,7 @@
     {% endfor %}
   {% endif %}
 
-  {% if current_user.has_any_role('admin', 'admin-ccs-category') %}
+  {% if current_user.has_role('admin-framework-manager') %}
```
katstevens: The future! 😻

TheDoubleK:
Cheers Kat, I've added those tests, will merge once Travis passes.

katstevens:
Cracking 👍 Good to go.

PR #321, opened 2017-11-29
152-admin-can-toggles-service-status-public-removed

benvand:
Trello:
https://trello.com/c/sco9Pz3H/152-admin-can-toggles-service-status-public-removed

Prototype:
https://docs.google.com/document/d/1lRu5a4hx6ZLy77eiVKtjhKvrFdBzR1iiRlUGWlyVXYE/edit

* Allow `publish` and `remove` query params on view service view
  * Propagate these back to template
* Remove old update status form on bottom of page
  * Update tests for this
* Add 'Remove service' link for admins if service not `published`
* Add 'Publish service' link for admins if service not `published`
  * Link on 'This service was removed by' banner
* Add new 'Are you sure you want to publish' banner, with action to puplish
* Add new 'Are you sure you want to remove' banner with action to remove
* Tests, including updates to existing

screen shot 2017-11-30 at 14 54 04

screen shot 2017-11-29 at 17 14 20

screen shot 2017-11-29 at 17 14 29

screen shot 2017-11-29 at 17 14 40

screen shot 2017-11-29 at 17 14 51

### DIFF
--- a/app/templates/view_service.html
+++ b/app/templates/view_service.html
@@ -107,31 +107,6 @@
         {% endcall %}
       {% endcall %}
     {% endfor %}
-
-    <form action="{{ url_for('.update_service_status', service_id=service_id ) }}" method="post">
-        <div style="display:none;"><input id="csrf_token" name="csrf_token" type="hidden" value="{{ csrf_token() }}"></div>
-        <fieldset class="question">
-            <legend class="question-heading">
-                Service status
-            </legend>
-            <label class="selection-button">
-                <input type="radio" name="service_status" id="service_status_disabled" value="removed" {% if service_data['status'] == 'disabled' %}checked="checked"{% endif %} />
-                Removed
-            </label>
-            <label class="selection-button">
-                <input type="radio" name="service_status" id="service_status_private" value="private" {% if service_data['status'] == 'enabled' %}checked="checked"{% endif %} />
-                Private
-            </label>
-            {% if service_data.status != 'disabled' %}
-            <label class="selection-button">
-                <input type="radio" name="service_status" id="service_status_published" value="public" {% if service_data['status'] == 'published' %}checked="checked"{% endif %} />
-                Public
-            </label>
-            {% endif %}
-        </fieldset>
-        <button type="submit" class="button-save">Update status</button>
-    </form>
-
   {% else %}
     <h1>Error</h1>
     <p>
--- a/app/templates/view_service.html
+++ b/app/templates/view_service.html
@@ -71,15 +71,16 @@
       {% include "toolkit/page-heading.html" %}
     {% endwith %}
 
-    {% if service_data['frameworkFramework'] == 'g-cloud' %}
-      {%
-        with
-        url = "/{}/services/{}".format(service_data['frameworkFramework'], service_id),
-        text = "View service"
-      %}
-        {% include "toolkit/secondary-action-link.html" %}
-      {% endwith %}
-    {% endif %}
+    {# links #}
+    <ul class="list-no-bullet">
+      {% if service_data['frameworkFramework'] == 'g-cloud' %}
+        <li><a href="{{ "/{}/services/{}".format(service_data['frameworkFramework'], service_id) }}">View service</a></li>
+      {% endif %}
+      {% if current_user.has_role('admin-ccs-category') and service_data['status'] == 'published' %}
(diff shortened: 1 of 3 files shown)


TheDoubleK:
Going to look at the code now, but in the meantime, FWIW I like the look of links closer together, but probably worth showing @laurenceberry on your machine if possible to get a call on best layout.

The "foo" commit probably wants squashing in somewhere.

laurenceberry:
Ben and I took a look at the link spacing and came up with a good solution by wrapping them in a <ul class="list-no-bullet">.

app/templates/view_service.html
```diff
@@ -71,15 +111,16 @@
...
-    {% if service_data['frameworkFramework'] == 'g-cloud' %}
-      {%
-        with
-        url = "/{}/services/{}".format(service_data['frameworkFramework'], service_id),
-        text = "View service"
-      %}
-        {% include "toolkit/secondary-action-link.html" %}
-      {% endwith %}
-    {% endif %}
+    {# links #}
```
TheDoubleK: Same with this - is it just a name to make things easier to read?  Or does it do something?

tests/app/main/views/test_services.py
```diff
@@ -90,12 +92,6 @@ def test_service_view_no_features_or_benefits_status_disabled(self, data_api_cli
             "normalize-space(string(//td[@class='summary-item-field']//*[@class='service-id']))"
         ) == "314159265"
 
-        assert frozenset(document.xpath("//input[@type='radio'][@name='service_status']/@value")) == frozenset((
```
TheDoubleK: Are some of these tests duplicates now that the assertions about statuses have gone?  Feels like we can lose one or two of them altogether probably?

tests/app/main/views/test_services.py
```diff
@@ -239,6 +221,29 @@ def test_service_view_shows_info_banner_for_removed_and_private_services(self, d
             )
         ]
 
+    @pytest.mark.parametrize('service_status', ['disabled', 'enabled'])
+    def test_info_banner_contains_publish_link_for_ccs_category(self, data_api_client, service_status):
```
TheDoubleK: Nice work on the test coverage, but I think we're missing a test that there is NO publish link for regular admins.

tests/app/main/views/test_services.py
```diff
@@ -239,6 +221,29 @@ def test_service_view_shows_info_banner_for_removed_and_private_services(self, d
...
+        self.user_role = 'admin-ccs-category'
```
TheDoubleK: We don't need to set this user_role per test as it's set for the class as `admin-ccs-category` at line 40, so you only need to set `self.user_role` for tests that need something other than Category.

app/templates/view_service.html
```diff
@@ -25,41 +25,82 @@
 
 {% block main_content %}
 
-  {% with messages = get_flashed_messages(with_categories=true) %}
-    {% if messages %}
+  {% block before_heading %}
+
+    {# banners #}
```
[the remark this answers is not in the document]
benvand: Yeah just a comment. It was all getting a bit hard to read. Happy to remove if you like. The renderer skips them entirely so they don't effect the html.

tests/app/main/views/test_services.py
benvand: Hmmm Let me look into this.

tests/app/main/views/test_services.py
benvand: Will change

benvand:
@TheDoubleK Apologies for the confusion. I've updated the screenshot with the links @laurenceberry  and me came up with

app/templates/view_service.html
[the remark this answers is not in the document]
TheDoubleK: No keep it, it's good. I should have spotted the #s to figure it's a comment, was being dense, sorry.

tests/app/main/views/test_services.py
benvand: Ah that was brought in by the rebase if I'm reading the diff right. Thought I'd checked that. I've checked and where necessary updated all the classes in that file in a new commit.

benvand:
@TheDoubleK That's all updated now. Splitting out the assertion for `find_audit_events.called` let me parametrize one of the tests to get rid of the (almost) duplicates.

PR #324, opened 2017-11-30
Enable admin manager to edit admin users

CrystalPea:
Trello ticket: https://trello.com/c/j1jx6KkX/154-allow-super-admin-edit-admin-roles
Prototype screenshots: https://docs.google.com/document/d/1MjvYx9aQWgcbryiaGrRLOPt_lVm8gIbqOH3vmkKx_fo/edit#heading=h.qdo4oz9trpdq

Screenshots of the real thing:

screen shot 2017-11-30 at 16 58 01

screen shot 2017-11-30 at 16 58 32

screen shot 2017-11-30 at 16 58 55

### DIFF
--- a/app/main/views/admin_manager.py
+++ b/app/main/views/admin_manager.py
@@ -5,9 +5,10 @@ from dmutils.email.user_account_email import send_user_account_email
 
 from ... import data_api_client
 from .. import main
-from ..forms import InviteAdminForm
-from ..auth import role_required
 
+from ..forms import InviteAdminForm
+from dmapiclient import HTTPError, APIError
+from ..auth import role_required
 
 @main.route('/admin-users', methods=['GET'])
 @role_required('admin-manager')
@@ -63,3 +64,15 @@ def invite_admin_user():
         form=form,
         errors=errors,
     ), 200 if not errors else 400
+
+
+
+@role_required('admin')
+@main.route('/admin-users/<string:admin_user_id>/edit', methods=['GET'])
+@role_required('admin-manager')
+def edit_admin_users(admin_user_id):
+    admin_user = data_api_client.get_user(admin_user_id)
+    return render_template(
+        "edit_admin_user.html",
+        admin_user=admin_user["users"],
+    )
--- a/app/main/views/admin_manager.py
+++ b/app/main/views/admin_manager.py
@@ -6,10 +6,11 @@ from dmutils.email.user_account_email import send_user_account_email
 from ... import data_api_client
 from .. import main
 
-from ..forms import InviteAdminForm
-from dmapiclient import HTTPError, APIError
+from ..forms import InviteAdminForm, EditAdminUserForm
+from dmapiclient import HTTPError
 from ..auth import role_required
-from ..forms import EditAdminUserForm
+from ..helpers.common_helpers import string_to_bool
+
 
 @main.route('/admin-users', methods=['GET'])
 @role_required('admin-manager')
@@ -67,22 +68,38 @@ def invite_admin_user():
     ), 200 if not errors else 400
 
 
-
-
-@main.route('/admin-users/<string:admin_user_id>/edit', methods=['GET', 'POST'])
-@role_required('admin-manager')
+@main.route("/admin-users/<string:admin_user_id>/edit", methods=["GET", "POST"])
+@role_required("admin-manager")
 def edit_admin_users(admin_user_id):
     admin_user = (data_api_client.get_user(admin_user_id))["users"]
-    edit_admin_user_form = EditAdminUserForm()
+    status_code = 200
+    edit_admin_user_form = EditAdminUserForm(
+        edit_admin_name=admin_user["name"],
+        edit_admin_permissions=admin_user["role"],
+        edit_admin_status=admin_user["active"])
 
-    for choice in edit_admin_user_form.edit_admin_permissions.choices:
-        choice["input_value"] = admin_user["role"]
+    if edit_admin_user_form.validate_on_submit():
+        try:
+            edited_admin_name = request.form.get("edit_admin_name")
+            edited_admin_permissions = request.form.get("edit_admin_permissions")
+            edited_admin_status = string_to_bool(request.form.get("edit_admin_status"))
 
-    for choice in edit_admin_user_form.edit_admin_status.choices:
-        choice["input_value"] = str(admin_user["active"])
+            data_api_client.update_user(
+                admin_user_id,
+                name=edited_admin_name,
+                role=edited_admin_permissions,
+                active=edited_admin_status
+            )
+            flash("{} has been updated.".format(admin_user["emailAddress"]), "message")
+            return redirect(url_for('.manage_admin_users'))
+        except HTTPError as e:
+            status_code = 400
+            raise e
+    elif edit_admin_user_form.edit_admin_name.errors:
+        status_code = 400
 
     return render_template(
         "edit_admin_user.html",
         admin_user=admin_user,
         edit_admin_user_form=edit_admin_user_form,
-    )
+    ), status_code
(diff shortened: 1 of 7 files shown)


constancecerf:
@karlchillmaid All good?

karlchillmaid:
@constancecerf 
Yep. Looks good.

app/main/helpers/common_helpers.py
```diff
@@ -0,0 +1,5 @@
+def string_to_bool(string):
```
katstevens: I don't think we need this. Python has a built in `strtobool` function:

```
>>> from distutils.util import strtobool
>>> val = "False"
>>> bool(val)
True
>>> strtobool(val)
0
>>> bool(strtobool(val))
False
```

app/main/views/admin_manager.py
```diff
@@ -63,3 +66,40 @@ def invite_admin_user():
         form=form,
         errors=errors,
     ), 200 if not errors else 400
+
+
+@main.route("/admin-users/<string:admin_user_id>/edit", methods=["GET", "POST"])
+@role_required("admin-manager")
+def edit_admin_user(admin_user_id):
+    admin_user = (data_api_client.get_user(admin_user_id))["users"]
```
katstevens: Do we need parentheses around the `data_api_client` call?

app/templates/edit_admin_user.html
```diff
@@ -0,0 +1,97 @@
...
+  <form method="post">
+    <div class="grid-row">
+      <div class="column-two-thirds">
+        <input id="csrf_token" name="csrf_token" type="hidden" value="{{ csrf_token() }}">
+        {%
+          with
+          question = "Name",
+          name = edit_admin_user_form.edit_admin_name.name,
+          value = edit_admin_user_form.edit_admin_name.data,
+          error = ""
```
katstevens: Just to check my understanding, is this error blank because we're only displaying the error in the header, not next to the form field?

tests/app/main/views/test_admin_manager.py
```diff
@@ -1,6 +1,7 @@
+from __future__ import unicode_literals
```
katstevens: We don't need this `__future__` import.

tests/app/main/views/test_admin_manager.py
```diff
@@ -110,6 +117,25 @@ def test_should_list_alphabetically_with_all_suspended_users_below_active_users(
...
+        document = html.fromstring(response.get_data(as_text=True))
+
+        links = document.xpath("//td[@class='summary-item-field-with-action']//a")
+
+        assert links[0].values()[0] == "/admin/admin-users/9089/edit"
+        assert links[1].values()[0] == "/admin/admin-users/9088/edit"
+        assert links[2].values()[0] == "/admin/admin-users/9087/edit"
+        assert links[3].values()[0] == "/admin/admin-users/9092/edit"
+        assert links[4].values()[0] == "/admin/admin-users/9090/edit"
+        assert links[5].values()[0] == "/admin/admin-users/9091/edit"
```
katstevens: With a bit of xpath magic this could be rewritten as:

```
        links = document.xpath("//td[@class='summary-item-field-with-action']//a/@href")

        assert links == [
            "/admin/admin-users/9089/edit",
            "/admin/admin-users/9088/edit",
            "/admin/admin-users/9087/edit",
            "/admin/admin-users/9092/edit",
            "/admin/admin-users/9090/edit",
            "/admin/admin-users/9091/edit"
        ]
```

tests/app/main/views/test_admin_manager.py
```diff
@@ -204,3 +230,143 @@ def test_successful_post_flashes(self, data_api_client, send_user_account_email)
...
+@mock.patch("app.main.views.admin_manager.data_api_client")
+class TestAdminManagerEditsAdminUsers(LoggedInApplicationTest):
+
+    admin_user_to_edit = {
+        "users": {
+            "active": True,
+            "emailAddress": "reality.auditor@digital.cabinet-office.gov.uk",
+            "id": 2345,
+            "locked": False,
+            "name": "Auditor of Reality",
```
katstevens: 🙀

tests/app/main/views/test_admin_manager.py
```diff
@@ -204,3 +230,143 @@ def test_successful_post_flashes(self, data_api_client, send_user_account_email)
...
+
+        response = self.client.get("/admin/admin-users/2345/edit")
+        assert response.status_code == 200
+
+        document = html.fromstring(response.get_data(as_text=True))
+
+        assert document.xpath('//h1')[0].text.strip() == "reality.auditor@digital.cabinet-office.gov.uk"
+        assert document.xpath('//input[@class="button-save"]')[0].value.strip() == "Update user"
+
+    def test_should_correctly_load_and_prefill_edit_name_question(self, data_api_client):
```
katstevens: In general I'd avoid test names that say `should_correctly_do_something` unless it's 100% obvious what the correct behaviour is. 

I like the format `test_thing_x_does_action_y_under_condition_z` with a bit of common sense and flexibility in ordering/grammar (usually so it fits on one line!). If condition z is the default, that probably doesn't need to go in the test name, for example. 

So here it would be something like `test_edit_admin_user_form_prefills_edit_name_with_user_name` (and for the test below, `test_edit_admin_user_form_prefils_permission_for_category_user` would do).

tests/app/main/views/test_admin_manager.py
```diff
@@ -204,3 +230,143 @@ def test_successful_post_flashes(self, data_api_client, send_user_account_email)
...
+                "edit_admin_status": "False"
+            }
+        )
+        assert response1.status_code == 302
+        assert_flashes(self, "reality.auditor@digital.cabinet-office.gov.uk has been updated", "message")
+        assert data_api_client.update_user.call_args_list == [mock.call(
+            "2345", name="Lady Myria Lejean", role="admin", active=False
+        )]
+
+        response2 = self.client.get(response1.location)
```
katstevens: Can we assert the `location` we've redirected to?

app/main/helpers/common_helpers.py
CrystalPea: Sweet, thank you for that, I updated the code accordingly :)

app/main/views/admin_manager.py
CrystalPea: I checked now, we don't :). I will get rid of them :D

app/templates/edit_admin_user.html
CrystalPea: I think so, but mostly I copied this question structure from add_buyer_email_domain.html template.

tests/app/main/views/test_admin_manager.py
CrystalPea: ok :)

tests/app/main/views/test_admin_manager.py
CrystalPea: Ahh, neat 💖

tests/app/main/views/test_admin_manager.py
CrystalPea: 💀

tests/app/main/views/test_admin_manager.py
CrystalPea: Good idea, I've done that now :)

tests/app/main/views/test_admin_manager.py
```diff
@@ -204,3 +231,154 @@ def test_successful_post_flashes(self, data_api_client, send_user_account_email)
...
+        redirect_text = html.fromstring(response1.get_data(as_text=True)).text_content().strip()
+        expected_url = '/admin-users.'
+
+        assert expected_url in redirect_text
```
katstevens: Ah - I meant asserting the location directly from the response itself, `response1.location == 'http://localhost/admin/admin-users`.

I don't think we need to get `response2` separately or assert the update message, as we're already asserting the flash message content above.

PR #325, opened 2017-12-04
Fix a bug that happened when trying to suspend an admin user

CrystalPea:
Introduced on PREVIEW only with this story: https://trello.com/c/j1jx6KkX/154-allow-super-admin-edit-admin-roles


TheDoubleK:
This seems a weird thing to have to do and I don't see how casting a bool to a bool can make any difference?  The docs say that `strtobool` will return true or false (unless it raises an error): http://nullege.com/codes/search/distutils.util.strtobool

So how does casting to a bool make any difference?

katstevens:
`strtobool` returns 0 or 1 - while it evaluates OK in the view (and on the app tests), it didn't validate when passed to the API client.

PR #326, opened 2017-12-05
Change from flake8-putty to flake8-per-file-ignores.

samuelhwilliams:
This allows us to support newer Python 3 syntax including f-strings and
type annotations.

Report after changing extensions:
```
samuelwilliams:~/git/digitalmarketplace-admin-frontend [master]$ make test-flake8
[ -z $VIRTUAL_ENV ] && [ ! -d venv ] && virtualenv -p python3 venv || true
/Users/samuelwilliams/git/digitalmarketplace-admin-frontend/venv/bin/flake8 .
./app/main/views/__init__.py:0:1: X100 Superfluous per-file-ignores for C901
./app/main/views/admin_manager.py:0:1: X100 Superfluous per-file-ignores for C901
./app/main/views/agreements.py:0:1: X100 Superfluous per-file-ignores for C901
./app/main/views/buyers.py:0:1: X100 Superfluous per-file-ignores for C901
./app/main/views/buyers.py:19:5: E722 do not use bare except'
./app/main/views/communications.py:0:1: X100 Superfluous per-file-ignores for C901
./app/main/views/service_updates.py:0:1: X100 Superfluous per-file-ignores for C901
./app/main/views/services.py:0:1: X100 Superfluous per-file-ignores for C901
./app/main/views/stats.py:0:1: X100 Superfluous per-file-ignores for C901
./app/main/views/suppliers.py:0:1: X100 Superfluous per-file-ignores for C901
./app/main/views/users.py:0:1: X100 Superfluous per-file-ignores for C901
make: *** [test-flake8] Error 1```

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -3,13 +3,13 @@
 # D203: 1 blank line required before class docstring
 # F401: 'identifier' imported but unused
 # E402: module level import not at top of file
-# C901: too complex
 # W503: line break before binary operator
 exclude = venv,venv3,__pycache__,node_modules,bower_components,app/content
 ignore = D203,W503
 max-complexity = 24
 max-line-length = 120
-putty-ignore =
-    **/__init__.py : +F401
-    app/__init__.py, app/main/__init__.py, app/status/__init__.py : +E402
-    app/main/views/*.py : +C901
+per-file-ignores =
+    **/__init__.py : F401
+    app/__init__.py : E402
+    app/main/__init__.py : E402
+    app/status/__init__.py : E402
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -2,8 +2,8 @@
 coverage==3.7.1
 cssselect==0.9.1
 freezegun==0.3.7
-flake8==2.6.2
-flake8-putty==0.4.0
+flake8==3.5.0
+flake8-per-file-ignores==0.4.0
 mock==2.0.0
 pytest==3.2.3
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: app/main/views/buyers.py [hunk 1]
```diff
@@ -16,7 +16,7 @@ def find_buyer_by_brief_id():
     try:
         brief = data_api_client.get_brief(brief_id).get('briefs')
 
-    except:
+    except HTTPError:
```
katstevens: Looks like the broad `try...except` was catching `None` errors on `data_api_client.get_brief(brief_id).get('briefs')` as well. I think `except (HttpError, AttributeError)` would do the trick.

samuelhwilliams: Hmm, what would be `None` in that instance? I think get_brief only returns either a valid brief or triggers a 404, so I don't think I can see what would return that. Can you fill me in please? :)

benvand: Hey yeah weirdly I had a look into this exact line of code yesterday for Kev and Peas PR and the try except has been changed there 😁 Once this is rebased the Try except is a `HTTPError` and this is no longer a problem.
See: https://github.com/alphagov/digitalmarketplace-admin-frontend/pull/331#discussion_r155479224

benvand:
@samuelhwilliams rebased, resolved conflicts and ran locally just to test my theory and it worked so pushed up. 👍

PR #327, opened 2017-12-05
176 Split the supplier csv into 2 (Active and All interested)

TheDoubleK:
For this ticket: https://trello.com/c/ci1yeRFZ/176-split-the-supplier-csv-into-2-active-and-all-interested

**NOTE** I've left the breadcrumb as one step away from the admin homepage, as that's where it will be after the home page reorg ticket (the intermediate page at `/admin/users/download` is going to go away altogether).

## New page ##

screen shot 2017-12-05 at 15 23 07

### DIFF
--- a/tests/app/main/views/test_users.py
+++ b/tests/app/main/views/test_users.py
@@ -292,7 +292,8 @@ class TestUsersExport(LoggedInApplicationTest):
             data_api_client.get_framework.side_effect = HTTPError(mock.Mock(status_code=404))
 
         return self.client.get(
-            '/admin/frameworks/<_valid_framework/users/download{}'.format(
+            '/admin/frameworks/{}/users/download{}'.format(
+                self._valid_framework['slug'],
                 '?on_framework_only=True' if only_on_fwk else ''
             ),
             data={'framework_slug': framework_slug}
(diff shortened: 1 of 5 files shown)


tests/app/main/views/test_users.py
```diff
@@ -337,6 +340,39 @@ def test_download_csv(self, data_api_client):
...
+            }
+        ]
+
+        response = self._return_user_export_response(data_api_client, framework, users, only_on_fwk=True)
+        assert response.status_code == 200
+        rows = [line.split(",") for line in response.get_data(as_text=True).splitlines()]
+
+        assert len(rows) == 2
+        assert rows[0] == ["email address", "user_name", "supplier_id"]
+        assert rows[1] == ["test.user@sme2.com", "Test User 2", "2"]
```
katstevens: This test looks fine, but the two above it seem like they could be combined? Or if we want to keep the filetype/name test separate from the content test, could we have the assertions on `test_download_csv` match those on the new test, so it's clear what the differences are between it and the `only_on_fwk` case?

TheDoubleK:
Yep good point - I've dropped the separate filename test and wrapped it in with the assertions for the with/without framework_only flag.  And have made those tests follow the same pattern to make it clearer what we expect to be the same/different in each case.

I had to fix merge conflicts three times over in that test file to bring it into line with the framework manager role stuff 😞

PR #329, opened 2017-12-06
Move find a user by email field to its own page

CrystalPea:
**All changes here are included in: https://github.com/alphagov/digitalmarketplace-admin-frontend/pull/331 - so if you feel like a bigger review then head straight over there!**

Trello ticket: https://trello.com/c/GfqT3M7t/166-move-find-a-user-by-email-field-to-its-own-page

 - Also refactor some tests

### Find a user page:
screen shot 2017-12-06 at 12 22 37

---
### Relevant section on the homepage:
screen shot 2017-12-06 at 12 22 52

PR #331, opened 2017-12-06
Move find buyer by opportunity field to its own page

TheDoubleK:
**This follows the commits here: https://github.com/alphagov/digitalmarketplace-admin-frontend/pull/329**
**Only the final two commits here are new since then**

For this ticket: https://trello.com/c/tuG2TN4H/165-move-find-buyer-by-opportunity-field-to-its-own-page

### New link on home page:
screen shot 2017-12-06 at 16 07 04

### Updated "find a buyer" page:
screen shot 2017-12-06 at 16 07 29

### DIFF
--- a/app/main/views/buyers.py
+++ b/app/main/views/buyers.py
@@ -12,26 +12,21 @@ from ... import data_api_client
 @role_required('admin', 'admin-ccs-category')
 def find_buyer_by_brief_id():
     brief_id = request.args.get('brief_id')
+    brief = None
+    status_code = 200
 
     try:
-        brief = data_api_client.get_brief(brief_id).get('briefs')
-
+        brief = data_api_client.get_brief(brief_id)
     except:
-        flash('no_brief', 'error')
-        return render_template(
-            "view_buyers.html",
-            users=list(),
-            brief_id=brief_id
-        ), 404
+        if not brief and 'brief_id' in request.args:
+            flash('no_brief', 'error')
+            status_code = 404
 
-    users = brief.get('users')
-    title = brief.get('title')
     return render_template(
         "view_buyers.html",
-        users=users,
-        title=title,
+        brief=brief.get('briefs') if brief else None,
         brief_id=brief_id
-    )
+    ), status_code
 
 
 @main.route('/buyers/add-buyer-domains', methods=['GET', 'POST'])
--- a/app/main/views/buyers.py
+++ b/app/main/views/buyers.py
@@ -17,8 +17,8 @@ def find_buyer_by_brief_id():
 
     try:
         brief = data_api_client.get_brief(brief_id)
-    except:
-        if not brief and 'brief_id' in request.args:
+    except HTTPError:
+        if 'brief_id' in request.args:
             flash('no_brief', 'error')
             status_code = 404
 
--- a/app/main/views/users.py
+++ b/app/main/views/users.py
@@ -19,10 +19,9 @@ CLOSED_BRIEF_STATUSES = ['closed', 'awarded', 'cancelled', 'unsuccessful']
 def find_user_by_email_address():
     template = "view_users.html"
     users = None
-
     email_address = request.args.get("email_address", None)
     if email_address:
-        users = data_api_client.get_user(email_address=request.args.get("email_address"))
+        users = data_api_client.get_user(email_address=email_address)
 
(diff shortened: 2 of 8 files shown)


app/main/views/buyers.py
```diff
@@ -12,26 +12,21 @@
...
 def find_buyer_by_brief_id():
     brief_id = request.args.get('brief_id')
+    brief = None
+    status_code = 200
 
     try:
-        brief = data_api_client.get_brief(brief_id).get('briefs')
-
+        brief = data_api_client.get_brief(brief_id)
     except:
```
benvand: We can trim down this try except to a except `HTTPError` looks like that's what we do elsewhere. This is the only instance of such a broad try except and looks like it was probably just an oversight, especially seeing as theres no chance of a `KeyError` now we're not operating on a dictionary (not that there was before).

app/main/views/buyers.py
```diff
@@ -12,26 +12,21 @@
...
-        flash('no_brief', 'error')
-        return render_template(
-            "view_buyers.html",
-            users=list(),
-            brief_id=brief_id
-        ), 404
+        if not brief and 'brief_id' in request.args:
```
benvand: Because the brief assignment will never fully execute if we are at this point (ie if there is an exception in getting and assigning the brief) brief will always be None here so we don't need to check `if not brief`

app/main/views/buyers.py
```diff
@@ -12,26 +12,21 @@
...
+            flash('no_brief', 'error')
+            status_code = 404
 
-    users = brief.get('users')
-    title = brief.get('title')
     return render_template(
         "view_buyers.html",
-        users=users,
-        title=title,
+        brief=brief.get('briefs') if brief else None,
```
benvand: Like this 👍  we could have defaulted `brief` to an empty dict at the start of the method then instead of `brief.get('briefs') if brief else None` here used just `brief.get('briefs', None)` but I think this is more readable and signals our intent better. 
_If we haven't managed to find a brief then pass back `None` to the template_

app/main/views/users.py
```diff
@@ -24,19 +24,19 @@ def find_user_by_email_address():
     if email_address:
         users = data_api_client.get_user(email_address=request.args.get("email_address"))
```
benvand: Slightly outside the changed code here but we assign the `request.args email address` or `None` to the variable `email_address` earlier in the method then repeatedly reference it from the `request.args` anyway. Would be good to make this consistent.

app/main/views/buyers.py
TheDoubleK: Totally an oversight, will fix.

app/main/views/buyers.py
TheDoubleK: Had spotted this too but didn't fix.  Will do.

app/main/views/users.py
```diff
@@ -24,19 +24,19 @@ def find_user_by_email_address():
...
             template,
             users=list(),
             email_address=None
         ), 404
+    else:
+        return render_template(
+            template,
+            users=[users['users']] if users else list(),
+            email_address=request.args.get("email_address")
+        )
```
benvand: It wouls be awesome if we could make this view like the other one. I much prefer that style of having a single return statement if we're using a single template because we always want to return the same things.

```
@main.route('/users', methods=['GET'])
@role_required('admin', 'admin-ccs-category')
def find_user_by_email_address():
    users = None
    email_address = None
    response_code = 200

    if "email_address" in request.args:
        email_address = request.args.get("email_address")
        if email_address:
            users = data_api_client.get_user(email_address=email_address)
        if not users:
            flash('no_users', 'error')
            response_code = 404

    return render_template(
        "view_users.html",
        users=[users['users']] if users else list(),
        email_address=email_address
    ), reponse_code
```

Feels like setting defaults for what we want to operate on at the top, keeping the logic in the middle, then demonstrating what we want to return really reduces clutter and makes the whole thing more readable. It breaks with the convention of failing/ dropping out early but easier supports DRY.

Totally happy with leaving it up to you two. This isn't a refactor PR 💯

TheDoubleK: Nice one - yes we've done this now.

benvand:
Awesome let me know when it's merged and I can fun the functional tests I have prepped against it 💯

PR #333, opened 2017-12-07
Move 'find supplier and services' fields to their own page

CrystalPea:
Trello ticket: https://trello.com/c/E7fLF72t/167-move-find-supplier-and-services-fields-to-their-own-page

Associated smoketest changes in a PR here **MERGE THESE AT THE SAME TIME AS THIS ONE**:
 - [ ] https://github.com/alphagov/digitalmarketplace-functional-tests/pull/436

### Admin homepage with a link instead of the forms:
screen shot 2017-12-07 at 16 18 38

---

### New 'Find suppliers and services' page:
screen shot 2017-12-07 at 16 19 34

### DIFF
--- a/app/main/views/search.py
+++ b/app/main/views/search.py
@@ -1,0 +1,14 @@
+from flask import render_template, request, redirect, url_for, flash
+from flask_login import current_user
+
+from ... import data_api_client
+from .. import main
+
+from dmapiclient import HTTPError
+from ..auth import role_required
+
+
+@main.route('/find-suppliers-and-services', methods=['GET'])
+@role_required('admin', 'admin-ccs-category', 'admin-ccs-sourcing', 'admin-framework-manager')
+def find_suppliers_and_services():
+    return render_template("find_suppliers_and_services.html")
--- a/app/main/views/search.py
+++ b/app/main/views/search.py
@@ -1,10 +1,5 @@
-from flask import render_template, request, redirect, url_for, flash
-from flask_login import current_user
-
-from ... import data_api_client
+from flask import render_template
 from .. import main
-
-from dmapiclient import HTTPError
 from ..auth import role_required
 
 
--- a/app/main/views/services.py
+++ b/app/main/views/services.py
@@ -31,7 +31,7 @@ def index():
 @role_required('admin', 'admin-ccs-category', 'admin-framework-manager')
 def find():
     if request.args.get("service_id") is None:
-        return render_template("index.html"), 404
+        return render_template("find_suppliers_and_services.html"), 404
     return redirect(
         url_for(".view", service_id=request.args.get("service_id")))
 
@@ -43,11 +43,11 @@ def view(service_id):
         service = data_api_client.get_service(service_id)
         if service is None:
             flash({'no_service': service_id}, 'error')
-            return redirect(url_for('.index'))
+            return redirect(url_for('.find_suppliers_and_services'))
         service_data = service['services']
     except HTTPError:
         flash({'api_error': service_id}, 'error')
-        return redirect(url_for('.index'))
+        return redirect(url_for('.find_suppliers_and_services'))
 
     removed_by = removed_at = None
     if service_data['status'] != 'published':
--- a/app/main/views/services.py
+++ b/app/main/views/services.py
@@ -111,7 +111,7 @@ def update_service_status(service_id):
 @main.route('/services/<service_id>/edit/<section_id>', methods=['GET'])
 @main.route('/services/<service_id>/edit/<section_id>/<question_slug>', methods=['GET'])
 @role_required('admin-ccs-category')
-def edit(service_id, section_id, question_slug=None):
+def edit_service(service_id, section_id, question_slug=None):
     service_data = data_api_client.get_service(service_id)['services']
 
     content = content_loader.get_manifest(service_data['frameworkSlug'], 'edit_service_as_admin').filter(service_data)
--- a/app/main/views/services.py
+++ b/app/main/views/services.py
@@ -33,12 +33,12 @@ def find_service():
     if request.args.get("service_id") is None:
         return render_template("find_suppliers_and_services.html"), 404
     return redirect(
-        url_for(".view", service_id=request.args.get("service_id")))
+        url_for(".view_service", service_id=request.args.get("service_id")))
 
 
 @main.route('/services/<service_id>', methods=['GET'])
 @role_required('admin', 'admin-ccs-category', 'admin-framework-manager')
-def view(service_id):
+def view_service(service_id):
     try:
         service = data_api_client.get_service(service_id)
         if service is None:
@@ -90,7 +90,7 @@ def update_service_status(service_id):
         backend_status = translate_frontend_to_api[frontend_status]
     else:
         flash({'bad_status': frontend_status}, 'error')
-        return redirect(url_for('.view', service_id=service_id))
+        return redirect(url_for('.view_service', service_id=service_id))
 
     try:
         data_api_client.update_service_status(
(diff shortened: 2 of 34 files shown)


tests/app/main/views/test_search.py
```diff
@@ -0,0 +1,60 @@
...
+        )
+        assert title_1_exists == form_1_exists, (
+            "Role {} {} see the Find a supplier by name form".format(role, "can not" if form_1_exists else "can")
+        )
+        assert title_2_exists == form_2_exists, (
+            "Role {} {} see the Find a supplier by DUNS number form".format(role, "can not" if form_2_exists else "can")
+        )
+        assert title_3_exists == form_3_exists, (
+            "Role {} {} see the Find a service by service ID form".format(role, "can not" if form_3_exists else "can")
+        )
```
katstevens: These assertions feel a bit hard to get my head around. We're not really asserting the existence of the forms, just the headers. Could we instead get _all_ the `question-heading` spans then assert the content for each role?

```
question_headings = [heading.strip() for heading in document.xpath('//span[@class="question-heading"]')]
assert ("Find a supplier by name" in question_headings) == form_1_exists, "Role {} {} see the blah blah blah..."
```

tests/app/main/views/test_search.py
```diff
@@ -0,0 +1,67 @@
...
+        response = self.client.get('/admin/find-suppliers-and-services')
+        page_html = response.get_data(as_text=True)
+        document = html.fromstring(page_html)
+        header = document.xpath(
+            '//header[@class="page-heading page-heading-without-breadcrumb"]//h1/text()')[0].strip()
+
+        assert response.status_code == 200
+        assert header == expected_header
+
+    @pytest.mark.parametrize("role,form_1_exists,form_2_exists,form_3_exists", [
```
katstevens: It'd be good to have these as `supplier_name_form`, `supplier_duns_form`, `service_id_form` rather than form1, form2 etc.

app/templates/find_suppliers_and_services.html
```diff
@@ -0,0 +1,115 @@
...
+
+  <div class="grid-row">
+    <div class="column-two-thirds">
+
+      <form action="{{ url_for('.find_suppliers') }}" method="get" class="question">
+        {%
+          with
+          question = "Find a supplier by name",
+          name = "supplier_name_prefix",
+          hint = "You don't have to put in the full name to get a result."
```
katstevens: Should there be a curly apostrophe for `don't`?

app/templates/find_suppliers_and_services.html
```diff
@@ -0,0 +1,115 @@
...
+{% elif current_user.role == "admin-framework-manager" %}
+  {% set page_title = "View suppliers and services" %}
+{% endif %}
+
+{% block breadcrumb %}
+  {%
+      with items = [
+          {
+              "link": url_for('.index'),
+              "label": page_title
```
katstevens: I'm confused - is this the right breadcrumb text for `.index`?

app/templates/find_suppliers_and_services.html
```diff
@@ -0,0 +1,115 @@
...
+          with
+          type = "save",
+          label = "Search"
+        %}
+          {% include "toolkit/button.html" %}
+        {% endwith %}
+      </form>
+
+      {% if current_user.has_any_role('admin', 'admin-ccs-category', 'admin-framework-manager') %}
+        <form action="{{ url_for('.find') }}" method="get" class="question">
```
katstevens: I would dearly love `.find` to be renamed to something more specific, but probably not something to do in this story...

app/templates/find_suppliers_and_services.html
```diff
@@ -0,0 +1,115 @@
...
+          {%
+            with
+            question = "Find a service by service ID",
+            name = "service_id",
+            hint = "For example 123456789012345"
+          %}
+            {% include "toolkit/forms/textbox.html" %}
+          {% endwith %}
+          {%
+            with
```
benvand: It would be good to distinguish these `Search` buttons with individual names.
Something like:
```
name = "find_supplier_by_name_search"
name = "find_supplier_by_duns_number_search"
name = "find_service_search"
```
might work?

app/main/views/search.py
```diff
@@ -0,0 +1,9 @@
+from flask import render_template
+from .. import main
+from ..auth import role_required
+
+
+@main.route('/find-suppliers-and-services', methods=['GET'])
```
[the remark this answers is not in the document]
TheDoubleK: We agreed this at kick-off with Laurence and Tobi - we didn't just freestyle it 😄

app/templates/find_suppliers_and_services.html
TheDoubleK: No - this should be `Admin home` still.

app/templates/find_suppliers_and_services.html
TheDoubleK: :eyes: Yes - definitely.

app/templates/find_suppliers_and_services.html
TheDoubleK: That was my feeling too.

tests/app/main/views/test_search.py
```diff
@@ -0,0 +1,67 @@
...
+    ):
+        self.user_role = role
+        response = self.client.get('/admin/find-suppliers-and-services')
+        document = html.fromstring(response.get_data(as_text=True))
+        question_headings = [heading.strip() for heading in document.xpath('//span[@class="question-heading"]/text()')]
+
+        # We test that we can find the relevant question headings as a proxy for finding the forms themselves
+        assert len(question_headings) == (
+            supplier_name_form_should_exist + supplier_duns_form_should_exist + service_id_form_should_exist
+        )
```
katstevens: 👍 👍 👍 Love it

app/templates/errors/500.html
```diff
@@ -1,5 +1,20 @@
...
+  {%
+      with items = [
+          {
+              "link": url_for('main.index'),
+              "label": "Admin home"
+          }
+      ]
+  %}
+    {% include "toolkit/breadcrumb.html" %}
+  {% endwith %}
```
katstevens: We repeat this default breadcrumb stuff in a lot of places in the admin - is it possible to put it in the `_base_page.html` block? Or is visibility per page preferable?

app/main/views/services.py
```diff
@@ -133,6 +133,60 @@ def edit_service(service_id, section_id, question_slug=None):
     )
 
 
+@main.route('/services/<service_id>/edit/<section_id>', methods=['POST'])
+@main.route('/services/<service_id>/edit/<section_id>/<question_slug>', methods=['POST'])
+@role_required('admin-ccs-category')
+def update_service(service_id, section_id, question_slug=None):
```
katstevens: Where do we use this route? I just changed the name of it to `update_service_apocalypse` and nothing broke 🙀

TheDoubleK: The GET route posts to itself.  So the name of the method doesn't matter, but the route is used.

app/templates/errors/500.html
TheDoubleK: But then the index page would get a breadcrumb to itself, which we probably don't want?

Could override it there with an empty block I guess?  

I don't mind having each template responsible for its own breadcrumb - it just so happens that in admin most things are one page away from the home page, but I don't think that necessarily means we should make it generic.

app/main/views/services.py
TheDoubleK: There are tests for it, but they call it by URL rather than method name.

tests/app/main/views/test_search.py
TheDoubleK: Ta - I was quietly pleased with this myself.

app/main/views/services.py
katstevens: Gotcha 👍 I guess that's an issue with having the form action buried in ungreppable toolkit templates...

app/templates/errors/500.html
katstevens: Fair enough. If we did decide to change it, it should probably go in a separate PR anyway.

PR #334, opened 2017-12-08
Remove link-button from admin app

laurenceberry:
There is no need for this link-button style, this commit removes the link-button (and the javascript that makes it act like a button) in use from the admin app.


MuriloDalRi:
Looks good, just needs a rebase before merging 👍

PR #337, opened 2017-12-08
Make buyer download filename nicer

TheDoubleK:
I've kept the timestamp as that seems important information, but I think the filename is a bit more human-friendly this way.

Old filename example:
`buyers_20160805T160000.csv`
New filename example:
`all-digital-marketplace-buyers-on-2016-08-05_at_16-00-00.csv`

### DIFF
--- a/app/main/views/users.py
+++ b/app/main/views/users.py
@@ -137,7 +137,7 @@ def download_buyers():
         csv_generator.iter_csv(rows_iter),
         mimetype='text/csv',
         headers={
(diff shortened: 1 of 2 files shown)


samuelhwilliams:
This is a friendly human readable format but it doesn't seem consistent with other files we generate, which have the date/time at the start of the filename. It's perhaps a bit generous to call it a convention, but should we stick to that instead?

If/if not, the underscores could probably still be dashes. I get that it's meant to separate the date and the time, but it looks a bit odd considering we have dashes through all the rest of it.

PR #338, opened 2017-12-11
Bring in API client 14.0.1

idavidmcdonald:
Fixes bug related to API responses no longer always containing
the `links` key


samuelhwilliams:
Unless we need to update Werkzeug to use the new apiclient, probably better to drop that requirement bump.

samuelhwilliams:
Buttons are close together.

PR #350, opened 2018-01-16
Use toolkit form macros instead of app form macros

idavidmcdonald:
We have a central set of form macros in our toolkit. This app
had slightly differing macros. In principle, there is no reason
why we should be doing things different to the toolkit so the aim
was to swap straight to them. Most of the differences between
the macros in this app and the toolkit made no difference for
example

- `hint_underneath` param for `radios` macro. It does not matter
if the value is `None` or `False` as both will evaluate falsey
and our template ends up not minding. Therefore there is no reason
to keep the divergence of `question_content.hint_underneath or False`

- the `pricing` macro was passing `hidden` and `id` variables into
our html template, however these variables are not actually used
in the template so there is no reason to keep them

- we don't have a `percentage` macro in the toolkit because we
don't have any questions of type `percentage`. I assume this was
leftover from a long time ago. Instead questions that ask for
a percentage value are of type `number` with the unit field `%`.
Therefore, no reason to keep this macro as it would never be used.

- all the macros were using `hint=question_content.hint or ''`. If
`question_content.hint` is not set than the value from the content
loader is `None` which is falsey like `''` and makes no difference
to our templates. Therefore we do not need the empty string option
which is different to the toolkit.

There were however some differences spotted but I believe it is OK
to accept these changes (will confirm this with a designer/PM).

- The `multiquestion` macro diverges reasonabally. The one in the
toolkit outputs `question_content.question_advice` if it exists and
`question_content.hint` if it exists. This means that we may be
showing additional context about the question at the top of
multiquestions

- The toolkit macros show additional content. For example the
`uploads` macro passes the following to the template:
`question_advice=kwargs.get("question_advice", question_content.question_advice`.
Before we were not passing any question_advice to the uploads
 template, even if `question_content.question_advice` did exist.
There are several similar examples of this across the new
toolkit macros but it should be fine as they are showing
additional context where it exists.

Here are two examples to highlight some of the before and after differences. Note, these are not all of the locations where things have changed but should give you an idea of the type of changes.

Before:
image

After:
image

Before:
image

After:
image


idavidmcdonald:
Note, Laurence is happy with this

PR #352, opened 2018-01-26
Update requirements and remove six

TheDoubleK:
I was a bit worried that the admin app might be affected by the same issue with Werkzeug update as the supplier app (see https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/779).

So I tried updating and everything is fine here - tests pass and I've tried out the file upload pages locally and all seems well.

As there was nothing much to do with that I removed use of `six` from the app and tests, as we're all Python 3 these days.

### DIFF
--- a/tests/app/main/views/test_services.py
+++ b/tests/app/main/views/test_services.py
@@ -1525,7 +1525,7 @@ class TestServiceUpdates(LoggedInApplicationTest):
         # service status:
         "disabled",
         # expected message about the oldest unapproved edit:
-        "Changed on Wednesday 3 February 2010 at 10:11am",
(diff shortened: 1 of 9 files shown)


tests/app/main/views/test_services.py
```diff
@@ -1641,7 +1633,7 @@ def _mock_find_audit_events_side_effect(find_audit_events_api_response, implicit
         # service status:
         "enabled",
         # expected message about the oldest unapproved edit:
-        "Changed on Saturday 30 June 2012 at 9:01pm",
+        "Changed on Saturday 30 June 2012 at 9:01pm BST",
```
katstevens: Now with added timezone localisation woohoo 🎉

PR #357, opened 2018-02-13
 Revert back to eea/odfpy

samuelhwilliams:
Update utils to use eea/odfpy rather than our interim alphagov/odfpy
fork, which we created to allow us to patch a bug. Now that the bug has
been accepted into the main library as of release 1.3.6
(eea/odfpy#72) we no longer need to support own
our fork.

Also adds a comment when registering the external blueprint from 
dm-utils to ensure it is registered last.

### DIFF
--- a/tests/app/main/views/test_suppliers.py
+++ b/tests/app/main/views/test_suppliers.py
@@ -191,9 +191,9 @@ class TestSupplierUsersView(LoggedInApplicationTest):
         assert response.status_code == 200
         assert "Test User" in response.get_data(as_text=True)
         assert "test.user@sme.com" in response.get_data(as_text=True)
-        assert "10:33:53" in response.get_data(as_text=True)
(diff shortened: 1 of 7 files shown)


tests/app/main/views/test_suppliers.py
```diff
@@ -191,7 +191,7 @@ def test_should_show_user_details_on_page(self, data_api_client):
         assert response.status_code == 200
         assert "Test User" in response.get_data(as_text=True)
         assert "test.user@sme.com" in response.get_data(as_text=True)
-        assert "10:33:53" in response.get_data(as_text=True)
+        assert "09:33:53" in response.get_data(as_text=True)
```
galund: Is there a time zone in there now, if so can our assertion cover that being in there?

samuelhwilliams: There's no timezone added in this instance.

PR #363, opened 2018-02-28
Revert "Bump digitalmarketplace-utils dependency to 34.1.0"

risicle:
This reverts commit 1cd9cd765412ec53fe167ab9ba68e72a5188f7ac.

Because we don't *set* the default for DM_REQUEST_ID_HEADER, apiclient,
which turns out to rely on this value, isn't able to find the correct
header to use when forwarding the request. back this out until we've
updated/fixed either the utils or apiclient

Trello https://trello.com/c/dlbLmRDB/343-make-use-of-zipkin-headers-in-preference-to-request-id-pt-1

I'm going AFK right now so if someone wants to approve **and maybe merge** this that would be cool.


katstevens:
Looks fine to me as well - will merge once Travis has done its thing.

PR #364, opened 2018-03-02
request id handling: bump digitalmarketplace-utils dependency to 34.1.1

risicle:
...and re-freeze requirements.txt. this should bring in the new request id header handling, and actually work this time.

https://trello.com/c/dlbLmRDB/343-make-use-of-zipkin-headers-in-preference-to-request-id-pt-1

### DIFF
--- a/config.py
+++ b/config.py
@@ -38,7 +38,6 @@ class Config(object):
     DM_LOG_LEVEL = 'DEBUG'
     DM_PLAIN_TEXT_LOGS = False
     DM_APP_NAME = 'admin-frontend'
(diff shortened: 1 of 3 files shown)


config.py
```diff
@@ -38,7 +38,6 @@ class Config(object):
     DM_LOG_LEVEL = 'DEBUG'
     DM_PLAIN_TEXT_LOGS = False
     DM_APP_NAME = 'admin-frontend'
-    DM_DOWNSTREAM_REQUEST_ID_HEADER = 'X-Amz-Cf-Id'
```
galund: This isn't mentioned in your commit message, but I assume it's an intentional re-removal of the obsolete ting, or something like that.

PR #366, opened 2018-03-06
User research mailing list

MuriloDalRi:
https://trello.com/c/xyI1s1Qx/132-create-csv-for-user-research-participants-on-admin

### DIFF
--- a/app/main/views/users.py
+++ b/app/main/views/users.py
@@ -1,9 +1,9 @@
 from datetime import datetime
-from itertools import chain
 
 from dmutils import csv_generator
+from dmutils.config import convert_to_boolean
 from flask import abort, render_template, request, Response
-from flask_login import flash
+from flask_login import flash, current_user
 
 from .. import main
 from ..auth import role_required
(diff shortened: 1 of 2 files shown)


tests/app/main/views/test_users.py
```diff
@@ -296,14 +294,18 @@ def test_download_csv_for_all_framework_users(self, data_api_client):
...
         # All users returned from the API should appear in the CSV
         for index, user in enumerate(users):
-            assert sorted([str(val) for val in user.values()]) == sorted(rows[index + 1])
+            assert sorted(
+                [str(val) for key, val in user.items() if key in expected_headings]
+            ) == sorted(rows[index + 1])
 
     def test_download_csv_for_on_framework_only(self, data_api_client):
-        framework = self._valid_framework
+        # framework = self._valid_framework
```
benvand: Ah looks like this line can be deleted

MuriloDalRi: Oh yeah, I forgot to delete that. 👍

MuriloDalRi:
@benvand the buyer download seems to be working locally for me but the supplier is just giving me a list with no users. It might be something with my local DB, maybe try it and see if it works for you?

benvand:
@MuriloDalRi Nice! That's actually the correct behaviour because the new field defaults to false. 
If you run: 
```python application.py db upgrade
psql digitalmarketplace -c "update users set user_research_preference = ((coalesce(supplier_id, 0)::int)%2)::bool" ```

Then half(ish) of your suppliers will be set to `true` for `user_research_preference`

**Comment by MuriloDalRi** [comment]
You're right! 👍 
Just need to change that variable name and start merging things I guess? What was the general consensus on the name?

**Comment by benvand** [comment]
Yeah sounds good. We landed on 'user_research_opted_in' in the end because it conveyed that it was an opt-in flag and the subject.

To test that the new field name migration works locally you'll want to run either of these:

Automatically roll back the migration:
```
python application.py db downgrade 1090
```

or

Manually reset the db state to how we assume it was:
```
psql digitalmarketplace -c 'update alembic_version set version_num = 1090'
psql digitalmarketplace -c 'alter table users drop column user_research_preference'
```


Then you can run the migration with the changed name.

File: app/main/views/users.py [hunk 2]
```diff
@@ -90,19 +103,23 @@ def download_users(framework_slug):
...
+    if current_user.role == 'admin':
+        user_attributes = user_attributes + ('userResearchOptedIn',)
+
     users = data_api_client.find_users_iter(role="buyer")
 
     rows_iter = chain(
         (
             # header row
-            ("email address", "name"),
+            ("email address", "name", "user research opted in" if current_user.role == 'admin' else None),
```
benvand: So I didn't realise this but dropping in `None` here adds an empty column to the csv header apparently 😒

PR #367, opened 2018-03-12
Accessible text for Edit links

katstevens:
Trello: https://trello.com/c/sYJYKnK9/58-how-summary-tables-handle-their-links-means-pages-can-contain-multiple-links-with-duplicate-text

If there are multiple links on a page with the same text, it can be confusing for screen reader users. This PR adds context in hidden text spans for the following links:
- Editing an admin user (`admin-manager` role)
- Uploading/downloading a supplier agreement file (`admin-ccs-sourcing` role)
- Editing a supplier declaration (`admin-ccs-sourcing` role)
- Editing a supplier service (`admin-ccs-category` role)
- Viewing unapproved supplier service edits (`admin-ccs-category` role)

This should be reviewed by a content designer (just in case) as well as a dev.

Also fixes some single/double apostrophe clashes.

See previous PRs for Brief Responses and Supplier FEs: 
https://github.com/alphagov/digitalmarketplace-brief-responses-frontend/32
https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/804

### DIFF
--- a/app/templates/_view_suppliers_agreements.html
+++ b/app/templates/_view_suppliers_agreements.html
@@ -23,55 +23,55 @@
 
     {% call summary.row() %}
       {% call summary.field() %}
-       <a href="{{ url_for(".view_supplier_declaration", supplier_id=item.id, framework_slug="g-cloud-7") }}">Edit declaration</a>
+       <a href="{{ url_for('.view_supplier_declaration', supplier_id=item.id, framework_slug='g-cloud-7') }}">Edit <span class="visually-hidden">G-Cloud 7 </span>declaration</a>
       {% endcall %}
       {% call summary.field() %}
-        <a href="{{ url_for(".view_supplier_declaration", supplier_id=item.id, framework_slug="digital-outcomes-and-specialists") }}">Edit declaration</a>
+        <a href="{{ url_for('.view_supplier_declaration', supplier_id=item.id, framework_slug='digital-outcomes-and-specialists') }}">Edit <span class="visually-hidden">Digital Outcomes and Specialists </span>declaration</a>
       {% endcall %}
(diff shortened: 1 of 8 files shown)


katstevens:
Agreed that the whole `view_suppliers_agreements.html` is a bit odd. It's hard to navigate as a sighted user, I can't imagine how annoying it must be for a screen reader user:

view-suppliers-agreements-ugh

I will see if I can make the links a little less awful.

CrystalPea:
@katstevens oh wow, that is a one crowded page 😮. 
Maybe we will be able to get rid of G-Cloud-7 at some point at least? 🤔 

Anyway, thanks for taking time to improve the accessibility 💖.

katstevens:
The up and down arrows aren't part of the link and so it's difficult to sensibly put them in a `summary.link`, which also renders the table cell. Most of the other cells have `rowspan=3` which is also not supported by `summary.link`. I think either this table has to get a full redesign, or the messy links stay 😿

CrystalPea:
I guess they stay for now then and we can highlight it to @laurenceberry and to whoever else it would be good to highlight it to? What do you think? Also if you need help let me know :).

katstevens:
We should flag it to whoever is going to have to update the Admin for G-10 😸 The poor CCS Sourcing team will have to wade through 6 columns instead of 5.

CrystalPea:
👍

katstevens:
Have added a tech debt ticket - this may (should?) get prioritised to a mission team dealing with G10 next quarter: https://trello.com/c/T2mJhHMp/389-admin-framework-agreement-management-page-will-need-to-fit-in-g-cloud-10-somehow

PR #373, opened 2018-03-22
Flash messages in the Admin FE

katstevens:
Trello: https://trello.com/c/arqFIzNG/301-flash-messages-admin-app

- Moves flash message content out of templates and into Python code
- Replaces `notification-banner.html` toolkit template with `flash_messages.html`
- Replaces a destructive action (for removing countersigned agreements) flash message with a conditional query param flag, following the pattern for removing supplier services from a framework
- Refactors `assert_flashes` helper function to be a method on the `BaseApplicationTest` class, as per other apps. The method can now assert multiple flash messages on the same page (eg uploaded communications/clarifications).

I've split out the different features into their own commits for ease of review - I can squash later.

### DIFF
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -6,6 +6,7 @@ from dateutil.parser import parse as parse_date
 from dmapiclient import HTTPError, APIError
 from dmapiclient.audit import AuditTypes
 from dmutils import s3
+from dmutils.config import convert_to_boolean
 from dmutils.documents import (
     AGREEMENT_FILENAME, COUNTERPART_FILENAME,
     file_is_pdf, get_document_path, get_extension, get_signed_url,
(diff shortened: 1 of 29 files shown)


app/main/views/suppliers.py
```diff
@@ -22,6 +22,23 @@
...
+AGREEMENT_APPROVED_MESSAGE = 'The agreement for {organisation_name} was approved. ' \
+                             'They will receive a countersigned version soon.'
+AGREEMENT_APPROVAL_CANCELLED_MESSAGE = 'The agreement for {organisation_name} had its approval cancelled. ' \
+                                       'You can approve it again at any time.'
+UPLOAD_COUNTERSIGNED_AGREEMENT_MESSAGE = "Countersigned agreement file was uploaded"
+COUNTERSIGNED_AGREEMENT_NOT_PDF_MESSAGE = "Countersigned agreement file is not a PDF"
+SUPPLIER_SERVICES_REMOVED_MESSAGE = "You removed all of {supplier_name}'s '{framework_name}' services"
+SUPPLIER_USER_MESSAGES = {
+    'user_invited': 'User invited',
+    'user_not_invited': 'Not invited',
```
galund: This one isn't actually used?

app/main/views/suppliers.py
```diff
@@ -232,6 +246,8 @@ def list_countersigned_agreement_file(supplier_id, framework_slug):
     agreements_bucket = s3.S3(current_app.config['DM_AGREEMENTS_BUCKET'])
     countersigned_agreement_document = agreements_bucket.get_key(supplier_framework.get('countersignedPath'))
 
+    remove_countersigned_agreement_confirm = request.args.get('remove_countersigned_agreement') == 'True'
```
galund: (Nitpicking now) Is it me or is the capital letter here a bit eugh?
```
bool(yaml.safe_load(request.args.get('remove_countersigned_agreement', '')))
```
I think maybe even for suggesting this I should probably go and have a lie down.

risicle: We do somewhere amongst our legions of functions which call functions which call functions (...) have a "is string yes-sy?" function.

katstevens: Yes-y or truthy? I don't really want to accept any old truthy string - I've been stung by params like `are_you_sure=no` in the past.

I will definitely make `true` lower case though. As you may be able to tell I didn't expend a great deal of thought on this bit :)

risicle: Yes-y. Truthy would just be `bool(mystr)`.

katstevens: I've used the utils `convert_to_boolean` (I think that's the one you meant?) which should cover the common cases.

katstevens:
Thanks both - I'll squash then merge.

PR #375, opened 2018-04-04
config: add user.marketplace.team to DM_NOTIFY_REDIRECT_DOMAINS_TO_ADDRESS

risicle:
This means our functional tests which use sanitized db dump accounts could
send emails and they also then have an email domain available to them that
is admin-qualified.

What do people think?

Should I _also_ add a `admin.example.gov.uk` explicitly for this purpose?

### DIFF
--- a/config.py
+++ b/config.py
@@ -101,6 +101,7 @@ class Live(Config):
     DM_NOTIFY_REDIRECT_DOMAINS_TO_ADDRESS = {
         "example.com": "success@simulator.amazonses.com",
         "example.gov.uk": "success@simulator.amazonses.com",
+        "user.marketplace.team": "success@simulator.amazonses.com",
     }
 
     NOTIFY_TEMPLATES = {


galund:
I don't understand what the point is of redirecting any of these emails to the Amazon test address?

risicle:
Because *not* redirecting them potentially annoys Notify because they get loads of bounces.

galund:
If we're going to do something special anyway, then we might as well not bother sending the emails. 

I'm quite surprised that Notify care given that they must handle zillions of bounces every day, I'd have thought they'd *more* care that we are making API calls at all when we don't even actually want to send an email :-)

risicle:
But that's why it's an *integration* test. See https://trello.com/c/YQB2HkiD/225-functional-tests-send-emails-to-invalid-email-addresses

galund:
Sure, okay fine then. I still think it's a bug in Notify if their simulate-delivered addresses don't actually correctly simulate delivery - we shouldn't really be using the amazon one here.

PR #379, opened 2018-04-27
 Add link to 'change password' to admin dashboard

CrystalPea:
The link has been placed in a new section, 'Account settings', to make it consistent with other apps

This work follows our security report recommendation (see ticket for link to report)
The Trello ticket: https://trello.com/c/TieWrBG5/44-password-reset-link-from-logged-in-session

screen shot 2018-04-27 at 12 12 24

### DIFF
--- a/app/templates/_base_page.html
+++ b/app/templates/_base_page.html
@@ -22,6 +22,7 @@
   {% block breadcrumb %}{% endblock %}
   <div id="wrapper">
     <main id="content" role="main">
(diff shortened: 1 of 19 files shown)


katstevens:
We've looked at the dashboard with different admin role accounts, and the larger heading size looks kind of weird as well (because the admin dashboard itself never really got a proper design), but has the added disadvantage of drawing attention to a link that's less frequently used. So we'll stick with the smaller heading for now: we can always change it later.

PR #380, opened 2018-05-08
3. Update user csv download to reflect new api response

benvand:
See:
https://github.com/alphagov/digitalmarketplace-api/pull/748


samuelhwilliams:
As-is this will leave us incompatible for a short interim period. Are we happy with this? If yes and we can release at a time where users won't be affected, that's probably alright.

If not, we'll need to do this as two PRs.

PR #381, opened 2018-05-08
1. Remove field as interim compatibility measure whilst we update the api

benvand:
Remove field to make way for changes to api
https://github.com/alphagov/digitalmarketplace-api/pull/748


samuelhwilliams:
As discussed, we could maybe duplicate the field under the new key name so that the report still compiles with all of the information.

PR #382, opened 2018-05-10
Public stats page

samuelhwilliams:
## Summary
Remove the role check on the framework application stats page. This will
allow us to display these stats e.g. in the office without having to
leave a privileged admin user logged into the marketplace. The admin
site is protected by an IP whitelist by the digitalmarketplace-router,
which should provide sufficient protection - added to the fact that the
data we show here should not be particularly sensitive and much of it is
available on the performance platform already.

 ## Ticket
https://trello.com/c/zTkPCGSY/289-please-create-an-admin-account-which-can-access-framework-application-stats-so-we-can-look-at-them-on-the-big-tv

### DIFF
--- a/app/main/__init__.py
+++ b/app/main/__init__.py
@@ -3,6 +3,7 @@ from flask_login import current_user, login_required
 
 
 main = Blueprint('main', __name__)
+public = Blueprint('public', __name__)  # Admin login not required
 
 from .views import (
     agreements, communications, search, service_updates, services, suppliers, stats, users, buyers, admin_manager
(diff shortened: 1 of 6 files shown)


samuelhwilliams:
Um yes that's a much better idea 🤦‍♂️ 

Not sure what I was thinking... :)

app/main/__init__.py
```diff
@@ -1,8 +1,9 @@
-from flask import Blueprint, current_app
+from flask import Blueprint, current_app, request
```
katstevens: Don't think we need `request` here anymore. The Admin FE has flake8 configured to ignore unused imports in all the `__init__`s, so this is easy to miss!

samuelhwilliams: Removed. Well spotted 👍

PR #387, opened 2018-05-16
List all supplier users

samuelhwilliams:
## Summary
Iterate over all pages of users when returning user accounts associated
with a given supplier account, otherwise we will only list the first 100
accounts.

 ## Ticket
https://trello.com/c/jDveO4Ro/463

### DIFF
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -431,11 +431,11 @@ def find_supplier_users():
         abort(404)
 
     supplier = data_api_client.get_supplier(request.args['supplier_id'])
-    users = data_api_client.find_users(request.args.get("supplier_id"))
+    users = data_api_client.find_users_iter(request.args.get("supplier_id"))
 
     return render_template(
         "view_supplier_users.html",
-        users=users["users"],
+        users=users,
         invite_form=EmailAddressForm(),
         move_user_form=MoveUserForm(),
         supplier=supplier["suppliers"]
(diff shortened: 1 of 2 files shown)


app/main/views/suppliers.py
```diff
@@ -431,11 +431,11 @@ def find_supplier_users():
         abort(404)
 
     supplier = data_api_client.get_supplier(request.args['supplier_id'])
-    users = data_api_client.find_users(request.args.get("supplier_id"))
+    users = data_api_client.find_users_iter(request.args.get("supplier_id"))
```
katstevens: I see we call `data_api_client.find_users` in the `invite_user` and `move_user_to_new_supplier` views as well. I think we only use the result if there's a form validation error, but if I remember rightly that error page shows the list of users. Whether it should do that or not is another question, but we probably need to use `iter` in these cases too.

samuelhwilliams: Good idea. All calls to `find_users` have been updated with calls to `find_users_iter`. I'm going to also go through the other apps and do the same, and then create a tech+product debt ticket to indicate that we should consider adding pagination to these views.

lfdebrux: Would a better upgrade path be to change `find_users_iter` to return the same type as `find_users` and then change `find_users` to use `find_users_iter`? Is there any reason not to use an iterator?

samuelhwilliams: I think our ideal end-game is for views that list users to be paginated properly (like our G-Cloud search or G-Cloud supplier A-Z), rather than create an excessively long page.

This PR is the minimum fix required to allow users to complete the task they're trying to accomplish, but ideally this should get product input and add pagination to each of the views in order to remain performant and scale appropriately (i.e. if a supplier had 1000 users, we wouldn't want to hit the API 10 times in order to get all of those users because the page just gets slower and slower as more users are added - we'd want to either add pagination or better search functionality, etc).

PR #388, opened 2018-05-23
Update utils version to wrap errors from all third party services

CrystalPea:
Connected to this pull request: https://github.com/alphagov/digitalmarketplace-utils/pull/393/files


katstevens:
I'm pretty sure we don't actually use Mandrill in the Admin. There is a more recent utils version (36.12.0) that includes Notify and Mailchimp logging too - shall we pull that one in instead?

samuelhwilliams:
Yessss please 👍

samuelhwilliams:
This PR still says 36.10.1 - think you might need to push the update :)

CrystalPea:
@samuelhwilliams you are right, I forgot to `git add` when amending the commit, whoops 😳

PR #389, opened 2018-05-29
Add G-Cloud 10 management to ccs-sourcing

samuelhwilliams:
## Summary
CCS Sourcing need to be able to manage and edit declarations for G-Cloud
10. Until we have an opportunity to re-design this page/journey, we will
simply add G-Cloud 10 to the existing page. It does make things a little
tight and the design clearly doesn't hold up, but we have a ticket to
prioritise this work in the next sprint or so.

 ## Ticket
https://trello.com/c/Voz6ZjDV/216


samuelhwilliams:
Yep, just given we expect to be removing this page soon, I went for MVP.

PR #391, opened 2018-06-05
Change session time-out from 4 hours to 1 hour

CrystalPea:
Trello ticket: https://trello.com/c/KvdDwvnk/400-inadequate-session-timeout


CrystalPea:
It turns out we do not need to add session timeout variable to admin. The one on user-frontend is enough for all the frontends.

PR #392, opened 2018-06-07
Supplier CSV download for admin framework manager

katstevens:
Trello: https://trello.com/c/gVkJo8Zj/70-update-framework-manager-user-download-list-to-include-additional-fields

- Adds a new CSV download view for supplier info.
- Changes link URL and copy on `download_framework_users.html` template
screenshot-localhost-2018 06 13-10-59-24

This uses a new API client version (16.3.0 https://github.com/alphagov/digitalmarketplace-apiclient/pull/140) and a new API view (https://github.com/alphagov/digitalmarketplace-api/pull/797).

### DIFF
--- a/app/main/views/users.py
+++ b/app/main/views/users.py
@@ -121,6 +121,61 @@ def download_users(framework_slug):
     )
 
 
+@main.route('/frameworks/<framework_slug>/suppliers/download', methods=['GET'])
+@role_required('admin-framework-manager', 'admin')
+def download_suppliers(framework_slug):
+    framework = data_api_client.get_framework(framework_slug).get("frameworks")
+
+    supplier_rows = data_api_client.export_suppliers(framework_slug).get('suppliers', [])
+
+    supplier_and_framework_headers = [
+        "supplier_id",
+        "supplier_name",
+        "supplier_organisation_size",
+        "duns_number",
+        "companies_house_number",
+        "registered_name",
+        "declaration_status",
+        "application_status",
+        "application_result",
+        "framework_agreement",
+        "variations_agreed",
+    ]
+    service_count_headers = [lot['slug'] for lot in framework['lots']]
+    contact_info_headers = [
+        'contact_name',
+        'contact_email',
+        'contact_phone_number',
+        'address_first_line',
+        'address_city',
+        'address_postcode',
+        'address_country',
+    ]
+
+    download_filename = "suppliers-on-{}.csv".format(framework_slug)
+
+    formatted_rows = [
+        supplier_and_framework_headers + [
+            "service-count-{}".format(h) for h in service_count_headers
+        ] + contact_info_headers
+    ]
+    for row in supplier_rows:
+        formatted_rows.append(
+            [row[heading] for heading in supplier_and_framework_headers] +
+            [row['published_services_count'][heading] for heading in service_count_headers] +
+            [row['contact_information'][heading] for heading in contact_info_headers]
+        )
+
+    return Response(
+        csv_generator.iter_csv(formatted_rows),
+        mimetype='text/csv',
+        headers={
+            "Content-Disposition": "attachment;filename={}".format(download_filename),
+            "Content-Type": "text/csv; header=present"
+        }
+    )
+
+
 @main.route('/users/download/buyers', methods=['GET'])
 @role_required('admin-framework-manager', 'admin')
 def download_buyers():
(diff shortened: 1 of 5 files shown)


app/main/views/users.py
```diff
@@ -121,6 +121,62 @@ def download_users(framework_slug):
...
+    download_filename = "suppliers-on-{}.csv".format(framework_slug)
+
+    formatted_rows = []
+    for row in supplier_rows:
+        formatted_rows.append(
+            [row[heading] for heading in supplier_and_framework_headers] +
+            [row['published_services_count'][heading] for heading in service_count_headers] +
+            [row['contact_information'][heading] for heading in contact_info_headers]
+        )
+    formatted_rows.insert(
```
benvand: Reads a bit out of the logical order here. We're building the csv into the `formatted_rows` variable and then going back and inserting the header afterpopulating the rows. Even though the header doesn't depend on the rows.
It's a bit like building the csv backwards, also that way we need to use the `[].insert` method instead of keeping it the same as above and using `[].append`.

app/templates/download_framework_users.html
```diff
@@ -28,13 +28,13 @@
     with
     items = [
       {
-          "title": "Suppliers who’ve applied or started an application",
+          "title": "Users who are on the framework, have applied or started an application",
```
benvand: This content reads a bit ambiguous to me. It reads like we're telling the user that users who are on the framework, have applied or started an application.

Does it mean:
`'Users who are on the framework, have applied, or have started an application'`
`'Users who are on the framework and have applied or started an application'`

tests/app/main/views/test_users.py
```diff
@@ -362,6 +362,137 @@ def test_download_csv_for_on_framework_only(self):
...
+        )
+        assert response.status_code == 200
+        assert response.mimetype == 'text/csv'
+        assert (response.headers['Content-Disposition'] ==
+                'attachment;filename=suppliers-on-digital-outcomes-and-specialists-2.csv')
+
+        rows = [line.split(",") for line in response.get_data(as_text=True).splitlines()]
+
+        assert len(rows) == len(list_of_expected_supplier_results) + 1
+        expected_top_level_headings = [
```
benvand: I think this splitting the `expected_headings` into 3 different variables here mirrors the code a bit too much. This is only done because the code in the view does it as far as I can tell.

We should be testing the code as if we didn't know how it was written. Only test the expected output rather than testing _how_ we get the expected outcome.

Really, technically, what we want to assert is that the headings are correct. Not that the headings are equal to `expected_top_level_headings + expected_service_count_headings + expected_contact_info_headings`. In this case it is the same thing but it doesn't add value to the test to split them because someone reading this test might wonder, 'Why should the headings equal 1 + 2 + 3?' Then they have to look at the code to find out that it's not important.

tests/app/main/views/test_users.py
```diff
@@ -362,6 +362,137 @@ def test_download_csv_for_on_framework_only(self):
...
+            'address_first_line',
+            'address_city',
+            'address_postcode',
+            'address_country',
+        ]
+
+        assert rows[0] == expected_top_level_headings + expected_service_count_headings + expected_contact_info_headings
+        # All suppliers returned from the API should appear in the CSV
+
+        for index, expected_supplier in enumerate(list_of_expected_supplier_results):
```
benvand: You can tell enumerate where to start the index with a second arg.
http://book.pythontips.com/en/latest/enumerate.html#enumerate

tests/app/main/views/test_users.py
```diff
@@ -362,6 +362,137 @@ def test_download_csv_for_on_framework_only(self):
...
+            assert rows[index + 1] == (
+                [str(expected_supplier[key]) for key in expected_top_level_headings] +
```
benvand: There's not much harm in just having hard coded outputs here.

Then this test would illustrate really nicely
1) the expected input `list_of_expected_supplier_results`
2) the expected process `GET /admin/frameworks/{}/suppliers/download`
3) the expected output (the hard coded csv values)

tests/app/main/views/test_users.py
katstevens: Can you tell I wrote this test in about 5 minutes 😹 I'll have another look this morning and tidy it up.

app/main/views/users.py
katstevens: Happy to change - as you can probably tell this was copied wholesale from the users export view above.

PR #404, opened 2018-07-12
digitalmarketplace-utils dependency: bump to 40.6.0 

risicle:
This brings a few logging niceties and also switches views to use `timed_render_template`.


risicle:
You're an idiot, Mr Scott - you've left `timed_render_template` with the message `Spent >{threshold}s in render_template` even when it was triggered due to the zipkin sampling header. Fix that.

PR #406, opened 2018-07-18
Do not allow admins to edit users whose personal data is removed

benvand:
https://trello.com/c/F3Bp7CVV/57-can-admin-reactivate-copy-change

This removes the 'Activate'/ 'Deactivate' and 'Unlock' buttons in user tables in the admin app. These are on the 'supplier users' and 'users' pages.

The pages now will display the attribute value ('Active', 'Inactive/ Disabled', 'Locked') intead of giving the admins a button to perform the respective action. This is in line with the view that admin users without the deactivate/ activate, unlock permissions get.

### Before 

screen shot 2018-07-23 at 14 28 42

screen shot 2018-07-23 at 14 28 52

### After

screen shot 2018-07-23 at 14 29 08

screen shot 2018-07-23 at 14 29 16

### DIFF
--- a/app/templates/view_supplier_users.html
+++ b/app/templates/view_supplier_users.html
@@ -67,7 +67,7 @@
       {% endcall %}
 
       {% call summary.field() %}
-      {% if item.locked and current_user.has_role('admin') %}
+      {% if item.locked and current_user.has_role('admin') and not item.personalDataRemoved %}
       <form action="{{ url_for('.unlock_user', user_id=item.id) }}" method="post">
           <input type="hidden" name="csrf_token" value="{{ csrf_token() }}"/>
           {%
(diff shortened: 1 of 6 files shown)


app/templates/view_supplier_users.html
```diff
@@ -87,7 +87,7 @@
 
       {% call summary.field() %}
       {% if item.active %}
-        {% if current_user.has_role('admin') %}
+        {% if current_user.has_role('admin') and not item.personalDataRemoved %}
```
Wynndow: Could you have the situation where a user was active, but also had their personal data removed? If not, this second conditional might be extra.

app/templates/view_users.html
```diff
@@ -108,7 +108,7 @@
 
       {% call summary.field() %}
           {% if item.active %}
-            {% if current_user.has_role('admin') %}
+            {% if current_user.has_role('admin') and not item.personalDataRemoved %}
```
Wynndow: Same question as above

tests/app/main/views/test_users.py
```diff
@@ -176,6 +186,27 @@ def test_should_show_deactivate_button(self):
         assert deactivate_button == 'Deactivate'
         assert return_link.attrib['value'] == '/admin/users?email_address=test.user%40sme.com'
 
+    @pytest.mark.parametrize('case', ({'active': False}, {'personalDataRemoved': True}))
```
risicle: Go on, add the case where they're both deactivated *and* personal data's removed... very cheap extra test.

risicle:
Testing locally...

PR #417, opened 2018-08-30
Remove requirement on virtualenv

lfdebrux:
Python36 includes the venv module which means we no longer need virtualenv to be installed on the development system


Makefile
```diff
@@ -34,7 +34,7 @@ requirements-dev: virtualenv test-requirements requirements-dev.txt
 .PHONY: freeze-requirements
 freeze-requirements:
 	rm -rf venv-freeze
-	virtualenv -p python3 venv-freeze
+	python3 -m venv venv-freeze
 	$$(pwd)/venv-freeze/bin/pip install -r requirements-app.txt
```
risicle: hmm for me, `venv` doesn't automatically install `pip`, leading this line to then fail.

risicle:
I've pushed the equivalent changes required for `default.nix` to work with `venv`.

(Looking into them, still think `venv`s are every bit as much mickey mouse crap as `virtualenv` but whatevs)

lfdebrux:
@risicle does that fix the `pip` issue?

Makefile
lfdebrux: This might be related to [this Travis error](https://travis-ci.org/alphagov/digitalmarketplace-frontend-toolkit/builds/422511041?utm_source=github_status&utm_medium=notification); apparently Debian systems have `python-venv` as a separate package.

risicle:
Yes thanks.

PR #418, opened 2018-08-31
Rejig view suppliers agreements list

Wynndow:
For [this trello ticket.](https://trello.com/c/Voz6ZjDV)

The list was laid out so that new frameworks would appear in a new
column. It was getting crowded.

This switches the view so a new framework adds a new row instead.

It also pulls this list of frameworks to display from the api rather
than being a hard coded list. As we don't display all frameworks, and
some frameworks are displayed differently (old signing flow frameworks),
we need a couple of constants to control this behaviour.

The Jury is still out on if we can drop the older frameworks and so reduce
the number of columns. Waiting to here from CCS.

### Before
screen shot 2018-08-31 at 12 24 33

### After 
screen shot 2018-08-31 at 12 24 05

### DIFF
--- a/app/templates/_view_suppliers_agreements.html
+++ b/app/templates/_view_suppliers_agreements.html
@@ -1,86 +1,31 @@
-<div class="six-columns-fixed-width-table">
-{% set field_headings = [
-  "Name",
-  "G-Cloud 7",
-  "Digital Outcomes and Specialists",
-  "G-Cloud 8",
-  "Digital Outcomes and Specialists 2",
-  "G-Cloud 9",
-  "G-Cloud 10",
-]
-%}
+<div class="five-columns-table">
+{% for supplier in suppliers %}
+  {{ summary.heading(supplier.name) }}
 
-{% call(item) summary.list_table(
-  suppliers,
-  caption="Suppliers",
-  empty_message="No suppliers were found",
-  field_headings=field_headings,
-  field_headings_visible=True)
-%}
-
-  {% call summary.row() %}
-  
-    {{ summary.field_name(item.name, rowspan=5) }}
-
+  {% call(framework) summary.list_table(
+    interesting_frameworks,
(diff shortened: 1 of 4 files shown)


app/templates/_view_suppliers_agreements.html
```diff
@@ -1,86 +1,31 @@
...
-        <a href="{{ url_for('.view_signed_agreement', supplier_id=item.id, framework_slug='g-cloud-9') }}">View agreement <span class="visually-hidden">for G-Cloud 9</span></a>
-      {% endcall %}
-      {% call summary.field(rowspan=3) %}
-        <a href="{{ url_for('.view_signed_agreement', supplier_id=item.id, framework_slug='g-cloud-10') }}">View agreement <span class="visually-hidden">for G-Cloud 10</span></a>
-      {% endcall %}
-    {% endcall %}
-
-    {% call summary.row(no_border=True) %}
-      {% call summary.field() %}
-        ⬇ <a href="{{ url_for('.download_signed_agreement_file', supplier_id=item.id, framework_slug='g-cloud-7') }}"><span class="visually-hidden">Download G-Cloud 7 </span>Signed agreement</a>
```
katstevens: I vaguely remember from the Admin & Automation mission that we wanted to get rid of these  'upload'/'download' arrows but didn't for some reason, and I can't remember what that was. It might have been to help users easily spot them given the poor layout of the page, which I guess now isn't a problem?

Wynndow: Hmm...i would guess that it would that it would have made the text more cramped in the old layout? Hopefully (🤞🏻) those old frameworks can be removed if we here back from CCS. Either way, Kuba came up with the design and I guess he would have shouted if he thought them still necessary?

katstevens: Cool - if Kuba is happy then I am happy 😸

PR #419, opened 2018-08-31
Upgrade to flask 1.0.2

benvand:
Upgrade to flask 1.0.2

http://flask.pocoo.org/docs/1.0/changelog/#version-1-0-2

https://trello.com/c/bBDir2Gp/18-flask-1x-upgrade


tests/app/main/views/test_services.py
```diff
@@ -1206,10 +1206,10 @@ def test_service_update_with_multiquestion_validation_error(self):
             (anchor.text_content(), anchor.get('href')) for anchor in
             document.xpath("//a[@class='validation-masthead-link']")
         ]
-        assert validation_banner_links == [
+        assert sorted(validation_banner_links) == sorted([
```
lfdebrux: 🤔 shouldn't this normally be in the order of the forms on the page?

benvand: https://github.com/alphagov/digitalmarketplace-content-loader/pull/52

This fix was never brought in.
They _are_ in the order of fields on the form now. They were in alphabetical order before.
Using sorted here is a good idea because we're not testing the order of the errors here... that is done elsewhere. We do want to assert that the errors trigger when the fields are supplier with the wrong info though

lfdebrux: Okay, that makes sense to me

PR #422, opened 2018-09-13
Download supplier and user CSVs from S3

katstevens:
Trello: https://trello.com/c/FtV7sVMU/138-admin-download-all-suppliers-for-framework-endpoint-locks-up-an-api-process

- Removes the old (painfully slow) views that generate CSVs on demand.
- Adds links to signed S3 URLs for the three reports (in the same style as for framework agreement downloads)
- Updated copy on the download pages to let admin users know that the reports are generated daily, not on demand. Checked this with @constancecerf who suggested rewording one of the download links.

admin-user-lists

### DIFF
--- a/app/main/views/users.py
+++ b/app/main/views/users.py
@@ -1,9 +1,9 @@
 from datetime import datetime
 
-from dmutils import csv_generator
-from dmutils.config import convert_to_boolean
+from dmutils import csv_generator, s3
+from dmutils.documents import get_signed_url
 from dmutils.flask import timed_render_template as render_template
-from flask import abort, flash, request, Response, url_for
+from flask import abort, current_app, flash, redirect, request, Response, url_for
 from flask_login import current_user
 
 from .. import main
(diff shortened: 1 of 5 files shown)


lfdebrux:
I just tried this out on my machine, I found that if you go to the supplier lists page, wait for at least a minute (until the signed s3 links expire) and then click one of the links, you get this screen

Chrome displaying an XML response from Amazon S3

katstevens:
Yeah, this is still in progress - going to reuse the old download views to wrap the S3 link (and display a proper message if we get that delightful XML back), that way we can keep the Content Disposition `attachment` to force a download as well.

lfdebrux:
Did you decide against setting Content Disposition? I still get the results opened in Google Sheets

lfdebrux:
I'm also getting a 404 when trying to download supplier names and emails, but I assume that's just because the file's not on S3 😆

katstevens:
> Did you decide against setting Content Disposition? I still get the results opened in Google Sheets

I decided to go with the method used elsewhere in the Admin (which the same user roles use, and should be familiar with), e.g. downloading agreement files.

PR #425, opened 2018-10-04
Temporarily re-enable dos2

Wynndow:
For [this trello card](https://trello.com/c/oEYKxqDK).

Expiring DOS2 caused it to disappear in important places from the admin app. CCS need access to some of the functionality for the framework so they can do their job.

DOS2 doesn't legally close for a while and CCS need to be able to remove suppliers from the framework if they don't submit their MI for three months in a row. They remove them by disabling their services.

### DIFF
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -541,6 +541,7 @@ def find_supplier_services(supplier_id):
 
     frameworks = data_api_client.find_frameworks()['frameworks']
     supplier = data_api_client.get_supplier(supplier_id)["suppliers"]
+    # TODO replace this temporary fix for DOS2 when a better solution has been created.
     services = data_api_client.find_services(
         supplier_id=supplier_id,
         framework=','.join(
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -541,8 +541,13 @@ def find_supplier_services(supplier_id):
 
     frameworks = data_api_client.find_frameworks()['frameworks']
     supplier = data_api_client.get_supplier(supplier_id)["suppliers"]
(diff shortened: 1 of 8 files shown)


app/main/views/suppliers.py
```diff
@@ -541,8 +541,14 @@ def find_supplier_services(supplier_id):
...
-    services = data_api_client.find_services(supplier_id)['services']
-
+    # TODO replace this temporary fix for DOS2 when a better solution has been created.
+    services = data_api_client.find_services(
+        supplier_id=supplier_id,
+        framework=','.join(
+            [f['slug'] for f in frameworks if f['status'] == 'live'
+                or f['slug'] == 'digital-outcomes-and-specialists-2']
+        )
+    )['services']
```
katstevens: A heads up that this view will display the service status as well:

screenshot-localhost-2018 10 05-09-51-03

If the services themselves are still 'public' that's what CCS Category will see (so they can edit them as usual). Would this be confusing? Our different types of admin users should know that DOS services aren't actually visible to buyers regardless of framework status, right?

Wynndow: Hmm...I see what you mean. I **hope** that they realise that in this sense "public" means "active" or "viable" or something. 

It feels like another thing pointing towards the idea of having a new frameworks status in between live and expired.

I think for now it's fine, as this is just a short term thing to allow them to do what they need to do, and they are experts in the domain so should understand what's going on.

katstevens: Well I guess we'll find out 😸 

Agreed on the need for a new status. `sunset`?

Wynndow: Ooo....that's a good one. 

I was leaning towards `cooldown`, but `sunset` is much more picturesque.

PR #430, opened 2018-11-02
use url_for with external url scheme for direct award service page links

risicle:
https://trello.com/c/kjIZjvPz/224-external-route-for-public-g-cloud-service-page-frameworkfamily-services-serviceid

Depends on https://github.com/alphagov/digitalmarketplace-utils/pull/470

Seems to work, even.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,7 +9,7 @@ itsdangerous==0.24
 lxml==3.8.0
 
 git+https://github.com/alphagov/digitalmarketplace-content-loader.git@5.0.0#egg=digitalmarketplace-content-loader==5.0.0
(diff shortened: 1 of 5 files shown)


requirements.txt
```diff
@@ -20,28 +20,28 @@ certifi==2018.10.15
...
 docutils==0.14
 Flask-Script==2.0.6
-future==0.17.0
+future==0.17.1
 idna==2.7
 inflection==0.3.1
 Jinja2==2.10
 jmespath==0.9.3
-mailchimp3==2.0.11
+mailchimp3==3.0.6
```
katstevens: We don't use mailchimp in the admin, but we should probably stick to v2.0.11 like we do elsewhere.

PR #432, opened 2018-11-14
241-deploy-228-pip-is-case-insensitive-and-our-requirements-freeze-job-is-not-and-169-file-upload-sniffing-type-restrictions-across

benvand:
https://trello.com/c/XEL7F0e9/241-deploy-228-pip-is-case-insensitive-and-our-requirements-freeze-job-is-not-and-169-file-upload-sniffing-type-restrictions-across

* Update requirements to bring in file checking dependencies and new freeze-requirements command
* Update makefile to use new freeze-requirements python command
* Remove python2 compatibility import of BytesIO as StringIO in test_services
* Use valid_pdf_bytes testutils fixture for creating test files
* Avoid reusing opened, mutable file object in multiple tests in test_services


Makefile
```diff
@@ -32,15 +32,8 @@ requirements-dev: virtualenv test-requirements requirements-dev.txt
...
-	rm -rf venv-freeze
-	python3 -m venv venv-freeze
-	$$(pwd)/venv-freeze/bin/pip install -r requirements-app.txt
-	echo '# This file is autogenerated. Do not edit it manually.' > requirements.txt
-	cat requirements-app.txt >> requirements.txt
-	echo '' >> requirements.txt
-	$$(pwd)/venv-freeze/bin/pip freeze -r requirements-app.txt | sed -n '/The following requirements were added by pip freeze/,$$p' >> requirements.txt
-	rm -rf venv-freeze
+freeze-requirements: virtualenv
+	python -m dmutils.repoutils.freeze_requirements requirements-app.txt
```
katstevens: I think we'd need to either activate the virtualenv first, or call this with `${VIRTUALENV_ROOT}/bin/python`.

Also, until the major bump goes in we need to install the original requirements if they're not there - I tried this out locally and got `Error while finding module specification for 'dmutils.repoutils.freeze_requirements' (ModuleNotFoundError: No module named 'dmutils.repoutils')` as my current venv doesn't have the latest utils yet.

katstevens: Hmmm I get another error from `freeze_requirements.py` now: `FileNotFoundError: [Errno 2] No such file or directory: 'venv-freeze/bin/pip': 'venv-freeze/bin/pip'`

It's successfully created the `venv-freeze` folder, but it's only got `python` in there, not `pip` 😿

benvand: Done the ones from the first comment, not sure about the 2nd? Can you give it another go now with the new changes?

katstevens: Argh! While this solves the error messages, it also stops `freeze-requirements` doing its thing. The `requirements` makefile step includes `test-requirements` which fails if there's a change in `requirements-app.txt`. 

So running `make freeze-requirements` gives:

```
$ make freeze-requirements
[ -z $VIRTUAL_ENV ] && [ ! -d venv ] && python3 -m venv venv || true
< boto3==1.7.82
requirements.txt doesn't match requirements-app.txt.
Run 'make freeze-requirements' to update.
make: *** [test-requirements] Error 1
```

I guess we'd have to do something like:

```
.PHONY: freeze-requirements
freeze-requirements: virtualenv upgrade-pip
	${VIRUTALENV_ROOT}/bin/pip install -r requirements.txt
	${VIRTUALENV_ROOT}/bin/python -m dmutils.repoutils.freeze_requirements requirements-app.txt
```

benvand: 🤔 I'm not keen on that. It feels like something that should be done as part fo the setup target.

I think this is why `test-flake8` and `test-python` don't have the `requirements` target.

I'd rather just do:
```
.PHONY: freeze-requirements
freeze-requirements: virtualenv
    ${VIRTUALENV_ROOT}/bin/python -m dmutils.repoutils.freeze_requirements requirements-app.txt
```
And assume that, because it's a dev command, the devs should know they need to install the requirements. It won't be an issue for anyone going forward. Just for us 5 in the transition.

What do you think?

benvand: Changed my mind again. I think the `requirements-dev` step shouldn't have the test, we're not interested in whether the files are identical when we're installing dev requirements. Maybe we're developing and they're supposed to be different? :shrug:
Then, because the `test-` and `freeze-requirements` jobs are dev commands they should have the _dev_ requirements installed
Pushed a new commit

Makefile
```diff
@@ -28,19 +28,12 @@ requirements: virtualenv test-requirements upgrade-pip requirements.txt
 	${VIRTUALENV_ROOT}/bin/pip install -r requirements.txt
 
 .PHONY: requirements-dev
-requirements-dev: virtualenv test-requirements requirements-dev.txt
```
katstevens: Good shout on this 👍

lfdebrux:
@benvand would you mind splitting up the first commit so the freeze stuff and the tests stuff are separate? It just makes it hard to follow the history otherwise.

PR #433, opened 2018-11-14
233 Enable CCS Category user to unsuspend all services in admin app

katstevens:
Trello: https://trello.com/c/uYpLxQb0/233-enable-ccs-user-to-unsuspend-all-g10-services-in-admin-app

CCS Category admins can now re-enable ('unsuspend') all disabled services for a supplier. Previously they could suspend all services at once (effectively removing a supplier from the framework), but had to re-enable each services individually.

Firstly, the wording in the templates and flash messages has been changed from 'remove services' to 'suspend services'. 

CCS Category admins will now be able to see a 'Unsuspend all services' link if there is at least one disabled service in the list for that framework. They will have to confirm the `unsuspend` action, just as with the `suspend` action.  'Private' services (removed by the supplier) will not be affected. Other types of admins will not see the link.

I renamed the `disable_supplier_services` view to `toggle_supplier_services` (as it can now turn them back on as well as off). 

Note for reviewers: the code might feel a bit verbose in places (especially in the template), but I've tried to be consistent with the rest of the admin.


katstevens:
Adding one more commit for some additional content.

PR #438, opened 2018-11-30
Update Dockerfile

jonodrew:
This commit updates the Dockerfile to use the new base Docker image, which contains Node 8.12.0


lfdebrux:
Already applied by #439

PR #452, opened 2019-01-02
Fixup pyup config

jonodrew:
This commit fixes the pyup config file. For some reason, even with search turned off, pyup keeps trying to update requirements.txt. Therefore I've forcibly turned off updates for that file.

[Ticket](https://trello.com/c/lWzDwjQ7/288-observe-pyup-pr-on-monday-see-if-theres-any-problems)

### DIFF
--- a/.pyup.yml
+++ b/.pyup.yml
@@ -2,15 +2,19 @@
 # https://pyup.io/docs/bot/config/
 # Check dependencies in _only_ requirements-app and requirements-dev, every week, and open PRs with PyUp prefix.
 
-schedule: "every day"
 search: False
 
 requirements:
+
   - requirements-app.txt:
       update: all
       pin: True
+
   - requirements-dev.txt:
       update: all
       pin: True
+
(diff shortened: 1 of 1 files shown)


benvand:
I think we need a schedule line? https://pyup.io/docs/bot/config/#schedules

jonodrew:
> I think we need a schedule line? https://pyup.io/docs/bot/config/#schedules

I've removed the scheduling so that it raises a PR per package. There are fewer packages, and I think this will make it easier to identify and separate out a breaking change if one occurs.

.pyup.yml
```diff
@@ -2,15 +2,19 @@
 # https://pyup.io/docs/bot/config/
 # Check dependencies in _only_ requirements-app and requirements-dev, every week, and open PRs with PyUp prefix.
```
lfdebrux:
```suggestion
# Check dependencies in _only_ requirements-app and requirements-dev, and open PRs with PyUp prefix.
```

PR #464, opened 2019-01-17
Search suppliers registered name

katstevens:
Trello: https://trello.com/c/aSOduJAQ/319-admin-search-for-registered-name-as-well-as-supplier-name

Pulls in API client changes to search by supplier registered name using the `name` param. 

I changed the admin form name from `supplier_name_prefix` to `supplier_name` to be less misleading.

I've included a small update to the template copy to make it clear that a) registered names are included as well as trading name b) any part of the name will match, not just the prefix. The phrase 'trading name' might be meaningless for buyers/suppliers but user research with CCS reveals they use it to refer to the public name that appears on the Digital Marketplace. TL:DR I'm going to run this past content/product before merging 😸 

Copy change:
admin-supplier-search-copy

### DIFF
--- a/app/main/helpers/pagination.py
+++ b/app/main/helpers/pagination.py
@@ -1,0 +1,33 @@
+from urllib.parse import urlparse, parse_qs
+
+
+def get_nav_args_from_api_response_links(links, prev_next_label, request_args, preserved_kwargs):
+    """
+    Get prev/next page numbers from object's API response, plus any extra request args for the URL
+    :param links: dict of links from the API response, e.g. {'self': ..., 'prev': ... 'next': ...}
+    :param prev_next_label: string, either 'prev' or 'next'
+    :param request: the request_args MultiDict with any GET args
+    :param preserved_kwargs: list of string kwargs that should be kept to build the prev/next URLs
+    :return: dict of URL params, or None if there's no prev/next link.
+
+    e.g. for the 'prev' link on page 2 of a supplier name search, we'd want
+             {'page': 1, 'supplier_name': 'foo'}
+
+    and on for the 'next' link we'd want {'page': 3, 'supplier_name': 'foo'}.
+
+    If there are no prev/next links in the response then return None - we don't want to build the URL.
+    """
+    if prev_next_label not in links:
+        return None
+
+    page_arg = parse_qs(urlparse(links[prev_next_label]).query)
+
+    navigation_args = {
+        'page': page_arg['page'][0] if page_arg else None
+    }
+
+    for kwarg in preserved_kwargs:
+        if request_args.get(kwarg):
+            navigation_args[kwarg] = request_args.get(kwarg)
+
+    return navigation_args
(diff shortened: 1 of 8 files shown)


katstevens:
Updated to add pagination (thanks to @risicle for input on this).

tests/app/main/views/test_suppliers.py
```diff
@@ -231,47 +213,79 @@ def test_should_raise_http_error_from_api(self):
...
         ]
-        response = self.client.get("/admin/suppliers")
+        response = self.client.get("/admin/suppliers?supplier_name=foo&page=2")
         document = html.fromstring(response.get_data(as_text=True))
 
         assert response.status_code == 200
         assert len(document.cssselect('.summary-item-row')) == 2
 
+        page_content = response.get_data(as_text=True)
+        assert 'Previous page' in page_content
```
risicle: I'd *personally* do these with a bit of xpath like `.xpath("//a[normalize-space(string())=$t]", t="Previous page")` and then yank the href out of the returned element, but it's not like any of the other tests around here do that.

app/main/views/suppliers.py
```diff
@@ -50,12 +51,16 @@
 def find_suppliers():
     if request.args.get("supplier_id"):
```
risicle: You probably didn't design this, but the behaviour of this view is weird - specifying a `supplier_id` makes it suddenly ignore other filtering parameters... guess this might also be a limitation from the API.

app/main/helpers/pagination.py
```diff
@@ -0,0 +1,33 @@
+from urllib.parse import urlparse, parse_qs
+
+
+def get_nav_args_from_api_response_links(links, prev_next_label, request_args, preserved_kwargs):
+    """
+    Get prev/next page numbers from object's API response, plus any extra request args for the URL
+    :param links: dict of links from the API response, e.g. {'self': ..., 'prev': ... 'next': ...}
+    :param prev_next_label: string, either 'prev' or 'next'
+    :param request: the request_args MultiDict with any GET args
+    :param preserved_kwargs: list of string kwargs that should be kept to build the prev/next URLs
```
risicle: Hey isn't this an opportunity for some type hints? :grin:

app/main/views/suppliers.py
katstevens: Yeah. To be honest I'd prefer a separate view for that ID search, with an explicit url pattern looking for `/supplier/<supplier_id>`. We might look at that when we do the 'display supplier info' page work for the CCS data user as that page is going to need a new view anyway. I'll keep it here for now.

app/main/helpers/pagination.py
katstevens: Ha! (Let's keep that to versioned libraries for now until we get a better idea of how useful/annoying it is)

tests/app/main/views/test_suppliers.py
katstevens: Agree - laziness on my part as there was nothing to easily copy and past 😸 Will add a commit to sharpen it up.

PR #484, opened 2019-02-18
Supplier details page

katstevens:
Trello: https://trello.com/c/3ZUU9xIn/340-supplier-details-admin-page

## What we did

- New view and template for supplier company details page
- Added permissions for CCS Data Controller to view (but not edit) supplier users, services and agreements
- Updated breadcrumbs
- Link the supplier name in search results (replacing 'Change name')
- Pulls in toolkit fix for displaying street address
- Pulls in utils for country code template filter
- Agreement page no longer 404s if the S3 url is not found (shows error message instead)

## What it should look like

admin-company-details-page

## Associated PRs

Functional tests for new view https://github.com/alphagov/digitalmarketplace-functional-tests/pull/606

### DIFF
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -618,13 +618,10 @@ def find_supplier_services(supplier_id):
 
     frameworks = data_api_client.find_frameworks()['frameworks']
     supplier = data_api_client.get_supplier(supplier_id)["suppliers"]
-    # TODO replace this temporary fix for DOS2 when a better solution has been created.
+
     services = data_api_client.find_services(
         supplier_id=supplier_id,
-        framework=','.join(
-            [f['slug'] for f in frameworks if f['status'] == 'live'
-                or f['slug'] == 'digital-outcomes-and-specialists-2']
-        )
+        framework=','.join(f['slug'] for f in frameworks if f['status'] in ['live', 'expired'])
     )['services']
     frameworks_services = {
         framework_slug: list(framework_services)
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -89,6 +89,71 @@ def find_suppliers():
     )
 
 
+@main.route("/suppliers/<int:supplier_id>", methods=["GET"])
+@role_required("admin-ccs-category", "admin-ccs-data-controller")
+def supplier_details(supplier_id):
+    def company_details_from_supplier(supplier):
+        return {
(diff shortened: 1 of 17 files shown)


app/main/views/suppliers.py
```diff
@@ -89,6 +89,71 @@ def find_suppliers():
...
+
+    supplier = data_api_client.get_supplier(supplier_id)["suppliers"]
+    supplier_frameworks = data_api_client.get_supplier_frameworks(supplier_id)["frameworkInterest"]
+
+    for framework_interest in supplier_frameworks:
+        framework_interest["framework"] = \
+            data_api_client.get_framework(framework_interest["frameworkSlug"])["frameworks"]
+
+    supplier_frameworks = (
+        d for d in supplier_frameworks
```
jonodrew: Could you make this variable name clearer?

tests/app/main/views/test_suppliers.py
```diff
@@ -10,12 +10,175 @@
...
+        self.data_api_client = self.data_api_client_patch.start()
+
+    def teardown_method(self, method):
+        self.data_api_client_patch.stop()
+        super().teardown_method(method)
+
+    def test_get_returns_404_if_supplier_id_is_invalid(self):
+        assert self.client.get("/admin/suppliers/invalid").status_code == 404
+        assert self.client.get("/admin/suppliers/-1").status_code == 404
+        self.data_api_client.get_supplier_frameworks.side_effect = HTTPError(Response(404))
```
jonodrew: I don't really understand what these two lines (38/39) are testing. As far as I can see, you're testing that it would work, except you're altering the mock to return a 404. But then doesn't that just test that `mock` returns a 404 when you ask it to?

lfdebrux: Hmm, it's more demonstrative than a test; the scenario is that the supplied supplier id does not exist. In this case get_supplier_frameworks will return a 404 error. As you say it is a bit redundant, but I think it's important to check that part of the contract; we want to make sure the view doesn't do something strange like turn the 404 into a 503.

katstevens: Let's add in an assert for the api client call 👍

lfdebrux: Will do

lfdebrux: Done

PR #488, opened 2019-03-06
Tidy supplier results view

jonodrew:
[Ticket](https://trello.com/c/t6awHVr9/392-ccs-sourcing-supplier-results-page-tidy-up)
I'll probably squash most of these together before merging into master, but I wanted to show my working and make it easier for my reviewer to make sense of this.

This PR is to change the ccs-sourcing-admin's journey. It brings it into line with the other admin roles while adding a small tweak that prevents them from seeing links they can't access.

Previous search results:
Previous search results

New search results:
screen shot 2019-03-06 at 16 51 45

New detail page:
screen shot 2019-03-06 at 16 52 37


tests/app/main/views/test_suppliers.py
```diff
@@ -295,122 +280,34 @@ def test_details_link_is_shown_to_users_with_right_roles(self, role, link_should
...
-            for text in [
-                'Edit declaration',
-                'View agreement',
-                'Download agreement',
-                'Download signed agreement',
-                'Upload countersigned agreement'
-            ]:
-                assert not document.xpath("//tbody//text()='$text'", text=text)
-
-    def test_agreements_template_only_lists_frameworks_that_we_care_about_in_reverse_order(self):
```
katstevens: Ah. This test reminds me that CCS Sourcing users see a different subset of frameworks than the other admin roles.

CCS Sourcing are interested in all framework statuses except `coming` (see https://github.com/alphagov/digitalmarketplace-admin-frontend/blob/master/app/main/views/suppliers.py#L74 for where the 'interesting_frameworks' template var is populated). 

In the details page, we're only showing `live` and `expired` (see https://github.com/alphagov/digitalmarketplace-admin-frontend/blob/master/app/main/views/suppliers.py#L132), as that's what CCS Category and the other admin roles care about. So we'll have to tweak that view to ensure sourcing can see links for `open`, `pending` and `standstill` frameworks.

app/templates/view_suppliers.html
```diff
@@ -24,11 +24,7 @@
   {% endwith %}
 
   <div class="page-section">
-    {% if current_user.has_role("admin-ccs-sourcing") %}
-      {% include "_view_suppliers_agreements.html" %}
```
katstevens: We can delete `_view_suppliers_agreements.html` now as it's not used - or can we? It's the only place where CCS Sourcing users can download documents for old frameworks (G7 and earlier - these can't be viewed as an embedded file, unlike later framework docs). They won't need the 'Upload countersigned agreement' link for the old frameworks as they should be done and dusted now.

A possible solution could be to amend the 'View agreements' link on the supplier details page's frameworks table for G7 and earlier, to point to the 'Download signed agreement' url that was previously on the search results.

tests/app/main/views/test_suppliers.py
jonodrew: I think I fixed this with 5c6c5b1

app/main/views/suppliers.py
```diff
@@ -120,17 +121,28 @@ def company_details_from_supplier_framework_declaration(declaration):
...
+                supplier_framework for supplier_framework in all_frameworks
+                if supplier_framework["framework"]["status"] != "coming"
+            )
+        else:
+            frameworks = (
+                supplier_framework for supplier_framework in supplier_frameworks
+                if supplier_framework["framework"]["status"] in ["live", "expired"]
+            )
+        oldest_interesting_id = data_api_client.get_framework(OLDEST_INTERESTING_FRAMEWORK_SLUG)['frameworks']['id']
+        return [framework for framework in frameworks if framework['framework']['id'] >= oldest_interesting_id]
```
katstevens: So many frameworks 😹 But these are `supplier_framework` objects, right? Might be better to rename here.

jonodrew:
> My functional tests passed locally  However given Jenkins is currently unhappy, I'd hold off merging for the moment as the preview tests are likely to fail (also the commits will need squashing).

Fair. I'll do the squashing now in any case

PR #491, opened 2019-03-15
Admin Data Controller can edit registered supplier details

katstevens:
Trello: https://trello.com/c/lw9XhnQS/403-data-controller-can-edit-supplier-details-in-admin

The CCS Data Controller role now sees this view on the supplier details admin page:
admin-company-details-page

- Pulls in accessible country autocomplete (frontend package), and some Python helper functions copied from the Supplier FE to load the files. (I'll create a tech debt ticket for this TBC)
- Pulls in latest API client, with declaration-updating methods
- Uses the newly refactored forms from the Supplier FE (including the country validation fix) for the registered name, address and number fields
- Refactors the code that fetches supplier info from either company details or the latest declaration, as we're going to reuse it
- New views and templates for the actual editing business (the templates are quite verbose at the moment, we can always consolidate later)


app/main/views/suppliers.py
```diff
@@ -195,6 +152,194 @@ def update_supplier_name(supplier_id):
...
 
 
+@main.route('/suppliers/<int:supplier_id>/edit/registered-name', methods=['GET', 'POST'])
+@role_required('admin-ccs-data-controller')
+def edit_supplier_registered_name(supplier_id):
+    supplier = data_api_client.get_supplier(supplier_id)['suppliers']
+
+    # Get SupplierFrameworks for frameworks the role is interested in, sorted by oldest frameworkLiveAtUTC first
+    supplier_framework_interests = data_api_client.get_supplier_frameworks(supplier_id)["frameworkInterest"]
+    supplier_frameworks = interesting_frameworks(supplier_framework_interests, current_user, data_api_client)
```
lfdebrux: We see this pattern a lot, where we get the framework interests and then filter using `interesting_frameworks`; is it worth adding the initial step into the helper?

katstevens: Potentially? I'm not super keen on the general pattern of passing the instantiated `data_api_client` through to a helper if we're using it at the view level as well, as it can often make test patching confusing.

lfdebrux: Hmm, the helper already uses the data_api_client?

katstevens: 🤦‍♀️ It probably shouldn't! And that's why this is a WIP...

tests/app/main/views/test_suppliers.py
```diff
@@ -1253,6 +1299,242 @@ def test_ccs_sourcing_role_can_not_update_supplier_name(self):
...
+        assert self.data_api_client.update_supplier_declaration.call_args_list == ([
+            mock.call(
+                1234, 'g-cloud-11',
+                {'supplierCompanyRegistrationNumber': "def98765"}, "test@example.com"
+            )
+        ] if update_declaration else [])
+        self.assert_flashes("The details for ‘ABC’ have been updated.")
+
+    @pytest.mark.parametrize('update_declaration', [True, False])
+    def test_data_controller_role_can_change_companies_house_to_other_registration_number(self, update_declaration):
```
lfdebrux: Do we need a test for the reverse case (other registration number to companies house number)?

tests/app/main/views/test_suppliers.py
```diff
@@ -169,57 +203,59 @@ def teardown_method(self, method):
             (
                 "admin-ccs-data-controller",
                 ["digital-outcomes-and-specialists-3", "g-cloud-10"],
-                ["g-cloud-11", "g-cloud-standstill", "g-cloud-pending", "g-cloud-open"]
+                ["g-cloud-11", "g-cloud-standstill-1", "g-cloud-pending-1", "g-cloud-open-1"]
```
lfdebrux: Why did you need to append "-1"?

katstevens: There was some function or other that was grabbing the suffix of the framework slug (that's a thought, I wonder what it did for `digital-outcomes-and-specialists`...)

app/templates/suppliers/edit_duns_number.html
```diff
@@ -0,0 +1,39 @@
...
+    ]
+  %}
+    {% include "toolkit/breadcrumb.html" %}
+  {% endwith %}
+{% endblock %}
+
+{% block main_content %}
+<div class="grid-row">
+<div class="column-two-thirds">
+  {% with heading = "Update {}’s DUNS number".format(supplier.name) %}
```
lfdebrux: Might be nicer to have these as

```suggestion
  {% with heading = "Update DUNS number for ‘{}’".format(supplier.name) %}
```

because then the name of the thing your editing is always in the same place (doesn't get pushed over by long supplier names) and we can put quotation marks around supplier name without worrying about apostrophes.

app/main/views/suppliers.py
```diff
@@ -191,10 +149,200 @@ def update_supplier_name(supplier_id):
...
+def edit_supplier_registered_address(supplier_id):
+    frameworks = data_api_client.find_frameworks()['frameworks']
+    supplier = data_api_client.get_supplier(supplier_id)['suppliers']
+    contact = supplier["contactInformation"][0]
+
+    # Get SupplierFrameworks for frameworks the role is interested in, sorted by oldest frameworkLiveAtUTC first
+    supplier_framework_interests = data_api_client.get_supplier_frameworks(supplier_id)["frameworkInterest"]
+    supplier_frameworks = interesting_frameworks(supplier_framework_interests, current_user, frameworks)
+
+    # Get the company details (either from supplier or recent declaration)
```
lfdebrux:
```suggestion
    # Get the company details (either from supplier or recent declaration)
    # TODO: handle case where they don't have a declaration better (maybe different form?)
```

katstevens: Actually I think I have a solution for this. Commit incoming...

tests/app/main/views/test_suppliers.py
katstevens: It's the `FrameworkStub` - for slugs starting `g-cloud-` it derives the lots and so on.

lfdebrux: Ah fair enough

app/templates/suppliers/edit_duns_number.html
katstevens: Approved - have added a commit.

tests/app/main/views/test_suppliers.py
katstevens: I think we're probably fine for now.

katstevens:
Re: updating older declarations. We did consider this, but given we can't update old framework agreement PDFs anyway, it seems sufficient to just update the details that will be displayed on the DM. So that means the centralised details on the Supplier object (and it associated contact information) for the Supplier FE, and the latest declaration (if there is one) where we source the details displayed on the Admin company details page. Any other declarations would never be shown anywhere to users (it's also a grey area about how far to backdate any changes).

app/main/views/suppliers.py
```diff
@@ -90,76 +102,22 @@ def find_suppliers():
...
+    frameworks = data_api_client.find_frameworks()["frameworks"]
     supplier = data_api_client.get_supplier(supplier_id)["suppliers"]
-    supplier_frameworks = data_api_client.get_supplier_frameworks(supplier_id)["frameworkInterest"]
+    supplier_framework_interests = data_api_client.get_supplier_frameworks(supplier_id)["frameworkInterest"]
 
-    for supplier_framework in supplier_frameworks:
-        supplier_framework["framework"] = \
-            data_api_client.get_framework(supplier_framework["frameworkSlug"])["frameworks"]
+    # Get SupplierFrameworks for frameworks the role is interested in, sorted by oldest frameworkLiveAtUTC first
+    supplier_frameworks = interesting_frameworks(supplier_framework_interests, current_user, frameworks)
```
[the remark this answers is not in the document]
katstevens: Yep, agreed. `interesting` is a hangover from the previous CCS Sourcing search result view, (now deleted) so happy to change.

app/main/views/suppliers.py
```diff
@@ -90,76 +102,22 @@ def find_suppliers():
...
-                if supplier_framework["framework"]["status"] in ["live", "expired"]
-            )
-        oldest_interesting_id = data_api_client.get_framework(OLDEST_INTERESTING_FRAMEWORK_SLUG)['frameworks']['id']
-        return [supplier_framework for supplier_framework in frameworks
-                if supplier_framework['framework']['id'] >= oldest_interesting_id]
-
+    frameworks = data_api_client.find_frameworks()["frameworks"]
     supplier = data_api_client.get_supplier(supplier_id)["suppliers"]
-    supplier_frameworks = data_api_client.get_supplier_frameworks(supplier_id)["frameworkInterest"]
+    supplier_framework_interests = data_api_client.get_supplier_frameworks(supplier_id)["frameworkInterest"]
```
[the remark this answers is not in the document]
katstevens: Yeah I think `get_framework_interest` is fetching a singular one, isn't it? OK, can change.

app/main/helpers/supplier_details.py
```diff
@@ -0,0 +1,88 @@
...
+            supplier_framework['frameworkName'] = framework_info.get(supplier_framework["frameworkSlug"], {})['name']
+            interesting_supplier_frameworks.append(supplier_framework)
+
+    return sorted(
+        interesting_supplier_frameworks,
+        key=lambda sf: framework_info[sf["frameworkSlug"]]['frameworkLiveAtUTC']
+    )
+
+
+def get_company_details_and_most_recent_interest(supplier_frameworks, supplier):
```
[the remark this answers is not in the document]
katstevens: Yeah - I've spotted a way where we can slim down even further - commit incoming.

PR #503, opened 2019-04-17
Suspend services should be available for live frameworks only

katstevens:
Trello: https://trello.com/c/Gz6rXcY8/721-admin-suspending-all-services-for-a-non-live-framework-gives-500s-when-trying-to-delete-from-elasticsearch

Following a bunch of 500 errors when the API was trying to de-index the removed services from a non-existent index (g-cloud-8).

- Only show the `Suspend services / Unsuspend services` links for live frameworks
- Only show the `remove` and `publish` links for individual services for live frameworks
- Raise a 400 if the POST route is called with a non-live framework
- Change the display of expired services to not be 'public'

expired-view-services-admi

The tests around this still use G8 fixtures 😿 but these can be refactored another time.

### DIFF
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -850,6 +850,10 @@ def toggle_supplier_services(supplier_id):
     if not toggle_action['framework_slug']:
         abort(400, 'Invalid framework')
 
+    framework = data_api_client.get_framework(toggle_action['framework_slug'])['frameworks']
+    if framework['status'] != 'live':
+        abort(400, "Cannot toggle services for framework {}".format(toggle_action['framework_slug']))
+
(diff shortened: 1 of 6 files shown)


app/main/views/suppliers.py
```diff
@@ -850,6 +850,10 @@ def toggle_supplier_services(supplier_id):
     if not toggle_action['framework_slug']:
```
risicle: This is actually about the line a few up from here - what happens if both `remove_services` and `publish_services` are truthy and are _different_ framework slugs?

katstevens: I doubt we thought ahead that far. I think it will default to the `remove` action if both are supplied.

PR #512, opened 2019-05-13
Fix bug where G8 suppliers would appear to have no company details

lfdebrux:
Ticket: https://trello.com/c/fY2frqFG/490-supplier-company-details-dont-use-details-from-g8-or-earlier

If the supplier framework declaration is not in the expected format (as used in G9/DOS2 and upwards) then show the supplier account company details.

### DIFF
--- a/tests/app/main/views/test_suppliers.py
+++ b/tests/app/main/views/test_suppliers.py
@@ -202,6 +202,57 @@ class TestSupplierDetailsView(LoggedInApplicationTest):
             }
         }
 
+    @mock.patch("app.main.views.suppliers.render_template", return_value="")
+    def test_if_most_recent_framework_has_old_declaration_show_account_company_details(self, render_template):
+        self.data_api_client.get_supplier_frameworks.return_value = {
+            "frameworkInterest": [
+                SupplierFrameworkStub(
+                    framework_slug="g-cloud-8",
+                    declaration={
+                        "dunsNumber": mock.sentinel.duns_number,
+                        "currentRegisteredCountry": "country:FR",
+                        "registeredVATNumber": "12345678",
+                        "registeredAddressBuilding": "123 Rue Morgue",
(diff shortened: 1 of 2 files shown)


tests/app/main/views/test_suppliers.py
```diff
@@ -202,6 +202,57 @@ def test_if_most_recent_framework_has_no_declaration_show_account_company_detail
...
+            "frameworkInterest": [
+                SupplierFrameworkStub(
+                    framework_slug="g-cloud-8",
+                    declaration={
+                        "dunsNumber": mock.sentinel.duns_number,
+                        "currentRegisteredCountry": "country:FR",
+                        "registeredVATNumber": "12345678",
+                        "registeredAddressBuilding": "123 Rue Morgue",
+                        "registeredAddressTown": "Paris",
+                        "registeredAddressPostcode": "76876",
```
katstevens: I think the problem we were having was slightly different, in that there was a declaration, but it didn't have any of the fields above, just the application answers. However the supplier in question has since been updated, and I can't find another example! Perhaps this was a one off?

lfdebrux: Interesting...

I guess this fix should still work? I can add another test scenario

katstevens: OK. It's G7 where the company data isn't included in the declaration (but the declaration has other stuff in it).

The G7 declarations are pretty different to the subsequent ones so we shouldn't use them for fetching company details. They'll look something like:

```
"declaration": {
        "AQA3": true,
        "PR1": true,
        "PR2": true,
        "PR3": true,
        "PR4": true,
        "PR5": true,
        ...
}
```

Admins still want to view them, but we won't be able to get the DUNS etc from there.

lfdebrux:
@katstevens has said that actually we want it so that we're always showing the company details from the supplier's account, so imma close this PR and start a new one.

PR #516, opened 2019-05-16
Use supplier account details for company details view

lfdebrux:
Ticket: https://trello.com/c/fY2frqFG/490-supplier-company-details-always-show-admin-the-account-company-details

Using the most recent framework declaration was causing some confusion, and issues with changes between frameworks in how the company details were stored, so instead we are changing the behaviour so that the supplier account details are always shown.

This means we get to delete a load of code :fire:

### DIFF
--- a/tests/app/main/views/test_suppliers.py
+++ b/tests/app/main/views/test_suppliers.py
@@ -116,37 +116,7 @@ class TestSupplierDetailsView(LoggedInApplicationTest):
             )
 
     @mock.patch("app.main.views.suppliers.render_template", return_value="")
-    def test_view_shows_company_details_from_suppliers_last_framework_declaration(self, render_template):
-        framework_interest = SupplierFrameworkStub(framework_slug="g-cloud-11").response()
-        framework_interest["declaration"]["supplierRegisteredPostcode"] = mock.sentinel.postcode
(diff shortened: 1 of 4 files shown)


tests/app/main/views/test_suppliers.py
```diff
@@ -1541,16 +1463,10 @@ def test_data_controller_role_can_update_registered_company_name(self, from_decl
         self.data_api_client.update_supplier.assert_called_once_with(
             1000, {'registeredName': "Something New"}, "test@example.com"
         )
-        assert self.data_api_client.update_supplier_declaration.call_args_list == ([
-            mock.call(
-                1000, 'g-cloud-11',
-                {'supplierRegisteredName': "Something New"}, "test@example.com"
-            )
-        ] if from_declaration else [])
+        assert self.data_api_client.update_supplier_declaration.call_args_list == []
```
katstevens: I don't think we need to assert this non-call if the code being tested doesn't use the method anywhere. More stuff to delete 😸

lfdebrux: 👍

lfdebrux: 👍

PR #528, opened 2019-06-28
Add ability to delete/download communications

risicle:
https://trello.com/c/jZVhvJRA

I doing this I had to implement a number of extra things that weren't in the original card. The existing upload interface was extremely minimal.

To allow admins to choose a file to delete we first have to give them a listing of files which it didn't previously have.

image

Then if admins are supposed to be certain about the file they're about to delete, we really _have_ to give them the ability to download that file first, an ability it was missing.

Then of course the delete interface really did need to have a "confirm" page.

image

~The code is basically done, when I'm still working on is the element layout etc - I'll add more details of this on the trello card because I could do with some opinions...~

### DIFF
--- a/app/main/views/communications.py
+++ b/app/main/views/communications.py
@@ -1,11 +1,23 @@
+from pathlib import PurePath
+
 from dmutils import s3  # this style of import so we only have to mock once
 from dmutils.documents import file_is_pdf, file_is_csv, file_is_open_document_format
 from dmutils.flask import timed_render_template as render_template
-from flask import redirect, url_for, current_app, request, flash
+from flask import redirect, url_for, current_app, request, flash, abort
 
 from .. import main
 from ..auth import role_required
 from ... import data_api_client
+from ..helpers.frameworks import get_framework_or_404
+
+
+def _get_comm_type_root(framework_slug, comm_type):
+    # PurePath has the advantage that it doesn't handle ".." elements specially, and so we're safe from a whole
+    # category of tricks
+    return PurePath(framework_slug) / "communications" / "updates" / f"{comm_type}s"
+
+
+_comm_types = ("communication", "clarification",)
 
 
 @main.route('/communications/<framework_slug>', methods=['GET'])
@@ -65,3 +77,36 @@ def upload_communication(framework_slug):
             flash('New clarification was uploaded.')
 
     return redirect(url_for('.manage_communications', framework_slug=framework_slug))
+
+
+@main.route('/communications/<framework_slug>/delete/<string:comm_type>/<path:filepath>', methods=("GET", "POST",))
+@role_required('admin-framework-manager')
+def delete_communication(framework_slug, comm_type, filepath):
+    if comm_type not in _comm_types:
+        abort(404)
+
+    framework = get_framework_or_404(data_api_client, framework_slug)
+
+    if request.method == "POST":
+        if "confirm" not in request.form:
+            abort(400, "Expected 'confirm' parameter in POST request")
+
+        communications_bucket = s3.S3(current_app.config['DM_COMMUNICATIONS_BUCKET'])
+        full_path = _get_comm_type_root(framework_slug, comm_type) / filepath
+
+        # do this check ourselves - deleting an object in S3 silently has no effect, forwarding this behaviour to the
+        # user is a confusing thing to do
+        if not communications_bucket.path_exists(str(full_path)):
+            abort(404, f"{filepath} not present in S3 bucket")
+
+        communications_bucket.delete_key(str(full_path))
+
+        flash(f"{comm_type.capitalize()} ‘{filepath}’ was deleted for {framework['name']}.")
+        return redirect(url_for('.manage_communications', framework_slug=framework_slug))
+
+    return render_template(
+        'confirm_communications_deletion.html',
+        framework=framework,
+        comm_type=comm_type,
+        filepath=filepath,
+    )
--- a/app/main/views/communications.py
+++ b/app/main/views/communications.py
@@ -47,7 +47,7 @@ def manage_communications(framework_slug):
     )
 
 
-@main.route('/communications/<framework_slug>/files/<string:comm_type>/<path:filepath>', methods=['GET'])
+@main.route('/communications/<framework_slug>/files/<string:comm_type>/<safepurepath:filepath>', methods=['GET'])
 @role_required('admin-framework-manager')
 def download_communication(framework_slug, comm_type, filepath):
     if comm_type not in _comm_types:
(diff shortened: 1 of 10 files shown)


lfdebrux:
Screenshots look good. Has Stefan had any input on the content?

lfdebrux:
Can we do a bit of rearranging of the commit history so the tests for each feature are with the commit adding the feature?

app/main/views/communications.py
```diff
@@ -1,37 +1,69 @@
...
-        '{}/communications/updates/clarifications'.format(framework_slug), load_timestamps=True
-    ) or [None]
+    # generate a dict of comm_type: seq of s3 object dicts
+    comm_type_objs = {
+        comm_type: tuple(
+            {
+                **obj_dict,
+                # annotate on to object dicts their paths relative to comm_type_root
+                "rel_path": PurePath(obj_dict["path"]).relative_to(_get_comm_type_root(framework_slug, comm_type)),
+            } for obj_dict in communications_bucket.list(
```
lfdebrux: I feel like renaming `obj_dict` to `bucket_object` or `bucket_item` would make it a bit easier to follow what's going on here.

```suggestion
            } for bucket_listing in communications_bucket.list(
```

app/main/views/communications.py
```diff
@@ -1,37 +1,69 @@
...
 from ... import data_api_client
 
 
+def _get_comm_type_root(framework_slug, comm_type):
+    # PurePath has the advantage that it doesn't handle ".." elements specially, and so we're safe from a whole
+    # category of tricks
+    return PurePath(framework_slug) / "communications" / "updates" / f"{comm_type}s"
+
+
+_comm_types = ("communication", "clarification",)
```
lfdebrux: This might be overkill, but what about making this an enum type?

tests/app/main/views/test_communications.py
```diff
@@ -39,88 +52,198 @@ def test_get_page_should_only_be_accessible_to_specific_user_roles(self, role, e
         assert actual_code == expected_code, "Unexpected response {} for role {}".format(actual_code, role)
 
     def test_page_shows_empty_messages_if_no_files_uploaded(self):
-        self.s3.return_value.list.side_effect = ([], [])  # Empty lists for both communication and clarification files
-        self.data_api_client.get_framework.return_value = {"frameworks": {"status": "open"}}
+        self.s3.return_value.list.side_effect = lambda *args, **kwargs: []  # return a fresh empty list every time
```
lfdebrux: Is someone mutating these lists? :thinking:

tests/app/main/views/test_communications.py
```diff
@@ -39,88 +52,198 @@ def test_get_page_should_only_be_accessible_to_specific_user_roles(self, role, e
...
         response = self.client.get("/admin/communications/{}".format(self.framework_slug))
         document = html.fromstring(response.get_data(as_text=True))
-        file_upload_messages = document.xpath('//p[@class="file-upload-existing-value"]')
+        file_upload_messages = document.xpath('//p[@class="summary-item-no-content"]')
 
         assert len(file_upload_messages) == 2
-        assert file_upload_messages[0].text_content().strip() == "No communications have been uploaded yet"
-        assert file_upload_messages[1].text_content().strip() == "No clarification answers have been uploaded yet"
+        assert file_upload_messages[0].text_content().strip() == "No communications files"
+        assert file_upload_messages[1].text_content().strip() == "No clarification answers"
```
lfdebrux: Just out of curiosity, what was the reasoning behind changing the empty message?

tests/app/main/views/test_communications.py
```diff
@@ -39,88 +52,198 @@ def test_get_page_should_only_be_accessible_to_specific_user_roles(self, role, e
...
+
+        assert self.data_api_client.mock_calls == [
+            mock.call.get_framework(self.framework_slug)
+        ]
+        assert self.s3.mock_calls == [
+            mock.call("flop-slop-slap"),
+            mock.call().list('g-things-23/communications/updates/communications', load_timestamps=True),
+            mock.call().list('g-things-23/communications/updates/clarifications', load_timestamps=True),
+        ]
```
lfdebrux: I feel like these assertions would be clearer if we attached the return value of the s3 mock to the test class; then we could write things like `self.communications_bucket.list.mock_calls`

tests/app/main/views/test_communications.py
```diff
@@ -39,88 +52,198 @@ def test_get_page_should_only_be_accessible_to_specific_user_roles(self, role, e
...
+                    '/admin/communications/g-things-23/files/clarification/clarification-dax.pdf',
+                    'document: clarification-dax.pdf',
+                    '/admin/communications/g-things-23/delete/clarification/clarification-dax.pdf',
+                ),
+            ),
+        ) if framework_status == "open" else ())
+
+        # find the form that has our submit button
+        form = doc.xpath(
+            "//form[@method='post'][@enctype='multipart/form-data'][.//input[@type='submit'][@value='Upload files']]"
```
lfdebrux: Maybe we should just put an id on the form? :laughing:

risicle:
> Can we do a bit of rearranging of the commit history so the tests for each feature are with the commit adding the feature?

Yyyyeah there are some complications due to the way I refactored the test class but I'll see what I can do...

tests/app/main/views/test_communications.py
risicle: But that doesn't test anything. This tests that the form is set up correctly, right method, right enctype, that there's an appropriate button...

lfdebrux: Ah I see, carry on then

tests/app/main/views/test_communications.py
risicle: I think that would introduce more possibilities for ... weirdness. People thinking you can just switch out the `self.communications_bucket` in a test and it'll start returning that... I'd also need to start adding another line to separately test that the constructor is called with the right parameter. Really, `mock_calls` is very powerful and gives us quite a nice concise way to describe the things that have happened to the object/class and I think we should use it more tbh.

tests/app/main/views/test_communications.py
risicle: It's the whole "yet" thing. _Before_ it was possible to say that because the only way a section could have no entries is in the initial state (unless someone had fiddled with the bucket manually). But now a section can be empty because it's had its files _deleted_, and in that case the message is misleading.

tests/app/main/views/test_communications.py
risicle: Dunno. If it doesn't now, someone could change the function under test to do so next month without thinking about it, not for a minute thinking that the mocks we use here are strangely behaved - I'm just trying to make it behave as much like the real method as possible. The mocks should try not to know too much about what the function under test does and doesn't do.

app/main/views/communications.py
risicle: Because I already feel like I've overcooked this story.

app/main/views/communications.py
risicle: Guess so

app/main/views/communications.py
```diff
@@ -43,6 +43,24 @@ def manage_communications(framework_slug):
...
 
 
+@main.route('/communications/<framework_slug>/files/<string:comm_type>/<path:filepath>', methods=['GET'])
+@role_required('admin-framework-manager')
+def download_communication(framework_slug, comm_type, filepath):
+    if comm_type not in _comm_types:
+        abort(404)
+
+    # ensure this is a real framework
+    data_api_client.get_framework(framework_slug)
```
katstevens: Could we use `app.helpers.frameworks.get_framework_or_404` here?

risicle: Oh yeah

lfdebrux:
I'm happy 👍

PR #530, opened 2019-07-09
Fix unapproved services to fallback to supplierId

lfdebrux:
https://trello.com/c/JSPtBsU5/933-bug-in-admin-frontend

Some audit events do not have supplierName but have supplierId. This PR changes the service approval flow to handle this case. As the ticket notes this is just a short term fix to get the admin frontend working again while we find out what's producing the bad audit events.


lfdebrux:
Added some tests

PR #535, opened 2019-08-07
find_user_by_email_address view: enable for admin-ccs-data-controller role

risicle:
Straightforward fix for https://trello.com/c/dBEA3HaN

Required adding this to the data controller's front page, so now it looks like this:

image


risicle:
Unfortunately travis doesn't agree.

PR #538, opened 2019-08-12
Strip whitespace from DUNS number, or default to empty string

jonodrew:
We noticed that users were getting errors when they copy-pasted DUNS numbers as whitespaces were being introduced. This change will strip out trailing and leading whitespaces if the user has input a DUNS, and leave it as None if not.

I've also added a test to check that it works

[Ticket](https://trello.com/c/wKXRk59y/98-strip-whitespace-from-admin-duns-searches)

### DIFF
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -65,9 +65,11 @@ def find_suppliers():
         suppliers = [data_api_client.get_supplier(request.args.get("supplier_id"))['suppliers']]
         links = {}
     else:
+        duns_number = request.args.get("supplier_duns_number").strip() \
+            if request.args.get("supplier_duns_number") else None
         suppliers_response = data_api_client.find_suppliers(
             name=request.args.get("supplier_name"),
-            duns_number=request.args.get("supplier_duns_number"),
+            duns_number=duns_number,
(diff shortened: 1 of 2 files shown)


app/main/views/suppliers.py
```diff
@@ -67,7 +67,7 @@ def find_suppliers():
     else:
         suppliers_response = data_api_client.find_suppliers(
             name=request.args.get("supplier_name"),
-            duns_number=request.args.get("supplier_duns_number"),
+            duns_number=request.args.get("supplier_duns_number", "").strip(),
```
katstevens: I think this might end up with unexpected behaviour from the API client. Passing through `duns_number=None` to the client means we ignore that parameter completely. Passing through `duns_number=""`... may lead to it filtering for blank DUNSes?

See https://github.com/alphagov/digitalmarketplace-apiclient/blob/master/dmapiclient/data.py#L108

jonodrew: Very good shout. I've made the suggested change

katstevens: Cool - are the functional tests happy?

jonodrew: Positively elated ^-^

PR #539, opened 2019-08-12
"view supplier declaration": show value of onFramework as "Application status"

risicle:
https://trello.com/c/w90E7y1T

Not exceptionally pretty, but works. `True`/`False`/`None` are transformed into "Pass" "Fail" and "Pending" respectively.

Spectacle X22828

### DIFF
--- a/app/templates/suppliers/view_declaration.html
+++ b/app/templates/suppliers/view_declaration.html
@@ -39,6 +39,15 @@
       {% endwith %}
     </div>
   </div>
+  {% call summary.mapping_table(
+    caption="Application status",
+    field_headings_visible=False
+  ) %}
+    {% call summary.row() %}
+      {{ summary.field_name("Application status") }}
+      {{ summary.text("Pass" if supplier_framework.onFramework else ("Fail" if supplier_framework.onFramework is sameas false else "Pending")) }}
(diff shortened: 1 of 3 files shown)


app/templates/suppliers/view_declaration.html
```diff
@@ -39,6 +39,15 @@
       {% endwith %}
     </div>
   </div>
+  {% call summary.mapping_table(
+    caption="Application status",
+    field_headings_visible=False
+  ) %}
+    {% call summary.row() %}
+      {{ summary.field_name("Application status") }}
+      {{ summary.text("Pass" if supplier_framework.onFramework else ("Fail" if supplier_framework.onFramework is sameas false else "Pending")) }}
```
katstevens: `is sameas` 

Ooer, do we use this anywhere else? 🤔

risicle: I dunno but it's the jinja equivalent of doing `is` - `is` on its own makes jinja do funny exotic things because someone decided to invent a new expression language instead of just making something that looks like python ... be python. Don't really have a choice here because we're tied up in funny macros.

risicle: Also note how `False` is `false`.

PR #541, opened 2019-08-20
view_supplier_declaration: enable view for "expired" frameworks 

risicle:
This started out as https://trello.com/c/R9x4f7te but as you can see from that ticket this morphed a bit.

What we're now doing is enabling the "view declaration" view for `expired` frameworks. This was previously prevented (which is what was causing the 403 mentioned in the ticket).

This also makes sure `supplierFramework`s which don't have a declaration don't even show in the table on the `supplier_details` page, as there's nothing useful that can be done with such `supplierFramework`s from here.

### DIFF
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -31,7 +31,8 @@ from ..helpers.countries import COUNTRY_TUPLE
 from ..helpers.pagination import get_nav_args_from_api_response_links
 from ..helpers.supplier_details import (
     get_supplier_frameworks_visible_for_role,
-    get_company_details_from_supplier
+    get_company_details_from_supplier,
+    DEPRECATED_FRAMEWORK_SLUGS,
 )
 from ... import data_api_client, content_loader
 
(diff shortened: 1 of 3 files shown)


app/main/views/suppliers.py
```diff
@@ -315,9 +316,12 @@ def edit_supplier_duns_number(supplier_id):
 @main.route('/suppliers/<int:supplier_id>/edit/declarations/<string:framework_slug>', methods=['GET'])
 @role_required('admin-ccs-sourcing')
 def view_supplier_declaration(supplier_id, framework_slug):
+    if framework_slug in DEPRECATED_FRAMEWORK_SLUGS:
+        abort(404)
```
katstevens: Yep, I think this is fine. If admins really want to see g-cloud-4 details they're probably better off looking at a bit of paper somewhere...

risicle:
That is a good question.

lfdebrux:
Something funky with the git history is causing issues with Travis https://www.scivision.dev/travis-ci-git-depth-reference-not-a-tree/

lfdebrux:
I'm not sure it's worth reconfiguring Travis for this, we could just ignore the status check

risicle:
Also happening for me on https://github.com/alphagov/digitalmarketplace-utils/pull/536

lfdebrux:
It fixed itself? 🤔

risicle:
:balloon:

PR #544, opened 2019-08-22
Show pass/fail status of declaration answers to Sourcing admins

katstevens:
https://trello.com/c/3AIJtJTX/224-highlight-wrong-answers-in-the-admin-declaration-view

declaration-pass-fail

Adds a column showing 'Pass', 'Fail', 'Discretionary' or 'Not Applicable' for each declaration answer. 'Not Applicable' means the answer is not assessed (for example, a supplier can't 'fail' on their contact email address).

Annoyingly the columns don't line up between sections. This is a known issue with this pattern anyway. However the Sourcing team users are not too worried about layout..

The Modern Slavery section contains a single 'multiquestion' type question with several nested questions - only one of these is assessed, so I haven't been particularly strict on how multiple results would be displayed here. Suggestions on a less messy Jinja approach very welcome.

### DIFF
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -332,13 +332,18 @@ def view_supplier_declaration(supplier_id, framework_slug):
 
     content = content_loader.get_manifest(framework_slug, 'declaration').filter(sf.get("declaration", {}))
     declaration_sections = content.sections
-    question_numbers = OrderedDict(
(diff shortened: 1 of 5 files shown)


risicle:
As far as the jinja stuff goes, yeah it's a bit weird having the logic in jinja and we should probably make `dmcontent` able to do this for us. But for now it will probably do.

PR #545, opened 2019-08-23
Bump digitalmarketplace-utils dependency to 48.5.0

risicle:
https://trello.com/c/HsZ9mj7o

Also modify some tests to cope with newly set "probe cookie" for detecting cookie-disabled browsers


risicle:
Thanks

PR #548, opened 2019-09-03
set DM_COOKIE_PROBE_COOKIE_EXPECT_PRESENT, enabling cookie error page

risicle:
https://trello.com/c/HsZ9mj7o

This should only ever show itself in place of a CSRF token failure, so should be a fairly safe thing to enable.


benvand:
@risicle rebased this just to see what the merge options are

PR #549, opened 2019-09-05
Remove Yarn in favour of npm

aliuk2012:
We do not need to depend on Yarn as npm is Nodes default package manager
and we use npm to install yarn. npm does everything yarn does now.

Co-authored-by: Laurence de Bruxelles <laurence.debruxelles@digital.cabinet-office.gov.uk>

Trello ticket: https://trello.com/c/RrpjQDJi


lfdebrux:
I just realised this breaks our Dockerfile...

PR #555, opened 2019-09-09
Using GOV.UK Frontend buttons

aliuk2012:
## Noticeable differences.
- Search buttons are smaller but we don't think this will be an issue or is important to ensure they are the same size. 

- Previously "buttons" were actually `<input type="submit">` but now they are actually `<button>`. 
`<button>` is newer than `<input type="submit">`, is more semantic, easy to stylize and support HTML inside of it. There should not be any other behaviour changes because `<button>` has a `type="submit"` by default.

Trello ticket: https://trello.com/c/nsfgRLBj/289-replace-toolkit-button-with-govukbutton-in-admin-frontend

### DIFF
--- a/app/templates/_base_page.html
+++ b/app/templates/_base_page.html
@@ -1,5 +1,8 @@
 {% extends "toolkit/layouts/_base_page.html" %}
 
+{# Import GOV.UK Frontend components for globale usage#}
+{% from "components/button/macro.njk" import govukButton %}
+
 {% block footer_top %}
 {% endblock %}
 {% block footer_support_links %}
--- a/app/templates/invite_admin_user.html
+++ b/app/templates/invite_admin_user.html
@@ -34,14 +34,9 @@
         {{ form.email_address }}
         {{ form.role }}
         {% block save_button %}
-          {%
-            with
-            label="Invite user",
-            name="submit",
-            type="save"
-          %}
-            {% include "toolkit/button.html" %}
-          {% endwith %}
+          {{ govukButton({
+            "text": "Invite user"
+          }) }}
         {% endblock %}
       </form>
     </div>
--- a/app/templates/view_users.html
+++ b/app/templates/view_users.html
@@ -25,123 +25,112 @@
   {% endwith %}
 
   <form action="{{ url_for('.find_user_by_email_address') }}" method="get" class="question">
-  {%
-    with
-    question = "Find a user by email",
-    name = "email_address"
-  %}
-    {% include "toolkit/forms/textbox.html" %}
-  {% endwith %}
+    {%
+      with
+      question = "Find a user by email",
+      name = "email_address"
+    %}
+      {% include "toolkit/forms/textbox.html" %}
+    {% endwith %}
 
-  {{ govukButton({
-    "text": "Search"
-  }) }}
+    {{ govukButton({
+      "text": "Search"
+    }) }}
 
   </form>
 
   {% call(item) summary.list_table(
-          users,
-          caption="Users",
-          empty_message="No users to show",
-          field_headings=[
-              'Name',
-              'Role',
-              'Supplier',
-              'Last login',
-              'Last password change',
-              'Locked',
-              summary.hidden_field_heading("Change status")
-          ],
-          field_headings_visible=True)
-      %}
-      {% call summary.row() %}
+    users,
+    caption="Users",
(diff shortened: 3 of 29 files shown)


lfdebrux:
We're going to need to merge https://github.com/alphagov/digitalmarketplace-functional-tests/pull/690/ to get this to pass functional tests

app/templates/_base_page.html
```diff
@@ -1,5 +1,8 @@
 {% extends "toolkit/layouts/_base_page.html" %}
 
+{# Import GOV.UK Frontend components for globale usage#}
+{% from "components/button/macro.njk" import govukButton %}
```
katstevens: Are we going to have to import every component we use separately?

aliuk2012: Yes, this pattern is used in GOV.UK Prototype Kit(https://github.com/alphagov/govuk-prototype-kit/blob/master/app/views/layout.html#L4-L30). Ideally, it would have been nice to have these imports in another file that we could just use a single import but we had difficulties with jinja recognizing the imported macros in a separate file(Happy to talk through it if you can think of a way to import it). 

Jon's spike in supplier and GOV.UK Pays implementation all import on a page by page basis. All though this is nice to see which components are used and how often, it does cause an issue for us when we migrate from govuk-frontend 2.0 and 3.0 and 3.0 has a different folder structure so we would need to do a global find and replace. You also have to know and remove which components need to import and the syntax to use. 

There is an [outstanding issue in govuk-frontend](https://github.com/alphagov/govuk-frontend/issues/1278) asking for the package to include an "all.njk" in the same way there is a way to import all css components and all javascript.

app/templates/view_users.html
```diff
@@ -25,132 +25,112 @@
...
+            </form>
+          {% else %}
+            Deactivated
+          {% endif %}
+        {% endif %}
       {% endcall %}
 
-      {% endcall %}
-      {% endcall %}
-  </div>
```
aliuk2012: I found this stray closing div tag after sorting out the indentation problems in this code.

app/templates/_base_page.html
katstevens: Makes sense, thanks for the detailed explanation! I wonder why the macro import doesn't work... 🤔 Did you namespace it in the same way as the existing toolkit imports? eg 

```
{% import 'toolkit/my-macro-file.html' as foo %}
...
{% call foo.macro_name(args) %}
```

lfdebrux: @katstevens I think the problem is that you can't include a macro, you can only import it; ideally we would have an `all.njk` that lets you do this:

```
{% include 'components/all.njk' %}

{{ govukButton(params) }}
```

This would be very similar to what the examples are on the GOV.UK Design System. However, this doesn't work; the macros don't get included. It does work if you redefine the macros and namespace them, but then we can't use examples from the GOV.UK Design System.

app/templates/view_users.html
lfdebrux: Nice spot, let's get rid of it 🔥

app/templates/_base_page.html
lfdebrux: Also, we might decide to spin out a new base template that can be shared between apps that imports all macros; that's probably a future problem though.

katstevens: Ahh I see. OK, having the imports in the `_base_page` should be fine for now, anyway.

aliuk2012:
We will need to make sure that the [function-tests](https://github.com/alphagov/digitalmarketplace-functional-tests) pass with this branch before merging https://github.com/alphagov/digitalmarketplace-functional-tests

lfdebrux:
Another functional test change that this PR needs https://github.com/alphagov/digitalmarketplace-functional-tests/pull/693

app/templates/invite_admin_user.html
```diff
@@ -34,14 +34,10 @@
...
-          {%
-            with
-            label="Invite user",
-            name="submit",
-            type="save"
-          %}
-            {% include "toolkit/button.html" %}
-          {% endwith %}
+          {{ govukButton({
+            "name": "name",
```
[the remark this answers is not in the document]
aliuk2012: Good spot 👍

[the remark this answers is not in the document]
aliuk2012: Fixed.

PR #558, opened 2019-09-10
Using GOV.UK Frontend breadcrumbs

aliuk2012:
The biggest thing to look out for as part of code review is that the label
and href match what was previously used on the page.

breadcrumbs block will not be available once we migrate from govuk_template
to govuk-frontend template. We will need to move these inside a beforeContent block but for the moment we will need to place it inside `main` so that breadcrumbs are scoped to the width of the page

See:
- http://design-system.service.gov.uk/styles/layout/#limiting-width-of-content
- http://design-system.service.gov.uk/styles/page-template

Trello card: https://trello.com/c/9KcXEsfF

### DIFF
--- a/app/templates/suppliers/view_signed_agreement.html
+++ b/app/templates/suppliers/view_signed_agreement.html
@@ -14,6 +14,9 @@
       {
         "text": "{} agreements".format(framework.name),
         "href": url_for('.list_agreements', framework_slug=framework.slug)
+      },
+      {
+        "text": "Countersign framework agreements"
       }
     ]
   }) }}
--- a/app/templates/suppliers/view_signed_agreement.html
+++ b/app/templates/suppliers/view_signed_agreement.html
@@ -4,24 +4,20 @@
   Countersign {{ framework.name }} agreement for {{ supplier_framework.declaration.nameOfOrganisation }} – Digital Marketplace admin
 {% endblock %}
 
-{% block breadcrumb %}
-    {%
-    with items = [
-        {
-            "link": url_for('.index'),
-            "label": "Admin home"
-        },
-        {
-            "link": url_for('.list_agreements', framework_slug=framework.slug),
-            "label": "%s agreements" | format(framework.name)
-        },
-        ]
-    %}
-        {% include "toolkit/breadcrumb.html" %}
-    {% endwith %}
-{% endblock %}
-
 {% block main_content %}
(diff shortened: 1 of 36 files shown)


aliuk2012:
#558 needs to be merged first and then this PR can be rebased off the master branch again

risicle:
> #558 needs to be merged first and then this PR can be rebased off the master branch again

This comment had me go round in circles about 5 times "how many of these has he created??"

risicle:
Looks like we need to fix more tests on the nunjucks importer before travis will pass on this...

aliuk2012:
> > #558 needs to be merged first and then this PR can be rebased off the master branch again
> 
> This comment had me go round in circles about 5 times "how many of these has he created??"

Sorry, this is a WIP'd PR ;)

lfdebrux:
I've got the unit tests passing, however I haven't tried this with functional tests yet

app/templates/suppliers/view_signed_agreement.html
```diff
@@ -4,21 +4,22 @@
...
-    {% endwith %}
+{% block beforeContent %}
+  {{ govukBreadcrumbs({
+    "items": [
+      {
+        "text": "Admin home",
+        "href": url_for('.index')
+      },
+      {
+        "text": "%s agreements" | format(framework.name),
```
katstevens: I know this was pre-existing, but this seems like a good opportunity to be consistent in how we do our format syntax - in the file above we use "{} declaration".format(framework.name) which is more immediately recognisable for us Python folks 😸 Not a blocker however.

app/templates/suppliers/view_signed_agreement.html
```diff
@@ -4,24 +4,20 @@
...
-
 {% block main_content %}
+  {{ govukBreadcrumbs({
+    "items": [
+      {
+        "text": "Admin home",
+        "href": url_for('.index')
+      },
+      {
+        "text": "%s agreements".format(framework.name),
```
katstevens: Ah, this will need to be something like `"{} agreements".format(framework.name)'`. Otherwise we'll see something like this:

format-string-breadcrumbs

lfdebrux: Nice spot 🔍

lfdebrux: I'll fix this now

PR #560, opened 2019-09-13
🚀  Replacing govuk_template    ☠️

aliuk2012:
This PR includes some major changes to the application and performance benefits.

**Prep**
- Ensure all elements are using the majority of govuk-frontend equivalent classes
 
**Things that I had to look out for and consider:**
- Header component - We have decided to create a new Digital Marketplace GOV.UK Frontend header component and this PR implements this component for the first time.
- Existing meta tags and other functionality that was present in govuk_template and digitalmarketplace-frontend-toolkit but not in GOV.UK Frontend.
- Cookie banner, This will need to be a new custom component. I will do this as a separate PR and possibly something that needs to be shared across all our frontend applications

## Performance impact:

|                           | Current | New   | % difference  |
|---------------------------|---------|-------|---------------|
| Requests                  | 17      | 10    | 41% reduction | 
| Transferred (Gzip)        | 602KB   | 349KB | 42% reduction |
| Transferred (Uncompressed)| 1.1MB   | 812KB | 26% reduction |

### Current
Screen Shot 2019-10-01 at 20 20 27

### New
Screen Shot 2019-10-01 at 20 21 54

ticket: https://trello.com/c/tOseXpgE/325-replace-govuk-elements-template-with-govuk-frontend-template-in-admin-frontend


app/templates/_base_page.html
```diff
@@ -1,48 +1,73 @@
...
+              <input type="hidden" name="csrf_token" value="' + (csrf_token_value or csrf_token()) + '" />
+            </form>',
+      'href': url_for('external.render_login')
+    }]%}
+  {% else %}
+    {% set headerNavigation = headerNavigation + [{
+      'text': 'Log in',
+      'href': url_for('external.render_login')
+    }]%}
+  {% endif %}
```
aliuk2012: Is this even needed for admin frontend? Users can only use Admin frontend if they are logged in else they are redirected to user-frontend to authenticate.

app/templates/_base_page.html
```diff
@@ -1,48 +1,73 @@
...
-      {% block main_content %}
-      {% endblock %}
-    </main>
-  </div>
+{% block header %}
+
+  {% set headerNavigation = [
+    {
+      'text': 'Guidance',
+      'href': 'https://www.gov.uk/guidance/buying-and-selling-on-the-digital-marketplace'
```
aliuk2012: Do we have a `url_for` equivalent?

katstevens: We don't have these for external (including GOV.UK) links. 

The list of Flask views that are available across all the frontend apps is kept in utils: https://github.com/alphagov/digitalmarketplace-utils/blob/master/dmutils/external.py

lfdebrux: Might be an interesting tech debt ticket to add views to external that redirect to commonly used GOV.UK links; keep them all in one place and all.

katstevens: FYI we have a mix of framework-specific URLs (buried in the `frameworks` repo with the rest of the content, in theory easier to manage but certainly less visible when debugging) and framework-agnostic ones (which usually just get plonked in the templates, so easier to find but easier to forget about).

app/templates/_base_page.html
aliuk2012: No theres no reason for us to check if the user is logged in or not.

lfdebrux:
I'm a bit worried about all the stuff we need to pull in from dm-toolkits to make this PR happen... could we instead move them to digitalmarketplace-govuk-frontend? Or is that not the correct place for that sort of thing?

aliuk2012:
@lfdebrux good call. I think we could move some of these changes into digitalmarketplace-govuk-frontend

app/templates/index.html
```diff
@@ -2,12 +2,8 @@
...
 {% block page_title %}Digital Marketplace admin{% endblock %}
 
-{% block main_content %}
-<div class="govuk-grid-row">
-  <div class="govuk-grid-column-two-thirds">
-    <h1 class="govuk-heading-xl">Admin</h1>
-  </div>
-</div>
+{% block content %}
+<h1 class="govuk-heading-xl">Admin</h1>
```
lfdebrux: Why does this heading get moved out of the div?

app/templates/_base_page.html
```diff
@@ -6,33 +6,50 @@
 {% from "components/phase-banner/macro.njk" import govukPhaseBanner %}
 {% from "components/tabs/macro.njk" import govukTabs %}
 
+{% set assetPath = '/admin/static' %}
+
 {% block head %}
+  {% include "layouts/_custom_dimensions.html" %}
+  {% include "layouts/_site_verification.html" %}
```
lfdebrux: Can we split these changes (adding assetPath and adding custom dimensions / site verification) to separate commits?

app/templates/index.html
aliuk2012: There is no reason for it to be in a div (aka grid). The content is short and concise and it would display exactly the same no matter what device.

app/assets/scss/application.scss
```diff
@@ -1,7 +1,3 @@
-// Reset styles on core HTML elements that might be injected by eg a user's browser, so that we have a known clean base
-// for DM styles to build on.
-@import "toolkit/shared_scss/_reset.scss";
```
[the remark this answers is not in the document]
aliuk2012: No, we are not replacing resets. 

Resets are a bit of an anti-pattern when using design systems and components. 

This is what happens when resets are used:

1. tell the browser to remove any default stylings to all elements on a page. 
2. Tell the browser to style all the components using our classes.

With the design system, all components/elements have class names. The components are self-contained and styles do not bleed out into other components that might use the same html elements. It's more efficient to not reset and just go straight to the second instruction because the browser doesn't have to paint the page as part of the first reset and then repaint it again with all the new stylings.

The resets are also the cause for a lot of the custom classes/css introduced to our frontend applications to add spacing around elements.

Happy to chat about it more in detail if needed. 👍

config.py
```diff
@@ -60,6 +60,13 @@ def init_app(app):
         template_folders = [
             os.path.join(repo_root, 'app', 'templates'),
             os.path.join(repo_root, 'node_modules', 'digitalmarketplace-govuk-frontend', 'govuk-frontend'),
+            os.path.join(
+                repo_root,
+                'node_modules',
+                'digitalmarketplace-govuk-frontend',
+                'digitalmarketplace',
```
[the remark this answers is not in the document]
aliuk2012: it is coming from digitalmarketplace-govuk-frontend v0.2.0. 

Ideally, we should just do the following and use the folders to namespace our components.

Eg.
`os.path.join(repo_root, 'node_modules', 'digitalmarketplace-govuk-frontend'`

```
{% from "govuk/components/breadcrumbs/macro.njk" import govukBreadcrumbs %}
{% include "digitalmarketplace/templates/layouts/_custom_dimensions.html" %}
```

I think for the time being we should leave it as it is because v3 of govuk-frontend will require us to revisit this issue and update it to a single entry-point.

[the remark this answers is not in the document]
katstevens: I think it might need an explanatory comment. We obviously have a lot of things called `digitalmarketplace`, it might not be immediately obvious which one this refers to 😹 

Perhaps pulling out the list of template names into a constant would be useful?

```
FRONTEND_TEMPLATE_FOLDERS = [
    'node_modules', 
    'digitalmarketplace',   # Explain what this is here
    'digitalmarketplace-govuk-frontend',
    'templates'
]

os.path.join(repo_root, **FRONTEND_TEMPLATE_FOLDERS)
```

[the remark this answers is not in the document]
katstevens: Just writing all that out gave me a better understanding of what's going on 😹

app/assets/scss/application.scss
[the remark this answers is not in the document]
katstevens: Does this include the `body` element? I see that `_reset` is applying the `@include core-19` (typography, I think?) style there.

app/templates/edit_section.html
```diff
@@ -3,9 +3,9 @@
 
 {% extends "_base_page.html" %}
 
-{% set page_title = "Edit {}".format(section.name) %}
+{% set pageTitle = "Edit {}".format(section.name) %}
```
katstevens: This breaks the breadcrumbs, which is looking for `page_title`

katstevens: For the other apps, will it be possible to split out the pageTitle updates into their own tickets so we have greater visibility of this? There are a lot of changes in this PR to check...

aliuk2012: we could it just means more temporary workarounds to add before taking them out at a later stage. I do understand your point and I think going forward we should be able to break this out into another PR

app/assets/scss/application.scss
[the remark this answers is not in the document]
aliuk2012: We shouldn't be stying `body` directly. All components should have their own font sizes set using their respective classes eg. (`<p>` using `govuk-body` and `<a>` using `govuk-links` etc)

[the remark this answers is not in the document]
katstevens: Cool, as long as that's covered elsewhere 👍

aliuk2012:
I've created two more PR(#588 & #589), to help reduce the number of files changed in this PR. These two can be replicated across all the other frontend repos independently.

app/templates/_base_page.html
```diff
@@ -45,5 +47,7 @@
   {% block breadcrumbs %}{% endblock%}
   {% include "toolkit/flash_messages.html" %}
 {% endblock %}
-{% block footer_support_links %}
+
+{% block bodyEnd %}
+  <script src="{{assetPath }}/javascripts/application.js"></script>
```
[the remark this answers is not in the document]
aliuk2012: Good catch, fixed this now.

lfdebrux:
Visual regression and functional tests look good to me

PR #561, opened 2019-09-13
toggle_supplier_services view: don't wait_for_index in update_service_status() call

risicle:
This final piece of https://trello.com/c/LABe8BKA is simple enough, depending on https://github.com/alphagov/digitalmarketplace-apiclient/pull/223 and https://github.com/alphagov/digitalmarketplace-api/pull/968


lfdebrux:
I think the ticket mentioned something about a flash message, did you decide against that in the end?

risicle:
Oh well caught - I'll look into that on monday.

risicle:
Ok flash message update has been made. Do we think the language is clear enough?

PR #562, opened 2019-09-15
Replacing toolkit/page-headings with govuk frontend headings

aliuk2012:
This replaces toolkit/page-headings with govuk frontend page titles.

Generally, the headings were `h1` with a css class of `govuk-heading-xl`.
These were the mappings I used while replacing

Generally, use `govuk-heading-xl` class
If "smaller : true" when calling toolkit/page-heading, is 'h1` with
`govuk-heading-l`.
If "Context" used when calling toolkit/page-heading, add span above the
`h1` and use govuk-caption-xl or govuk-caption-l depending on the heading
used.

while replacing the headings I noticed that there is an inconsistency with
the markup where sometimes the headings are placed inside a grid and others
aren't. I also found a difference between heading titles font sizes, it
would be good to use a consistent size. I will try address these when
the grid system is replaced.

Trello card: https://trello.com/c/jjTqegdr

### DIFF
--- a/app/templates/download_framework_users.html
+++ b/app/templates/download_framework_users.html
@@ -21,11 +21,7 @@
 {% block main_content %}
   <div class="grid-row">
     <div class="column-two-thirds">
-      {%
-        with heading = "Download supplier lists for {}".format(framework['name'])
-      %}
-        {% include "toolkit/page-heading.html" %}
-      {% endwith %}
+      <h1 class="govuk-heading-xl">Download supplier lists for {{ framework['name'] }}</h1>
 
       <div class="explanation-list">
           <p class="lede">These lists contain details of all suppliers who:</p>
--- a/app/templates/suppliers/edit_registered_address.html
+++ b/app/templates/suppliers/edit_registered_address.html
@@ -28,11 +28,8 @@
 {% endblock %}
 
 {% block main_content %}
-
   {% include 'toolkit/forms/validation.html' %}
-  {% with heading = "Update registered company address for ‘{}’".format(supplier.name) %}
-    {% include "toolkit/page-heading.html" %}
-  {% endwith %}
+  <h1 class="govuk-heading-xl">Update registered company address for ‘{{supplier.name}}’</h1>
   <form method="post">
     <div class="grid-row">
       <div class="column-two-thirds">
--- a/app/templates/suppliers/edit_registered_name.html
+++ b/app/templates/suppliers/edit_registered_name.html
@@ -24,9 +24,7 @@
 
 {% block main_content %}
   {% include 'toolkit/forms/validation.html' %}
-  {% with heading = "Update registered company name for ‘{}’".format(supplier.name) %}
-    {% include "toolkit/page-heading.html" %}
-  {% endwith %}
+  <h1 class="govuk-heading-xl">Update registered company name for ‘{{supplier.name}}’</h1>
   <form method="post">
     <div class="grid-row">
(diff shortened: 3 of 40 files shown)


katstevens:
Are we bothered at all about losing the semantic `<header>` markup? Visually it just adds a bit of space, but I don't know if will affect eg screenreader users.

lfdebrux:
> Are we bothered at all about losing the semantic `<header>` markup?

Currently we have two `<header>` elements on our pages; from our [page header](http://alphagov.github.io/digitalmarketplace-frontend-toolkit/proposition-header.html) component and our [headings](http://alphagov.github.io/digitalmarketplace-frontend-toolkit/page-heading.html) components.

It's not clear to me from the [HTML standard](https://html.spec.whatwg.org/multipage/sections.html#the-header-element) whether it's okay to have two, but for me it seems confusing.

There doesn't seem to be much guidance about this on the internet generally, but looking at the GOV.UK Design System's [page template](https://design-system.service.gov.uk/styles/page-template/) they have only one header, which is the [GOV.UK header component](https://design-system.service.gov.uk/components/header/).

I just did a bit of reading into this, Mozilla doesn't have much guidance. `<header>`

katstevens:
Happy for us to align with the pattern the Design System site uses 👍

app/templates/download_framework_users.html
```diff
@@ -21,11 +21,7 @@
 {% block main_content %}
   <div class="grid-row">
     <div class="column-two-thirds">
-      {%
-        with heading = "Download supplier lists for {}".format(framework['name'])
-      %}
-        {% include "toolkit/page-heading.html" %}
-      {% endwith %}
+      <h1 class="govuk-heading-xl">{{ "Download supplier lists for {}".format(framework['name']) }}</h1>
```
katstevens: We can simplify all these `.format()` calls to:
 `<h1 class="govuk-heading-xl">Download supplier lists for {{ framework['name']) }}</h1>`

risicle: It's amazing being able to use our templating language as a templating language again.

app/templates/suppliers/edit_registered_address.html
```diff
@@ -28,11 +28,8 @@
 {% endblock %}
 
 {% block main_content %}
-
   {% include 'toolkit/forms/validation.html' %}
-  {% with heading = "Update registered company address for ‘{}’".format(supplier.name) %}
-    {% include "toolkit/page-heading.html" %}
-  {% endwith %}
+  <h1 class="govuk-heading-xl">{{ "Update registered company address for ‘{}’".format(supplier.name) }}</h1>
```
risicle: This too can just be `<h1 class="govuk-heading-xl">Update registered company address for ‘{{ supplier.name }}’</h1>`

app/templates/suppliers/edit_registered_name.html
```diff
@@ -24,9 +24,7 @@
 
 {% block main_content %}
   {% include 'toolkit/forms/validation.html' %}
-  {% with heading = "Update registered company name for ‘{}’".format(supplier.name) %}
-    {% include "toolkit/page-heading.html" %}
-  {% endwith %}
+  <h1 class="govuk-heading-xl">{{ "Update registered company name for ‘{}’".format(supplier.name) }}</h1>
```
risicle: Similar

lfdebrux:
@katstevens I think I've addressed your requested changes

PR #564, opened 2019-09-19
Replace user research banner

aliuk2012:
[Introduces GOV.UK Frontend phase banner component](https://design-system.service.gov.uk/components/phase-banner)

While we were trying to migrate from govuk_template to govuk-frontend
template. An issue was identified with having the user research sign up
banner positioned between the top black header and the colour. It would mean that we could not use GOV.UK Frontend header component
and we would still have to maintain the user research banner and it also
relied on govuk_templates cookie implementation which we would also need
to either copy or remove completely. As this was the only component using
GOVUK.cookie method it would be good to remove it.

This change also brings our designs closer inline with GOV.UK Design System and their recommendation:

> Your banner must be directly under the black GOV.UK header and colour bar.

### Before
image

### After
image (2)

---

Trello: https://trello.com/c/8oD7F688/309-replace-user-research-consent-banner-with-phase-banner-in-the-admin-frontend

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -16,7 +16,6 @@
 //= require ../../../node_modules/govuk_frontend_toolkit/javascripts/govuk/shim-links-with-button-role.js
 //= require ../../../node_modules/digitalmarketplace-frontend-toolkit/toolkit/javascripts/shim-links-with-button-role.js
 //= require ../../../node_modules/digitalmarketplace-frontend-toolkit/toolkit/javascripts/show-hide-content.js
-//= require ../../../node_modules/digitalmarketplace-frontend-toolkit/toolkit/javascripts/user-research-consent-banner.js
 //= require ../../../node_modules/digitalmarketplace-frontend-toolkit/toolkit/javascripts/report-a-problem.js
 //= require ../../../node_modules/govuk-frontend/all.js
 //= require _scroll-through-statistics.js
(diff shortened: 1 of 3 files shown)


app/templates/_base_page.html
```diff
@@ -3,12 +3,35 @@
...
 {# override content block to add a beforeContent block like the one in govuk-frontend template #}
 {% block content %}
   {% block top_header %}{% endblock %}
   <div id="wrapper">
+    {# This is only a temporary place for the phase banner until we start using govuk-frontend template #}
+    {{ govukPhaseBanner({
+      "tag": {
+        "text": "beta"
+      },
+      "html": 'Help us improve the Digital Marketplace - <a class="govuk-link" href="https://www.digitalmarketplace.service.gov.uk/help">send your feedback</a>'
```
katstevens: Can we use the `url_for()` here instead of the hardcoded link - See https://github.com/alphagov/digitalmarketplace-frontend-toolkit/blob/master/toolkit/templates/proposition-header.html#L10 

This way it will work for preview and staging as well.

app/assets/javascripts/application.js
```diff
@@ -16,7 +16,7 @@
 //= require ../../../node_modules/govuk_frontend_toolkit/javascripts/govuk/shim-links-with-button-role.js
 //= require ../../../node_modules/digitalmarketplace-frontend-toolkit/toolkit/javascripts/shim-links-with-button-role.js
 //= require ../../../node_modules/digitalmarketplace-frontend-toolkit/toolkit/javascripts/show-hide-content.js
-//= require ../../../node_modules/digitalmarketplace-frontend-toolkit/toolkit/javascripts/report-a-problem.js
+; // Hack needed because show-hide-content.js does not have a closing ";"
```
katstevens: I think we left it in 'just in case' we wanted to turn it back on again - I think it's been long enough now that we may as well remove it. It's in the history if we ever need it 😸

risicle:
Looks much nicer

app/templates/_base_page.html
aliuk2012: Thanks @katstevens, I'll keep that in mind for any future urls.

PR #566, opened 2019-09-23
Refactor CSS to bring it more inline with GOV.UK Frontend

aliuk2012:
- Removes two unnecessary css modifiers
- Start using govuk-font mix 
- [Start using govuk-frontend colour palette](https://design-system.service.gov.uk/styles/colour/). **(Colour palette did change slightly between v2 and v3 of govuk-frontend so I've had to still use v2 colour palette for the time being)**
- [Start using govuk-frontend spacing scale](https://design-system.service.gov.uk/styles/spacing/#spacing-on-custom-components)

### DIFF
--- a/app/assets/scss/application.scss
+++ b/app/assets/scss/application.scss
@@ -113,7 +113,7 @@ $govuk-compatibility-govukelements: true;
 }
 
 .location-autocomplete-fallback {
-  @include core-19;
+  @include govuk-font(19);
   width: 100%;
   height: 42px;
 }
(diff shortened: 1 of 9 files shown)


app/assets/scss/application.scss
```diff
@@ -62,23 +61,15 @@ $govuk-compatibility-govukelements: true;
...
   width: 100%;
-  border: 5px solid #dee0e2;
-}
-
-.lead {
-  margin-bottom: 20px;
+  border: govuk-spacing(1) solid #dee0e2;
 }
 
 .search-summary-border-bottom {
```
lfdebrux: Can we rename these styles to `.dmp-search-summary-border-bottom` or something like that?

aliuk2012: We can but I don't think now is the right time to do it because they might not be needed once we using govuk-frontend template. 

I have made a note to revisit the class names once we have migrated over.

PR #570, opened 2019-09-30
Move govukBreadcrumbs into its own block

aliuk2012:
This commit is needed as part of prepping for govuk-template as we will
need to use the phase banner and breadcrumbs inside the "beforeContent".

The way I implemented this means we do not have to add `super()` to each pages "beforeContent" block just to inherit the phase banner from base_page template.

### DIFF
--- a/app/templates/_base_page.html
+++ b/app/templates/_base_page.html
@@ -24,15 +24,16 @@
 {% block content %}
   {% block top_header %}{% endblock %}
   <div id="wrapper">
-    {# This is only a temporary place for the phase banner until we start using govuk-frontend template #}
-    {{ govukPhaseBanner({
-      "tag": {
-        "text": "beta"
-      },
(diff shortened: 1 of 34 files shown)


app/templates/_base_page.html
```diff
@@ -25,14 +25,16 @@
...
-      "html": 'Help us improve the Digital Marketplace - <a class="govuk-link" href="'  + url_for('external.help') + '">send your feedback</a>'
-    }) }}
+    {% block beforeContent %}
+      {{ govukPhaseBanner({
+        "tag": {
+          "text": "beta"
+        },
+        "html": 'Help us improve the Digital Marketplace - <a class="govuk-link" href="'  + url_for('external.help') + '">send your feedback</a>'
+      }) }}
+      {% block breadCrumbs %}{% endblock %}
```
lfdebrux: I think `breadcrumbs` should be one word like in the name of the macro, just to be consistent

aliuk2012: 🤦‍♂ good spot

PR #572, opened 2019-10-01
Applying styles to lists and refactoring lists of links

aliuk2012:
- All `<ul>` should use `govuk-list` class (https://design-system.service.gov.uk/styles/typography/#lists)

- Refactored the lists of links on the admin home page
As the links on the home page, are categorized. It feels more semantic
to declare them in an unordered list. This will help screen readers understand
how the links are related. I.e they are not just a group of random links
with no context.

### DIFF
--- a/app/templates/download_framework_users.html
+++ b/app/templates/download_framework_users.html
@@ -24,13 +24,13 @@
       <h1 class="govuk-heading-xl">Download supplier lists for {{ framework['name'] }}</h1>
 
       <div class="explanation-list">
-          <p class="govuk-body">These lists contain details of all suppliers who:</p>
-          <ul class='list-bullet'>
-            <li>started an application</li>
-            <li>are on the framework</li>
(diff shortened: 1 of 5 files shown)


app/templates/download_framework_users.html
```diff
@@ -24,13 +24,13 @@
...
       <div class="explanation-list">
-          <p class="govuk-body">These lists contain details of all suppliers who:</p>
-          <ul class='list-bullet'>
-            <li>started an application</li>
-            <li>are on the framework</li>
-            <li>weren't accepted on the framework</li>
-            <li>have been excluded from the framework</li>
-          </ul>
+        <p class="govuk-body">These lists contain details of all suppliers who:</p>
+        <ul class='govuk-list govuk-list--bullet'>
```
lfdebrux: Can we change these to double quotes to avoid the problem with searching for them in future?
```suggestion
        <ul class="govuk-list govuk-list--bullet">
```

PR #573, opened 2019-10-10
Change from govuk-frontend to digitalmarketplace-govuk-frontend

lfdebrux:
digitalmarketplace-govuk-frontend is out in the wild! :tada:

This PR changes the admin frontend to use the new NPM package instead of govuk-frontend.


gulpfile.js
```diff
@@ -14,7 +14,7 @@ const npmRoot = path.join(repoRoot, 'node_modules')
 const govukCountryPickerDist = path.join(npmRoot, 'govuk-country-and-territory-autocomplete', 'dist')
 const govukToolkitRoot = path.join(npmRoot, 'govuk_frontend_toolkit')
 const govukElementsRoot = path.join(npmRoot, 'govuk-elements-sass')
-const govukFrontendRoot = path.join(npmRoot, 'govuk-frontend')
+const govukFrontendRoot = path.join(npmRoot, 'digitalmarketplace-govuk-frontend/govuk-frontend')
```
aliuk2012: We should use the power of path.join and not rely on '/' directory seperator
```suggestion
const govukFrontendRoot = path.join(npmRoot, 'digitalmarketplace-govuk-frontend', 'govuk-frontend')
```

config.py
```diff
@@ -59,7 +59,7 @@ def init_app(app):
         repo_root = os.path.abspath(os.path.dirname(__file__))
         template_folders = [
             os.path.join(repo_root, 'app/templates'),
-            os.path.join(repo_root, 'node_modules/govuk-frontend'),
+            os.path.join(repo_root, 'node_modules/digitalmarketplace-govuk-frontend/govuk-frontend'),
```
aliuk2012:
```suggestion
            os.path.join(repo_root, 'node_modules','digitalmarketplace-govuk-frontend','govuk-frontend'),
```

config.py
```diff
@@ -59,7 +59,7 @@ def init_app(app):
         repo_root = os.path.abspath(os.path.dirname(__file__))
         template_folders = [
             os.path.join(repo_root, 'app/templates'),
```
aliuk2012:
```suggestion
            os.path.join(repo_root, 'app','templates'),
```

PR #577, opened 2019-10-18
Fixes up syntax and adds missing govuk-frontend classes

aliuk2012:
I picked up on this issue while I was working on https://github.com/alphagov/digitalmarketplace-admin-frontend/pull/560

### DIFF
--- a/app/templates/suppliers/upload_countersigned_agreement.html
+++ b/app/templates/suppliers/upload_countersigned_agreement.html
@@ -22,7 +22,7 @@
 
 {% block main_content %}
   {% if remove_countersigned_agreement_confirm %}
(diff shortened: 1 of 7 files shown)


app/templates/suppliers/upload_countersigned_agreement.html
```diff
@@ -22,7 +22,7 @@
 
 {% block main_content %}
   {% if remove_countersigned_agreement_confirm %}
-      <form method="post" action='{{ url_for(".remove_countersigned_agreement_file", supplier_id=supplier.id, framework_slug=framework.slug) }}'>
+      <form method="post" action="{{ url_for(".remove_countersigned_agreement_file", supplier_id=supplier.id, framework_slug=framework.slug) }}">
```
katstevens: If we are going to do this everywhere in future, can we swap the inner quotes as well please? As in, `url_for ('.remove_countersigned_agreement_file', ...`

PR #581, opened 2019-10-25
Adding vertical spacing to notifications

aliuk2012:
There is a current issue with a vertical spacing of notifications. This
commit adds it temporarily but ideally, the component itself should be
adding the margin spacing it requires.

Ticket: https://trello.com/c/9H3ADekE


app/templates/_base_page.html
```diff
@@ -32,10 +32,9 @@
         "html": 'Help us improve the Digital Marketplace - <a class="govuk-link" href="'  + url_for('external.help') + '">send your feedback</a>'
       }) }}
       {% block breadcrumbs %}{% endblock %}
+      {% include "toolkit/flash_messages.html" %}
```
lfdebrux: This is different to the other apps, where we left the flash messages inside the main. Is this change needed? 🤔

katstevens:
Error messages look good:

error-banner

app/templates/_base_page.html
aliuk2012: This is needed for the moment otherwise the flash message is pushed down a bit further than whats needed.

I agree, they should be identical but they apps are in different states at the moment, eg. not using govuk-grid or other sass mixins.

katstevens:
Confirmation banner with action looks 👍 

admin-flash-msg-with-action

app/templates/_base_page.html
lfdebrux: That's interesting; so this fix will become necessary for the other apps later on? What would happen if we applied this fix to the other apps now?

aliuk2012: I don't want to spend any more time on this issue as it is a temporary fix and the bigger issue is trying to fix notifications/temp/flash banners and will have to revisit this again.

PR #584, opened 2019-10-29
Remove use of secondary action link component

lfdebrux:
Using the GOV.UK Design System [spacing override classes] we can add spacing around a link without need for a custom component or even a div.

When adding spacing override classes to inline elements such as anchors, we do need to also add the `govuk-!-display-inline-block` override.

[spacing override classes]: https://design-system.service.gov.uk/styles/spacing/#spacing-override-classes

### DIFF
--- a/app/templates/compare_revisions.html
+++ b/app/templates/compare_revisions.html
@@ -26,13 +26,11 @@
   <span class="govuk-caption-l">{{ service['supplierName'] }}</span>
   <h1 class="govuk-heading-l">{{ service['serviceName'] }}</h1>
 
-  {%
-    with
-      url = url_for("external.direct_award_service_page", framework_family=service['frameworkFamily'], service_id=service["id"]),
-      text = "View service"
-  %}
-    {% include "toolkit/secondary-action-link.html" %}
-  {% endwith %}
+  <a class="govuk-link govuk-!-margin-bottom-3 govuk-!-display-inline-block"
+     href="{{ url_for('external.direct_award_service_page', framework_family=service['frameworkFamily'], service_id=service['id']) }}">
+    View service
+  </a>
+
     <div class="govuk-grid-row">
       <div class="govuk-grid-column-two-thirds page-section">
         <p class="govuk-body">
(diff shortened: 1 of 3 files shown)


app/templates/compare_revisions.html
```diff
@@ -26,13 +26,10 @@
...
   <h1 class="govuk-heading-l">{{ service['serviceName'] }}</h1>
 
-  {%
-    with
-      url = url_for("external.direct_award_service_page", framework_family=service['frameworkFamily'], service_id=service["id"]),
-      text = "View service"
-  %}
-    {% include "toolkit/secondary-action-link.html" %}
-  {% endwith %}
+  <a class="govuk-link govuk-!-margin-bottom-3 govuk-!-display-inline-block" href={{ url_for("external.direct_award_service_page", framework_family=service['frameworkFamily'], service_id=service["id"]) }}>
```
katstevens: Missing quotes on the `href`?

lfdebrux: Good spot...

lfdebrux: I wish that line was less long

app/templates/compare_revisions.html
```diff
@@ -26,13 +26,10 @@
...
   <h1 class="govuk-heading-l">{{ service['serviceName'] }}</h1>
 
-  {%
-    with
-      url = url_for("external.direct_award_service_page", framework_family=service['frameworkFamily'], service_id=service["id"]),
-      text = "View service"
-  %}
-    {% include "toolkit/secondary-action-link.html" %}
-  {% endwith %}
+  <a class="govuk-link govuk-!-margin-bottom-3 govuk-!-display-inline-block" href="{{ url_for("external.direct_award_service_page", framework_family=service['frameworkFamily'], service_id=service["id"]) }}">
```
katstevens: Can I be really picky and ask that we use single quotes around the `url_for` param if we're using double quotes on the `href`? For the sake of the colour scheme on my editor, if nothing else 😹

lfdebrux: Oh nuts yes that's a good point

katstevens:
Are we getting rid of related custom CSS as we remove components, or are we going to clean that up later? In `application.scss` we still have a `.question + .secondary-action-link` class...

lfdebrux:
> Are we getting rid of related custom CSS as we remove components, or are we going to clean that up later? In `application.scss` we still have a `.question + .secondary-action-link` class...

We can definitely get rid of that! 🔥 Nice spot

PR #590, opened 2019-11-05
Updates digitalmarketplace-govuk-frontend to 0.2.1

aliuk2012:
Fixes an issue with error page presentation and now uses digitalmarketplace-govuk-frontend error templates.

Ticket: https://trello.com/c/T9hBnYM5

### DIFF
--- a/config.py
+++ b/config.py
@@ -57,9 +57,12 @@ class Config(object):
     @staticmethod
     def init_app(app):
         repo_root = os.path.abspath(os.path.dirname(__file__))
+        digitalmarketplace_govuk_frontend = os.path.join(repo_root, "node_modules", "digitalmarketplace-govuk-frontend")
+
         template_folders = [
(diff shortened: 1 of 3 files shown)


config.py
```diff
@@ -60,6 +60,7 @@ def init_app(app):
         template_folders = [
             os.path.join(repo_root, 'app', 'templates'),
             os.path.join(repo_root, 'node_modules', 'digitalmarketplace-govuk-frontend', 'govuk-frontend'),
+            os.path.join(repo_root, 'node_modules', 'digitalmarketplace-govuk-frontend', 'digitalmarketplace', 'templates'),
```
lfdebrux: Can we put this line above `node_modules/digitalmarketplace-govuk-frontend/govuk-frontend`? We should be prioritising templates from the `digitalmarketplace` folder over ones in the latter.

lfdebrux:
I'll fix the Python linting error

PR #591, opened 2019-11-08
Removes dmspeak

aliuk2012:
- fixes an accessibility issue found on the index page
- removes dmspeak from application views

Ticket: https://trello.com/c/83H65zij


lfdebrux:
Awesome work

app/templates/dashboard/_admin_ccs_data_controller_index.html
```diff
@@ -1,5 +1,5 @@
 {# ADMIN CCS DATA CONTROLLER #}
-<h2 class="heading-xmedium">Search for users</h3>
+<h2 class="govuk-heading-m">Search for users</h3>
```
lfdebrux: It would be great if we could split these changes to remove `heading-*` classes into a separate commit or PR, so that we can translate those changes into a PR that other devs on the team can do.

app/templates/dashboard/_misc_index.html
```diff
@@ -60,43 +76,66 @@ <h3 class="heading-xmedium">Buyers</h3>
...
-        {% if current_user.has_any_role('admin-ccs-sourcing', 'admin-framework-manager') %}
+        {% if current_user.has_role('admin-ccs-sourcing') %}
           <li>
-            <a class="govuk-link" href="{{ url_for('public.view_statistics', framework_slug=framework['slug']) }}">View application statistics</a>
+            <a class="govuk-link" href="{{ url_for('public.view_statistics', framework_slug=framework['slug']) }}">
+              View application statistics
+            </a>
           </li>
-        {% endif %}
-        {% if current_user.has_role('admin-framework-manager') %}
```
lfdebrux: It's not clear to me what this change in logic is doing; it looks like we're just repeating the same HTML snippet twice now?

aliuk2012: It is repeating HTML, The reason for the change is to try to make it easier to read and understand what a specific type of user should be able to see by looking at one if statement rather than trying to build up a mental model of what it is they are meant to see on the page.

Does it make it easier to read?

app/templates/dashboard/_admin_ccs_data_controller_index.html
aliuk2012: I think it's going to be difficult to split it out because these classes can be inconsistent and it depends on the context that they being used whether the heading size is appropriate.

katstevens:
Can we make it explicitly clear in the PR description (or commit history) which bits of the old index page are moving into which new files. If we need to retrace history at some point and figure out what the old page structure/logic was like, having some clues in the new files for our future Detective Pikachus would be awesome.

I say this as someone who has had to recently wade through my own changes for the error pages like [Ariadne in the labyrinth](https://en.wikipedia.org/wiki/Ariadne)...

aliuk2012:
@katstevens  I totally agree and I was conscious of this while I was making the changes this is why I created this commit https://github.com/alphagov/digitalmarketplace-admin-frontend/pull/591/commits/d9f5ab5965a52bff04479be9b6e80035f5fcbbc9

app/templates/dashboard/_misc_index.html
lfdebrux: I would say it is easier to read. I'm not sure whether it's easier to maintain, I guess it depends on whether we expect the capabilities of roles to change a lot. Either layout is fine though.

katstevens:
I've had another look through and I'm not sure we're gaining much at this stage by splitting the index template. `misc_index` is still pretty complicated, and now we have to look in three places to work out what's going on for one view. If the permissions change for a role in the future (which is quite likely, the more we automate/iterate the framework process) then I can foresee the templates getting quite knotty. 

How about we keep this PR to the `govuk-frontend` upgrades, and create a separate tech debt ticket to focus on simplifying the index page (looking at the backend logic as well, maybe)?

Also I noticed we only have visual regression test scenarios for one role on the index page - can we add in the other roles there before we think about merging/deploying this?

lfdebrux:
> Also I noticed we only have visual regression test scenarios for one role on the index page - can we add in the other roles there before we think about merging/deploying this?

I can do this

app/templates/index.html
```diff
@@ -12,153 +12,198 @@ <h1 class="govuk-heading-xl">Admin</h1>
...
-        {# ADMIN CCS DATA CONTROLLER #}
-        <h3 class="heading-xmedium">Search for users</h3>
-        <ul class="govuk-list">
-          <li>
-            <a class="govuk-link" href="{{ url_for('.find_user_by_email_address') }}">Find a user by email</a>
-          </li>
-        </ul>
+    {% if current_user.has_role('admin-ccs-data-controller') %}
+      {# ADMIN CCS DATA CONTROLLER #}
+      <h2 class="govuk-heading-m">Search for users</h3>
```
katstevens: Mismatched elements here - `<h2>` vs `</h3>`

app/templates/index.html
```diff
@@ -12,153 +12,198 @@ <h1 class="govuk-heading-xl">Admin</h1>
...
       <ul class="govuk-list">
         <li>
-          <a class="govuk-link" href="{{ url_for('.search_suppliers_and_services') }}">{{ find_supplier_and_services_link_text[current_user.role] }}</a>
+          <a class="govuk-link" href="{{ url_for('.search_suppliers_and_services') }}">
+            {{ find_supplier_and_services_link_text[current_user.role] }}
+          </a>
         </li>
 
-        {% if current_user.has_any_role('admin-framework-manager') %}
+        {% if current_user.has_role('admin-framework-manager') %}
```
katstevens: Good spot on these 👍

app/templates/index.html
```diff
@@ -12,153 +12,198 @@ <h1 class="govuk-heading-xl">Admin</h1>
...
+                <li>
+                  <a class="govuk-link" href="{{ url_for('.user_list_page_for_framework', framework_slug=framework['slug']) }}">
+                    Contact suppliers
+                  </a>
+                </li>
+
+              {% elif current_user.has_role('admin-ccs-category') %}
+                <li>
+                  <a class="govuk-link" href="{{ url_for('.user_list_page_for_framework', framework_slug=framework['slug']) }}">
+                    Contact suppliers
```
katstevens: I think this duplication is fine for now 👍

app/templates/index.html
katstevens: (There are a couple of these)

aliuk2012: Good spot, thank you

katstevens: Fab 👍 Once we have the extra VR scenarios ready to go I'll approve

PR #595, opened 2019-11-27
agreement countersigning: pass company_details.registered_name to action views

risicle:
https://trello.com/c/nDi5GMyq

Instead of `declaration.nameOfOrganisation`, which no longer exists in recent frameworks. This was causing the company names to be missing from flash messages following actions.


risicle:
My thoughts exactly.

PR #596, opened 2019-11-28
remove internal stats view

risicle:
https://trello.com/c/qjqbNAEO

Instead now redirecting requests for this to the performance platform. Mapping of slug -> performance platform identifier is kept in `config`, which felt like the best place for it. Requests without a known mapping will be 410ed, the original _links_ to the stats view now point directly to the performance platform location using the same mapping.


app/assets/javascripts/application.js
```diff
@@ -7,8 +7,6 @@
 //= require ../../../node_modules/govuk_frontend_toolkit/javascripts/vendor/polyfills/bind.js
 //= require ../../../node_modules/jquery/dist/jquery.js
 //= require ../../../node_modules/hogan.js/web/builds/3.0.2/hogan-3.0.2.js
-//= require ../../../node_modules/d3/d3.js
-//= require ../../../node_modules/c3/c3.js
```
katstevens: There is a `_c3.scss` file we can ditch as well, I think

app/templates/index.html
```diff
@@ -164,19 +164,15 @@ <h3 class="govuk-heading-m">{{framework.name}}</h3>
                 </li>
               {% endif %}
 
-              {% if current_user.has_role('admin-ccs-sourcing') %}
+              {% if framework['slug'] in config.PERFORMANCE_PLATFORM_ID_MAPPING and current_user.has_any_role('admin-ccs-sourcing', 'admin-framework-manager') %}
                 <li>
-                  <a class="govuk-link" href="{{ url_for('public.view_statistics', framework_slug=framework['slug']) }}">
+                  <a class="govuk-link" href="{{ config.PERFORMANCE_PLATFORM_BASE_URL }}{{ config.PERFORMANCE_PLATFORM_ID_MAPPING[framework['slug']] }}">
```
katstevens: I think Jinja will let us simplify to `config.PERFORMANCE_PLATFORM_ID_MAPPING[framework.slug]`

app/assets/javascripts/application.js
risicle: There was also `views/_statistics.scss` which I've now got rid of too.

app/templates/index.html
risicle: I'll simplify _you_

risicle: Fixed.

katstevens: Did it get pushed?

risicle: Hmm. No.

PR #598, opened 2019-12-02
Remove Jasmine and Hogan

lfdebrux:
This PR removes two unneeded and outdated JavaScript libraries, Hogan.js and gulp-jasmine-phantom, by updating its version of digitalmarketplace-frontend-toolkit.

This version of admin frontend should have many fewer Snyk vulnerabilities 🤞

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -5,7 +5,6 @@
   directives to concatenate multiple Javascript files into one.
 */
 //= require ../../../node_modules/jquery/dist/jquery.js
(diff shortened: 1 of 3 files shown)


katstevens:
What this really highlights is that we don't have any JS unit tests in the admin...

aliuk2012:
Doesn't Admin use list-entry?

lfdebrux:
> Doesn't Admin use list-entry?

It does... I thought the idea of the change to frontend toolkit was to remove Hogan... I tested this version locally and the list entry component seemed to work...

aliuk2012:
image
## Another big one 👏

PR #600, opened 2019-12-09
Replace validation banner with GOV.UK Frontend error summary component

risicle:
https://trello.com/c/xdGjiDqF

Depends on https://github.com/alphagov/digitalmarketplace-utils/pull/548/

Spectacle h11232

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,7 +10,7 @@ lxml==4.4.1
 # Putting upgrade on hold due to v1.0.0 using sha512 instead of sha1 by default
 itsdangerous==0.24  # pyup: <1.0.0
 
-git+https://github.com/alphagov/digitalmarketplace-utils.git@50.2.1#egg=digitalmarketplace-utils==50.2.1
+git+https://github.com/alphagov/digitalmarketplace-utils.git@50.4.0#egg=digitalmarketplace-utils==50.4.0
 git+https://github.com/alphagov/digitalmarketplace-content-loader.git@7.1.1#egg=digitalmarketplace-content-loader==7.1.1
 git+https://github.com/alphagov/digitalmarketplace-apiclient.git@21.4.1#egg=digitalmarketplace-apiclient==21.4.1
 git+https://github.com/alphagov/govuk-frontend-jinja.git@v0.4.0-alpha#egg=govuk-frontend-jinja==0.4.0-alpha
@@ -18,8 +18,8 @@ git+https://github.com/alphagov/govuk-frontend-jinja.git@v0.4.0-alpha#egg=govuk-
 ## The following requirements were added by pip freeze:
 asn1crypto==1.2.0
 blinker==1.4
-boto3==1.10.33
-botocore==1.13.33
+boto3==1.10.36
+botocore==1.13.36
 certifi==2019.11.28
 cffi==1.13.2
 chardet==3.0.4
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,7 +10,7 @@ lxml==4.4.1
 # Putting upgrade on hold due to v1.0.0 using sha512 instead of sha1 by default
 itsdangerous==0.24  # pyup: <1.0.0
 
-git+https://github.com/alphagov/digitalmarketplace-utils.git@50.0.1#egg=digitalmarketplace-utils==50.0.1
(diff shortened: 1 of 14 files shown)


risicle:
Spectacle n11232

risicle:
Spectacle X12554

lfdebrux:
Great stuff.

Can we also remove the old validation banner from https://github.com/alphagov/digitalmarketplace-admin-frontend/blob/ris-fe-validation-banner/app/templates/invite_admin_user.html#L29?

risicle:
Hmmyes, not sure why I missed that one...

lfdebrux:
I've found an issue with the country picker when editing a suppliers registered address (http://localhost/admin/suppliers/700631/edit/registered-address), when there is an error in the input for that field the link doesn't go to the right place.

I'm going to try updating `govuk-country-and-territory-autocomplete` and see if that helps.

lfdebrux:
> I've found an issue with the country picker when editing a suppliers registered address (http://localhost/admin/suppliers/700631/edit/registered-address), when there is an error in the input for that field the link doesn't go to the right place.
> 
> I'm going to try updating `govuk-country-and-territory-autocomplete` and see if that helps.

Updating the component didn't work, instead the input id needed to change. I've pushed a fix, will also add this as a gotcha to the ticket.

risicle:
Yup, can confirm the fix works.

aliuk2012:
Can we remove these two lines now that validation component has been removed? 

CSS:
https://github.com/alphagov/digitalmarketplace-admin-frontend/blob/ris-fe-validation-banner/app/assets/scss/application.scss#L32

JavaScript:
https://github.com/alphagov/digitalmarketplace-admin-frontend/blob/ris-fe-validation-banner/app/assets/javascripts/application.js#L9

lfdebrux:
> Can we remove these two lines now that validation component has been removed?
> 
> CSS:
> https://github.com/alphagov/digitalmarketplace-admin-frontend/blob/ris-fe-validation-banner/app/assets/scss/application.scss#L32
> 
> JavaScript:
> https://github.com/alphagov/digitalmarketplace-admin-frontend/blob/ris-fe-validation-banner/app/assets/javascripts/application.js#L9

I seem to recall that we can't remove the CSS because it's also used in validation messages for fields; we should be able to remove the JavaScript though?

aliuk2012:
Ah yes, I remember we had that weird issue with styling the error fields.
Let's remove the line in the JS and perhaps add a comment above the line in the css to explain why it's still needed.

risicle:
ok

requirements.txt
```diff
@@ -10,17 +10,17 @@ lxml==4.4.1
 # Putting upgrade on hold due to v1.0.0 using sha512 instead of sha1 by default
 itsdangerous==0.24  # pyup: <1.0.0
 
-git+https://github.com/alphagov/digitalmarketplace-utils.git@50.0.1#egg=digitalmarketplace-utils==50.0.1
+git+https://github.com/alphagov/digitalmarketplace-utils.git@ldeb-add-forms-errors-helpers#egg=digitalmarketplace-utils==9999
```
katstevens: This is merged now, right?

lfdebrux: Yeah

lfdebrux: I've pushed a version with updated requirements

risicle:
Good catch I would have forgotten about that.

lfdebrux:
When I run the functional tests I get a JavaScript error:

```
When I click the summary table 'Change' link for 'Address'
features/step_definitions/common_steps.rb:333
One or more errors were raised in the Javascript code on the page. If you don't care about these errors, you can ignore them by setting js_errors: false in your Poltergeist configuration (see documentation for details).

TypeError: null is not an object (evaluating 'r.selectElement.options')
TypeError: null is not an object (evaluating 'r.selectElement.options')
    at http://localhost/admin/static/javascripts/location-autocomplete.min.js:1 in c
    at http://localhost/admin/static/javascripts/app-location-autocomplete.js:7 in CountryAutocomplete
    at http://localhost/admin/static/javascripts/app-location-autocomplete.js:20 (Capybara::Poltergeist::JavascriptError)
```

Which is probably from the JavaScript changes I made... any ideas what the issue might be @aliuk2012 ?

lfdebrux:
Ah, the functional tests do need updating though because the id of the autocomplete field has changed

PR #601, opened 2019-12-11
buyer domain addition: add validation, blacklisting known purchasable TLDs

risicle:
https://trello.com/c/9Ui8pyjG

Implementation of validator obviously heavily inspired by the password blacklist in `-user-frontend` https://github.com/alphagov/digitalmarketplace-user-frontend/blob/30baaa6e1a94e132aa2082a2375a0535a2db3274/app/main/forms/auth_forms.py#L29 but it all seemed *just* different enough for it to be more of a pain than a benefit sharing code between the two.

A trie would be slightly faster at searching for prefixes, but the python trie implementations I could find had slightly inconvenient interfaces for doing this kind of lookup. Using `bisect` instead also has the advantage of not adding another dependency.


tests/app/main/views/test_buyers.py
```diff
@@ -152,18 +152,46 @@ def test_post_page_should_only_be_accessible_to_specific_user_roles(self, role,
         actual_code = response.status_code
         assert actual_code == expected_code, "Unexpected response {} for role {}".format(actual_code, role)
 
-    def test_admin_user_can_add_a_new_buyer_domain(self):
+    @pytest.mark.parametrize("new_buyer_domain", (
+        "kev.uk",
+        "zz",
+        "keyes.ORG.uk",
+        ".GOV",
+        "om",  # a suffix of "com", but not if we're (correctly) considering an implicit preceding separator
```
katstevens: This is an interesting one. I wonder if we should also have a minimum length of say, 3, set? Is there any realistic situation where a public sector buyer is going to have a 2 character email domain? Out of scope of this PR, just thinking out loud...

katstevens: I guess this would go in the API's schema validation though

katstevens: `.GOV` will fail API validation - prob best not to use it as an example given we're mocking the response

app/main/forms.py
```diff
@@ -74,13 +80,73 @@ class MoveUserForm(FlaskForm):
...
+    "WTForms validator, asserting that supplied value is not a prefix of any of the values present in a blacklist"
+
+    # path, relative to flask app root_path, to look for suffix blacklist files. all files found here will be read,
+    # one suffix per line
+    BLACKLIST_DIR_PATH: str = "data/public_domain_suffix_blacklist"
+
+    @staticmethod
+    def _normalized_suffix(suffix: str) -> str:
+        return (
+            # all suffixes we deal with implicitly begin with a separator - make that explicit here
```
benvand: Can these inline comments just be a docstring?

> Return reversed suffix and ensure last character is a '.'

Also `        return (("" if suffix.startswith(".") else ".") + suffix.strip().lower())[::-1]` fits within line limit

app/main/forms.py
```diff
@@ -74,13 +80,73 @@ class MoveUserForm(FlaskForm):
...
+
+    # this value is not populated until first access because construction depends on current_app being available
+    _blacklist: typing.Optional[typing.Sequence[str]] = None
+
+    @classmethod
+    def get_blacklist(cls) -> typing.Sequence[str]:
+        # cache blacklist class-wide
+        if cls._blacklist is None:
+            # store this as a sorted tuple so we can use bisection to find specific prefixes
+            cls._blacklist = tuple(sorted(chain.from_iterable(
```
benvand: What does this look like?

app/main/forms.py
```diff
@@ -74,13 +80,73 @@ class MoveUserForm(FlaskForm):
...
+        return cls._blacklist
+
+    def __init__(self, message):
+        self.message = message
+
+    def __call__(self, form, field) -> None:
+        blacklist = self.get_blacklist()
+        normalized_data = self._normalized_suffix(field.data)
+        i = bisect_left(blacklist, normalized_data)
+        # if (the normalized) `field.data` is a prefix of any of the (normalized) entries in the blacklist, it
```
benvand: > should sort just before the first such entry. hence we just have to check whether normalized_data is a prefix of the entry that sorts right after it

This could use clarification. What is 'sort before'/ 'sort after' it?

app/main/forms.py
```diff
@@ -74,13 +80,73 @@ class MoveUserForm(FlaskForm):
         ])
 
 
+class NotSuffixOfBlacklist:
```
katstevens: How about `class NotInDomainSuffixBlacklist`? (That matches our password blacklist validator naming more closely)

app/main/forms.py
```diff
@@ -74,13 +80,73 @@ class MoveUserForm(FlaskForm):
...
+    # path, relative to flask app root_path, to look for suffix blacklist files. all files found here will be read,
+    # one suffix per line
+    BLACKLIST_DIR_PATH: str = "data/public_domain_suffix_blacklist"
+
+    @staticmethod
+    def _normalized_suffix(suffix: str) -> str:
+        return (
+            # all suffixes we deal with implicitly begin with a separator - make that explicit here
+            ("" if suffix.startswith(".") else ".") + suffix.strip().lower()
+            # we want to find suffixes. sorting works by prefix, so reverse all strings before we work with them
```
katstevens: This has blown my mind a little bit. In a good way.

app/main/forms.py
```diff
@@ -74,13 +80,73 @@ class MoveUserForm(FlaskForm):
...
+
+    def __call__(self, form, field) -> None:
+        blacklist = self.get_blacklist()
+        normalized_data = self._normalized_suffix(field.data)
+        i = bisect_left(blacklist, normalized_data)
+        # if (the normalized) `field.data` is a prefix of any of the (normalized) entries in the blacklist, it
+        # should sort just before the first such entry. hence we just have to check whether normalized_data is
+        # a prefix of the entry that sorts right after it. an "equal" match should work in the same way because
+        # we use bisect_left
+        if i < len(blacklist) and blacklist[i].startswith(normalized_data):
```
benvand: I'm don't think bisect actually is quicker than an `in` operation on a set. You've already got O(n) because you're iterating the blacklist to normalise the prefixes. Then a bisect left performs an insertion, which is also pretty much O(n).

You can achieve the same complexity with 

`cls._normalized_suffix(field) in {cls._normalized_suffix(line) for line in open(filepath)}`

Or something like that

benvand: https://stackoverflow.com/a/53023435/3967444

app/main/forms.py
risicle: docstrings are for a different thing - to document the general thing the function does from an outside caller's point of view

app/main/forms.py
risicle: e.g.

```
ku
ku.gro
ku.oc
...
```

tests/app/main/views/test_buyers.py
risicle: I think actually the API's json schema based validation will stop anything < 2 chars.

app/main/forms.py
risicle: Yyyyyyyeah I guess I was trying to make it more generic.

app/main/forms.py
risicle: But yes, the shorter explanation will probably do.

app/main/forms.py
risicle: Can't use `set`s because we're not doing exact matching. We want to match suffixes _of suffixes_ too (consider the blacklist containing `.foo.bar`, we want to disallow `.bar` too. Bisection is also O(log n) not O(n).

risicle: We're also not inserting anything, only using `bisect_left` to look up the hypothetical insertion _point_.

tests/app/main/views/test_buyers.py
risicle: Yup, as discussed what's actually going on here is I'm reying to squeeze unit-y tests of the validator into the _view_'s unit tests. Which is what makes it inappropriate to put otherwise-failing examples in here. Will split the parametrization out into separate tests.

app/main/forms.py
benvand: This confusion probably could have been avoided by not using `i` as a var name.

app/main/forms.py
benvand: We don't need the capacity to have multiple blacklist files so this can be simplified to:

```
cls._blacklist = sorted(cls._lines_from_filepath(os.path.join([current_app.root_path, cls.BLACKLIST_DIR_PATH])))
```

risicle:
Oh and "why implement this in the admin FE instead of the API"? Well, I weighed up both. It felt more like a safety catch than a hard rule. Devs with API keys making changes via the API should know better.

app/main/forms.py
risicle: Added pointer to bisect module documentation that explains more.

app/main/forms.py
risicle: Having the capacity for multiple files makes managing the list(s) easier. If we want to add a second source of TLDs, or maybe just a set of manual pragmas, we can add them as a new file named as such instead of having to manage a single merged frankenstein file. Then when someone wants to update the namecheap file they can replace it easily without worrying about flushing out any manual additions made to it. It's also copy-pasta'd code and there's an advantage to having things behave similarly in different places.

PR #607, opened 2020-01-07
add find_supplier_draft_services view 

risicle:
https://trello.com/c/g5P0RxQn

Spectacle X14222

Can't think of much I have to say - fairly straightforward. Currently implemented to just show drafts from all framework states to `admin-framework-manager`, nobody else.

Spectacle X14322
Spectacle h14401

### DIFF
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -3,6 +3,8 @@ from itertools import groupby, chain
 from operator import itemgetter
 
 from dateutil.parser import parse as parse_date
+from dmcontent.errors import ContentNotFoundError
+from dmcontent.utils import count_unanswered_questions
 from dmapiclient import HTTPError, APIError
 from dmapiclient.audit import AuditTypes
 from dmutils import s3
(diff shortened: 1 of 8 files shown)


app/main/views/suppliers.py
```diff
@@ -866,6 +868,48 @@ def toggle_supplier_services(supplier_id):
...
+            **draft_service,
+            "unansweredRequiredCount": count_unanswered_questions(
+                manifest.filter(draft_service).summary(draft_service)
+            )[0],
+        } for draft_service in draft_services
+    )
+
+
+@main.route('/suppliers/<int:supplier_id>/draft-services', methods=['GET'])
+@role_required('admin-framework-manager')
```
katstevens: I think we should probably add the Sourcing role here too (not sure about Category admins, they generally care about live services rather than drafts, but it's easy enough to add them in later if needed).

Can we add a quick test to check that the other roles 403 on this view? We have that test pattern for most of the other admin views:

```
@pytest.mark.parametrize("role,expected_code", [
        ("admin", 403),
        ("admin-ccs-category", 403),
        ("admin-ccs-sourcing", 200),
        ("admin-manager", 403),
        ("admin-framework-manager", 200),
    ])
...
```

app/templates/_view_suppliers_edit_list.html
```diff
@@ -1,10 +1,12 @@
 {% set field_headings = [
   "Supplier Name"
-] if current_user.has_role('admin-ccs-sourcing') else [
+] + ([
   "Supplier Name",
```
katstevens: I think this causes `Supplier Name` to appear twice in the table heading...

app/main/views/suppliers.py
risicle: Righto. Should sourcing also be able to see "all" framework statuses?

app/templates/_view_suppliers_edit_list.html
risicle: Doh. This was a late addition.

app/main/views/suppliers.py
katstevens: Yes.

risicle: great - done

katstevens:
Tech debt card: https://trello.com/c/qkDkefqS/1215-alignment-on-admin-fe-viewsupplierseditlist-table-headers

risicle:
Yes, the header alignment is ug-lee, but I couldn't see how to tweak it manually using our tookit.

PR #608, opened 2020-01-07
Bumps digitalmarketplace-frontend-toolkit to v36.1.0

aliuk2012:
Ticket: https://trello.com/c/RiF9PMI8


lfdebrux:
What changes are there in this bump?

aliuk2012:
Even though it is a major version bump, this doesn't break anything because the major release was for the footer layout changes and admin frontend does not use the footer.

the minor version bump is needed for pii.js to not rely on jQuery.

PR #609, opened 2020-01-16
Fix default link colour and other missing class names

aliuk2012:
## Before 
image

## After 
image

Ticket: https://trello.com/c/gpvJBf4t

### DIFF
--- a/app/assets/scss/application-ie7.scss
+++ b/app/assets/scss/application-ie7.scss
@@ -1,7 +1,0 @@
-// BASE STYLESHEET FOR IE 7 COMPILER
-
-$is-ie: true;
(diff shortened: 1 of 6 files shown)


katstevens:
Tested locally and the link colour is now fine 👍 The default text colour is still missing though - other apps are getting `color: #0b0c0c` from the govuk-template `body` element for plain text (eg inside summary tables) whereas on Admin it's using the browser default.

aliuk2012:
@katstevens should be fixed now. I think I just went in to deep and removed to much

lfdebrux:
Visual regression changes look good to me, I think we can merge this.

PR #610, opened 2020-01-16
content loader: use new `inplace_allowed` options in .filter() and .summary() methods where appropriate

risicle:
Basically the same as https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/1151

https://trello.com/c/ooTuJDCw

Significant performance gains here. Partly from the content loader bump as it contains some opaque optimizations, but mostly stemming from these additions. These are all in places where it can be seen that the original object reference is discarded afterwards, so it can be considered safe.


risicle:
Yes... mostly in some tests? But some of the places I've set it here needed to be slightly adapted to make it clear that it's safe.

PR #634, opened 2020-03-24
minimist JS library update

katstevens:
https://trello.com/c/Gb30qeGd/1556-vulnerability-fix-tracking

Addresses a Snyk alert for this package. Updated using `npm audit fix`.

Standard JS also complained about a trailing comma in the gulpfile, so I've fixed that.


katstevens:
Thanks @lfdebrux - I'll be bumping the other apps today as well. Travis still seems patchy in terms of queuing up jobs though 😿

PR #654, opened 2020-06-01
bump utils version

gidsg:
part of https://trello.com/c/PTdD5lEh/102-our-logs-show-paas-instance-guid-but-not-paas-instance-index


katstevens:
We're introducing a major version bump - usually for these it's good to call out in the PR description whether the breaking change might affect this app or not ([in this case it looks fine to me](https://github.com/alphagov/digitalmarketplace-utils/blob/master/CHANGELOG.md)).

PR #665, opened 2020-09-02
Admin view for e-signature framework agreements (for G12 onwards)

katstevens:
https://trello.com/c/psi3gVdH/157-2-admins-can-view-the-new-pdf-for-g12

- Simplify index page content to reduce confusion about whether an agreement needs a countersigning 'action' to happen
- Change agreement list page to make it clear that no countersigning action is required (and to contact suppliers if there is a problem)
- New template for showing signer details for G12+ frameworks, with link to countersigned PDF if present

Signed, but not countersigned:
admin-esig-not-countersigned

Signed and countersigned (link to PDF file generated by Jenkins):
admin-esig-countersigned

Agreement list page for an e-signature framework (eg G12):
admin-esig-agreement-list

### DIFF
--- a/app/main/views/agreements.py
+++ b/app/main/views/agreements.py
@@ -17,6 +17,17 @@ status_labels = OrderedDict((
                                                   # much easier to just act as a filter
 ))
 
+OLD_COUNTERSIGNING_FLOW_FRAMEWORKS = [
+    'digital-outcomes-and-specialists-4',
+    'digital-outcomes-and-specialists-3',
+    'digital-outcomes-and-specialists-2',
+    'digital-outcomes-and-specialists',
+    'g-cloud-11',
+    'g-cloud-10',
+    'g-cloud-9',
+    'g-cloud-8',
+]
+
 
 def _get_supplier_frameworks(framework_slug, status=None):
     return [
@@ -44,14 +55,26 @@ def list_agreements(framework_slug):
         supplier_framework['agreementReturnedAt'] = datetimeformat(
             parse_date(supplier_framework['agreementReturnedAt']))
 
+    # Determine which template to use.
+    # G-Cloud 7 and earlier frameworks do not have a frameworkAgreementVersion and use an old countersigning flow
+    # G-Cloud 12 and newer frameworks use e-signature flow (countersignatures are automated)
+    is_e_signature_flow = False
+    if framework.get('frameworkAgreementVersion'):
+        template = "view_agreements_list.html"
+        if framework['slug'] not in OLD_COUNTERSIGNING_FLOW_FRAMEWORKS:
+            status_labels['signed'] = "Waiting for automated countersigning"
+            is_e_signature_flow = True
+    else:
+        template = 'view_agreements.html'
+
     return render_template(
-        # G-Cloud 8 and newer frameworks have a frameworkAgreementVersion and use the new countersigning flow
-        "view_agreements_list.html" if framework.get("frameworkAgreementVersion") else 'view_agreements.html',
+        template,
         framework=framework,
         supplier_frameworks=supplier_frameworks,
         degenerate_document_path_and_return_doc_name=lambda x: degenerate_document_path_and_return_doc_name(x),
         status=status,
         status_labels=status_labels,
+        is_e_signature_flow=is_e_signature_flow
     )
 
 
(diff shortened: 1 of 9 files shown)


app/main/views/agreements.py
```diff
@@ -17,6 +17,17 @@
...
+    'digital-outcomes-and-specialists-4',
+    'digital-outcomes-and-specialists-3',
+    'digital-outcomes-and-specialists-2',
+    'digital-outcomes-and-specialists',
+    'g-cloud-11',
+    'g-cloud-10',
+    'g-cloud-9',
+    'g-cloud-8',
+]
+
```
gidsg, line 30: Should we move the e-signature check i.e. https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/bba069b6e3d680ef106872a7119e03636a6aff17/app/main/helpers/frameworks.py#L478-L479 into utils so admin and supplier can share the logic?

app/templates/suppliers/view_esignature_agreement.html
```diff
@@ -0,0 +1,65 @@
...
+
+{% block mainContent %}
+  <h1 class="govuk-heading-l">{{ company_details.registered_name }}</h1>
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+
+        <p><a href="{{ url_for('.supplier_details', supplier_id=supplier.id) }}">View company details</a></p>
+
+       <h2 class="govuk-heading-m">Signature details</h2>
+      {% set lots = "whatever" %}
```
gidsg, line 34: Is this `whatever` a placeholder?

app/main/views/agreements.py
katstevens: Hmm good question. In the admin it's a slightly different usage, as we show a different signature template for G-Cloud 7 and earlier. Having it be consistent across apps would be useful though...

app/templates/suppliers/view_esignature_agreement.html
katstevens: Hahaha whoops! Yes this is definitely a WIP 😹

NoraGDS:
On the Agreement list page for an e-signature framework, suggest a slight rewording of first sentence as there is no longer uploading of a file:

If there's a problem with the agreement, contact the person who signed it.

katstevens:
> On the Agreement list page for an e-signature framework, suggest a slight rewording of first sentence as there is no longer uploading of a file:
> 
> If there's a problem with the agreement, contact the person who signed it.

Ah great spot @NoraGDS! Have pushed a commit to fix.

app/main/views/suppliers.py
```diff
@@ -376,32 +377,49 @@ def view_signed_agreement(supplier_id, framework_slug):
...
+
+    if framework_slug in OLD_COUNTERSIGNING_FLOW_FRAMEWORKS:
+        is_e_signature_flow = False
+        # Fetch path to supplier's signature page
+        path = supplier_framework['agreementPath']
+        template = "suppliers/view_signed_agreement.html"
+    else:
+        is_e_signature_flow = True
+        # Fetch path to combined countersigned agreement, if available
+        path = supplier_framework.get('countersignedPath', "")
```
gidsg, line 404: What's the effect of the `""` parameter here?

app/templates/suppliers/view_esignature_agreement.html
```diff
@@ -0,0 +1,64 @@
...
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+
+        <p><a href="{{ url_for('.supplier_details', supplier_id=supplier.id) }}">View company details</a></p>
+
+       <h2 class="govuk-heading-m">Signature details</h2>
+        {% call(item)
+              summary.list_table(
+                [
+                  {"name": "Appointment is to", "value": ", ".join(lot_names) },
```
gidsg, line 37: Can we use jinja syntax here e.g. `lot_names | join(', ')`?

app/main/views/suppliers.py
katstevens: It sets the default fallback, so instead of `None` we have a string. I don't think we actually need it any more, as we're doing an existence check below. Happy to remove?

gidsg: @katstevens makes sense to remove I think, if subsequent behaviour is the same without it.

app/templates/suppliers/view_esignature_agreement.html
katstevens: We can indeed!

app/main/views/suppliers.py
katstevens: @gidsg Now done

gidsg (approved):
✨ 👍

katstevens:
Woohoo thanks @gidsg !

PR #667, opened 2020-09-07
Convert status_labels from global to function

katstevens:
https://trello.com/c/psi3gVdH/157-2-admins-can-view-the-new-pdf-for-g12

The mutable nature of dictionaries meant that when we changed the label to `waiting for automated countersigning` for an e-signature framework, it remained there the next time for a non e-signature framework (DOS4, in this case):

admin-esig-bug

Converting the status_labels to a function ensures we get a fresh copy each time.


gidsg (approved):
good spot 👍

PR #677, opened 2020-11-20
Pull updated digital marketplace dependencies

bjgill:
Trello: https://trello.com/c/naKfCp2j/2108-however-when-i-click-the-link-im-getting-this

Some of them are quite outdated, which is causing problems with DOS 5. Update so admins can access DOS 5 things.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1240,8 +1240,8 @@
       "integrity": "sha1-8NZtA2cqglyxtzvbP+YjEMjlUrc="
     },
     "digitalmarketplace-frameworks": {
-      "version": "git+https://github.com/alphagov/digitalmarketplace-frameworks.git#e02ef939c571a0a85a693eae0d72175e6283f1b5",
-      "from": "git+https://github.com/alphagov/digitalmarketplace-frameworks.git#v17.7.0"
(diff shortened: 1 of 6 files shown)


gidsg (approved):
👍 🆙

.nvmrc
```diff
@@ -1 +1 @@
-10.19.0
+10.23.0
```
gidsg, line 1: I think a global version is expected by dmrunner so you may need to update this first

gidsg (changes requested):

gidsg (approved):
Can you revert the node change for now as I think it'd be better to co-ordinate this with dmrunner, otherwise good to merge.

bjgill:
Removed the node bump. Though dmrunner seemed perfectly happy with it for me...

Also remembered that I actually needed to run `npm install` and `make freeze-requirements`

bjgill:
I've split the updates into their own commits to make it easier to debug. It turns out that it's the content loader update which is to blame.

I'm now out of time for today. If someone wants to pick this up on Monday, feel free. If you can work out what's wrong from the error, great. Otherwise, you'll probably want to bisect content loader versions until you work out which content loader version was the one that broke the tests.

lfdebrux (approved):

PR #689, opened 2021-01-15
Use e-signature boolean in API response

gidsg:
https://trello.com/c/qmUdpw9u/656-1-e-signature-cleanup-use-api-boolean-to-check-frameworks

Use e-signature boolean in API response to determine e-signature frameworks rather than hardcoded heuristic.

### DIFF
--- a/app/main/views/agreements.py
+++ b/app/main/views/agreements.py
@@ -20,18 +20,6 @@ def get_status_labels():
     ))
 
 
-OLD_COUNTERSIGNING_FLOW_FRAMEWORKS = [
-    'digital-outcomes-and-specialists-4',
-    'digital-outcomes-and-specialists-3',
-    'digital-outcomes-and-specialists-2',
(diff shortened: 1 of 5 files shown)


app/main/views/agreements.py
```diff
@@ -20,18 +20,6 @@ def get_status_labels():
...
-OLD_COUNTERSIGNING_FLOW_FRAMEWORKS = [
-    'digital-outcomes-and-specialists-4',
-    'digital-outcomes-and-specialists-3',
-    'digital-outcomes-and-specialists-2',
-    'digital-outcomes-and-specialists',
-    'g-cloud-11',
-    'g-cloud-10',
-    'g-cloud-9',
-    'g-cloud-8',
-]
```
bjgill, line 32: :heart:

bjgill (approved):
Looks sensible.

PR #709, opened 2021-03-16
Remove unnecessary try/except

bjgill:
The except block was just re-throwing the exception - `status_code` was not used when set in that code path.

Thus, we can simplify be removing the unnecessary try/except block.


lfdebrux (approved):
Nice spot

bjgill:
Missed unused import. Will merge once green.

PR #736, opened 2021-04-20
Show last two expired G-Cloud frameworks

bjgill:
https://trello.com/c/wTPrDiEp/1565-admin-framework-manager-bug

Admins still need to handle requests for old G-Cloud frameworks. Make this easier for them by including links for the last two expired G-Cloud frameworks on the homepage.

There was [a deliberate decision to exclude old G-Cloud frameworks made in 2017](https://trello.com/c/wxCSXbeH/159-update-homepage-group-framework-tasks-by-framework-and-add-headings). However, I can find nothing that says why the change was made. So I think we're probably safe enough to reverse that decision.

With this change, the homepage now looks like:

image

### DIFF
--- a/app/main/helpers/service.py
+++ b/app/main/helpers/service.py
@@ -2,6 +2,29 @@ import datetime
 import re
 
 
+def filter_relevant_frameworks(frameworks):
+    """
+    Show admins only the frameworks that they're likely to be interested in.
+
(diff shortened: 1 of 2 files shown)


bjgill:
I no longer remember why I paused work on this. It seems vaguely useful - particularly given that G12 expires later this year. Let's see if I can get this merged...

tim-s-ccs (approved):

PR #745, opened 2021-05-14
Use v2 syntax for dependency ignores

bjgill:
I was using v1 syntax, which was why Dependabot didn't like it, since the file uses v2 syntax.

I've now checked the updated version using the validator (https://dependabot.com/docs/config-file/validator/), so I am more confident this is finally correct.

Ignore docs: https://docs.github.com/en/code-security/supply-chain-security/configuration-options-for-dependency-updates#ignore

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -13,13 +13,11 @@ updates:
   schedule:
     interval: weekly
     time: "03:00"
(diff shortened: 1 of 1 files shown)


alex9smith (approved):
Also checked in the validator - looks fine now!

PR #767, opened 2021-07-06
Show page with status of all frameworks

bjgill:
https://trello.com/c/DXsXHeZo/96-let-admins-change-the-non-prod-framework-state

This is the first part of a firebreak project to give admins in non-prod environments the ability to view and change framework states. admin-frontend doens't use the design system yet, so I have to use the toolkit instead.

Since this is a firebreak experiment, I'm hiding this in production.

image

### DIFF
--- a/app/main/views/services.py
+++ b/app/main/views/services.py
@@ -1,3 +1,4 @@
+import os
 from collections import OrderedDict
 from itertools import chain, dropwhile, islice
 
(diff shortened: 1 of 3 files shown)


bjgill:
To give a sense of where this is headed: my plan is to add an 'edit' link to each row that will allow the user to change the framework state.

alex9smith (approved):
Looks good!

PR #768, opened 2021-07-07
Add form to change framework status

bjgill:
This is the second part of a firebreak project to give admins in non-prod environments the ability to view and change the framework state. Follows on from https://github.com/alphagov/digitalmarketplace-admin-frontend/pull/767.

Create the form that admins will use to change the state. Wire it up so that it shows the current state. The form doesn't yet do anything on POST other than redirect back to the previous page.

Admin frontend has not been modernised, so I'm following the existing patterns for radios in this app.

Since this is a firebreak experiment, I'm hiding this in production.

image

image

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -267,3 +267,17 @@ class EditSupplierCompanyRegistrationNumberForm(FlaskForm):
             valid = False
 
         return valid
+
+
+class EditFrameworkStatusForm(FlaskForm):
+
+    status = DMRadioField(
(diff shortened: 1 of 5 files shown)


bjgill:
I've tested this locally, and all seems well. The next (and final) bit of work will be to make the form actually be able to change the status.

bjgill:
Rebased to fix flake8 error

gidsg (approved):

PR #769, opened 2021-07-07
Revert "Bump gulp-sass from 4.1.0 to 5.0.0"

bjgill:
Reverts alphagov/digitalmarketplace-admin-frontend#762

I don't know why, but the visual regression tests are much flakier since we merged this. I can't find a single successful test run after this was merged. Revert to see if this stabilises the tests.


gidsg (approved):
👍 🤞

phantas:
This triggered a new dependabot alert that cannot be fixed due to dependacy: https://github.com/alphagov/digitalmarketplace-admin-frontend/security/dependabot/package-lock.json/trim-newlines/open
Any opinion on how to go from here?

bjgill:
Oh no. It looks as if we need to work out and fix what's going wrong and then reinstate the original change. Possibly worth raising as a story in the support board since it doesn't seem easy? Or even the main board if it's too big for support?

PR #770, opened 2021-07-07
Let non-prod admins change framework statuses

bjgill:
https://trello.com/c/DXsXHeZo/96-let-admins-change-the-non-prod-framework-state

This is the third part of a firebreak project to give admins in non-prod environments the ability to view and change the framework state. It follows on from https://github.com/alphagov/digitalmarketplace-admin-frontend/pull/768.

Update the 'change framework status' page to actually change the framework status. The full user journey now works.

I've tested that this works locally. This page is hidden in production, so should be low-risk if broken.

Note that this doesn't handle clarification questions - that's for future work.

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -273,6 +273,7 @@ class EditFrameworkStatusForm(FlaskForm):
 
     status = DMRadioField(
         "Framework status",
+        hint="Tell the rest of the team before making a change",
         validators=[
             validators.InputRequired(message='You must choose a framework status')
         ],
--- a/app/main/views/services.py
+++ b/app/main/views/services.py
@@ -77,11 +77,16 @@ def change_framework_status(framework_slug):
     if not framework:
         abort(404, "Framework not found")
 
-    if request.method == "POST":
-        return redirect(url_for('.view_frameworks'))
-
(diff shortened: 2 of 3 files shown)


gidsg:
Are there view changes that go with this or are they coming separately?

gidsg:
@bjgill Oh I see, this is wiring up the stuff from https://github.com/alphagov/digitalmarketplace-admin-frontend/pull/768
Looks fine to me, only thing is do we want a warning like, 'let the team know before you do this'? (Not essential for a firebreak thing) in case someone is deploying or something.

gidsg (approved):
Looks good to me aside from minor optional comment https://github.com/alphagov/digitalmarketplace-admin-frontend/pull/770#issuecomment-875653576

bjgill:
> Looks fine to me, only thing is do we want a warning like, 'let the team know before you do this'? (Not essential for a firebreak thing) in case someone is deploying or something.

That's a great idea. I'll add a question hint with such a warning.

Arguably, this is something that warrants a confirmation page to force people to confirm they've warned the rest of the team. That's a larger change, though, so I'll mention it in the story and consider it for a future PR

PR #772, opened 2021-07-09
Use govukTable for viewing users

bjgill:
We were previously using toolkit/summary-table.html. We want to replace all uses of the toolkit with design system components. This is part of our migration to Design System 3 to improve accessibility. Admin-frontend is not a priority app, but I was working in the area for my firebreak project, so did this in passing.

No obvious changes in behaviour - just minor formattting differences.

Old:

image

New:

image

### DIFF
--- a/app/templates/view_users.html
+++ b/app/templates/view_users.html
@@ -88,7 +88,7 @@
               <input type="hidden" name="csrf_token" value="{{ csrf_token() }}"/>
               <input type="hidden" name="source" value="{{ url_for('.find_user_by_email_address', email_address=email_address) }}"/>
               {{ govukButton({
-                "classes": "govuk-button--warning",
+                "classes": "govuk-button--warning govuk-!-margin-bottom-0",
                 "text": "Deactivate"
               }) }}
             </form>
@@ -101,7 +101,7 @@
               <input type="hidden" name="csrf_token" value="{{ csrf_token() }}"/>
               <input type="hidden" name="source" value="{{ url_for('.find_user_by_email_address', email_address=email_address) }}"/>
               {{ govukButton({
-                "classes": "govuk-button--secondary",
+                "classes": "govuk-button--secondary govuk-!-margin-bottom-0",
                 "text": "Activate"
               }) }}
             </form>
--- a/app/templates/view_users.html
+++ b/app/templates/view_users.html
@@ -1,5 +1,3 @@
-{% import "toolkit/summary-table.html" as summary %}
-
 {% extends "_base_page.html" %}
 
 {% block pageTitle %}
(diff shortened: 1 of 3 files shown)


app/templates/view_users.html
```diff
@@ -128,8 +108,35 @@ <h1 class="govuk-heading-xl">Find a user</h1>
...
+        {'text': 'Name'},
+        {'text': 'Role'},
+        {'text': 'Supplier'},
+        {'text': 'Last login'},
+        {'text': 'Last password change'},
+        {'text': 'Locked'},
+        {'text': 'Change status'},
+      ],
+      'rows': ns.rows,
+      "classes": "dm-table govuk-!-margin-bottom-6",
```
domoscargin, line 137: I don't think `dm-table` is strictly necessary here if we aren't doing any custom styling - the only use of it I can find is for an override on supplier frontend: https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/8022f84b54d89b8a877cac894ffa68289c3d1a97/app/assets/scss/overrides/_table.scss

app/templates/view_users.html
```diff
@@ -1,5 +1,3 @@
-{% import "toolkit/summary-table.html" as summary %}
```
domoscargin, line 1: 🎉

domoscargin (approved):
Yay!

Could probably be its own PR, but if it's a quick win to switch out the toolkit textbox on line 30 with a `govukInput` (and add `novalidate` to the form), that should mean we wouldn't need to revisit this page on the Design System work.

app/templates/view_users.html
bjgill: Do we even need the `govuk-!-margin-bottom-6`? It doesn't seem to make any difference if I remove it...

bjgill:
Good idea - seems sensible to just do the whole of this page in one go.

app/templates/view_users.html
domoscargin: Good catch - `govukTable` includes that margin by default: https://github.com/alphagov/govuk-frontend/blob/6cf094045edb4293fcc9531a955787f08ff8f8a4/src/govuk/components/table/_index.scss#L6

app/templates/view_users.html
```diff
@@ -23,87 +21,71 @@
...
-      name = "email_address"
-    %}
-      {% include "toolkit/forms/textbox.html" %}
-    {% endwith %}
+  <form action="{{ url_for('.find_user_by_email_address') }}" method="get" class="question" novalidate>
+    {{ govukInput({
+      "label": {
+        "text": "Find a user by email",
+        "classes": "govuk-label--m",
+        "isPageHeading": false
```
domoscargin, line 29: This isn't a required field and `false` is the default, but no harm in spelling it out.

app/templates/view_users.html
```diff
@@ -23,87 +21,71 @@
...
-      {% call summary.field() %}
-        {% if item.locked and current_user.has_any_role('admin', 'admin-ccs-category') and not item.personalDataRemoved %}
-          <form action="{{ url_for('.unlock_user', user_id=item.id) }}" method="post">
+      {% set locked %}
+        {% if user.locked and current_user.has_any_role('admin', 'admin-ccs-category') and not user.personalDataRemoved %}
+          <form action="{{ url_for('.unlock_user', user_id=user.id) }}" method="post">
             <input type="hidden" name="csrf_token" value="{{ csrf_token() }}"/>
             <input type="hidden" name="source" value="{{ url_for('.find_user_by_email_address', email_address=email_address) }}"/>
             {{ govukButton({
               "classes": "govuk-button--secondary",
```
domoscargin, line 74: I think if we add a `govuk-!-margin-bottom-0` to this, it'll align like the rest of the row. From the screenshot, it currently looks slightly off.

domoscargin (approved):
A coupla small comments, but otherwise this looks great!

app/templates/view_users.html
bjgill: Nice - looks much better now:

image

app/templates/view_users.html
bjgill: Removed.

bjgill:
Thanks - I've made a couple of small changes following your review. Will merge once the tests pass

PR #773, opened 2021-07-09
Let admins change user's names themselves

bjgill:
We occasionally get requests from users to change their name. CCS support forward these requests on to us. We change the name via the API. We don't actually provide any value - if an admin asks us to change a user's name, we do it.

So it is safe to cut out the developer in the middle and give admins the ability to change user's names themselves. This should save both sides time.

This action is already audited via the API, so we will have a record of any changes admins make.

Once this is merged, we should tell CCS about the change, and also update the manual (https://alphagov.github.io/digitalmarketplace-manual/2nd-line-runbook/support-tasks.html?highlight=admin#changing-a-user-s-name).

Example of the new link:

image

And the page:

image

### DIFF
--- a/app/templates/view_buyers.html
+++ b/app/templates/view_buyers.html
@@ -53,7 +53,13 @@
 
       {% call summary.row() %}
 
-        {{ summary.field_name(item.name) }}
+        {% call summary.field() %}
+          {% if current_user.has_any_role('admin') %}
+            <a class="govuk-link" href="{{ url_for('.change_user_name', user_id=item.id) }}">{{ item.name }}</a>
+          {% else %}
+            {{ item.name }}
+          {% endif %}
+        {% endcall %}
 
         {% call summary.field() %}
           {{ item.emailAddress }}
--- a/app/templates/view_users.html
+++ b/app/templates/view_users.html
@@ -40,6 +40,15 @@
   {% if users %}
     {% set ns = namespace(rows = []) %}
     {% for user in users %}
+
+      {% set name %}
+        {% if current_user.has_any_role('admin') %}
+          <a class="govuk-link" href="{{ url_for('.change_user_name', user_id=user.id) }}">{{ user.name }}</a>
+        {% else %}
+          {{ user.name }}
+        {% endif %}
+      {% endset %}
+
       {% set supplier %}
         {% if user.role == 'supplier' %}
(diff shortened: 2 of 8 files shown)


app/templates/view_buyers.html
```diff
@@ -53,7 +53,13 @@ <h1 class="govuk-heading-xl">Find a buyer</h1>
 
       {% call summary.row() %}
 
-        {{ summary.field_name(item.name) }}
+        {% call summary.field() %}
+          {% if current_user.has_any_role('admin') %}
+            <a class="govuk-link" href="{{ url_for('.change_user_name', user_id=item.id) }}">{{ item.name }}</a>
```
phantas, line 58: an edit button would probably be more intuitive but agree a link makes the most sense here :+1:

app/templates/view_users.html
```diff
@@ -113,7 +122,7 @@ <h1 class="govuk-heading-xl">Find a user</h1>
 
       {% set row = ns.rows.append(
         [
-          {"text": user.name},
+          {"text": name},
```
phantas, line 125: I would need to test, but in the case name is a link, shouldn't we use "html" instead of "text"?

tests/app/main/views/test_users.py
```diff
@@ -82,36 +82,16 @@ def test_should_show_buyer_user(self):
 
         document = html.fromstring(response.get_data(as_text=True))
 
-        name = document.xpath(
-            '//tr[@class="govuk-table__row"]//td/text()')[0].strip()
-        assert name == "Test User"
+        [name, role, supplier, last_login, last_password_changed, locked, _button_text] = [
+            field.xpath("normalize-space(string())")
+            for field in document.xpath('//tr[@class="govuk-table__row"]//td')
+        ]
```
phantas, line 88: 😻

phantas (commented):
Just a need a quick check before approving

app/templates/view_buyers.html
bjgill: Yes. I wanted to make the smallest possible change - there are a lot of things that one could edit here.

app/templates/view_users.html
bjgill: Hmm. That's odd. I agree with you - I would have thought so as well, but 'text' seems to work fine - the link renders correctly. If I switch it over to 'html', I can see no difference. :shrug: 

But there doesn't seem to be any harm, and 'html' is correct, so I've switched over

phantas (approved):
🚀

bjgill:
Rebased to fix conflict

PR #775, opened 2021-07-12
Change framework status radio

bjgill:
I introduced two new pages in https://github.com/alphagov/digitalmarketplace-admin-frontend/pull/770. They used toolkit components, so were not very accessible.

Replace the toolkit components with the more accessible design system equivalents. Also add some visually hidden text to help screen reader users understand the table.

Minor formatting changes - no difference in functionality.

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -7,7 +7,7 @@ from dmutils.forms.fields import DMRadioField, DMStripWhitespaceStringField
 
 from flask import current_app
 from flask_wtf import FlaskForm
-from wtforms import validators
+from wtforms import validators, RadioField
 from wtforms.validators import DataRequired, AnyOf, InputRequired, Length, Optional, Regexp, ValidationError
 
 from .. import data_api_client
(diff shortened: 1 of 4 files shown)


app/main/forms.py
```diff
@@ -271,14 +271,21 @@ def validate(self):
...
 
-    status = DMRadioField(
-        "Framework status",
-        hint="Tell the rest of the team before making a change",
+    status = RadioField(
+        id="input-status-1",  # TODO: change to input-copy_service when on govuk-frontend~3
+        label="Framework status",
+        description="Tell the rest of the team before making a change",
         validators=[
             validators.InputRequired(message='You must choose a framework status')
```
domoscargin, line 279: To follow [Design System guidance](https://design-system.service.gov.uk/components/radios#if-there-are-more-than-two-options), this should be "Select a framework status"

app/templates/view_frameworks.html
```diff
@@ -22,21 +21,27 @@
...
-    {% call summary.row() %}
-      {{ summary.text(framework.name) }}
-      {{ summary.text(framework.status) }}
-      {{ summary.edit_link(label="Change", link=url_for(".change_framework_status", framework_slug=framework.slug), hidden_text=framework.name) }}
-    {% endcall %}
-  {% endcall %}
+    {% set ns = namespace(rows = []) %}
+    {% for framework in frameworks %}
+      {% set change_link %}
+        <a class="govuk-link" href="{{ url_for('.change_framework_status', framework_slug=framework.slug) }}">Change<span class="govuk-visually-hidden"> status of {{ framework.name }}</span></a>
```
domoscargin, line 27: 👌

domoscargin (approved):
Small content note, but otherwise this looks good.

app/main/forms.py
bjgill: Good catch, thanks. Fixed.

bjgill:
Thanks. Fixed that, so will merge

